PR #2, opened 2018-05-17
Experimental: automatic whitelist/blacklist for fp16 training

cbcase:
This PR contains v0.1 of an experimental tool to automatically enable fp16 in PyTorch with a whitelist/blacklist model for handling automatic type conversion. The included README describes both the interface and implementation in more detail.

The primary apex-wide change is a couple hacks to enable compiling the custom fused scaling + overflow checking kernel that amp uses for it's loss scaling. Carilli and I are planning to move all the kernel implementations to the new (as of 0.4) cpp-extension interface, so my hope is these hacks are short-lived.

The strangest hack is including the `__init__.py` files under `apex/amp/_C/` in the source tree. They are synthesized by cffi, but when it's run for packaging (ie, `python setup.py install`), it doesn't synthesize them -- so the only way to get them correctly copied over to the install directory is for them to already exist. This will all go away soon.

### DIFF
--- a/apex/__init__.py
+++ b/apex/__init__.py
@@ -1,3 +1,4 @@
+from . import amp
 from . import RNN
 from . import reparameterization
 from . import fp16_utils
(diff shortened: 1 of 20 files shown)


mcarilli:
Merged by command line.

PR #6, opened 2018-05-31
LARC clipping+documentation

raulpuric:
* Proper implementation of LARC clipping
 * Documentation of LARC class
 * Modification of FP16_Optimizer to absorb optimizer instance that's being wrapped instead of creating new optimizer instance of same class.

### DIFF
--- a/apex/parallel/LARC.py
+++ b/apex/parallel/LARC.py
@@ -4,11 +4,45 @@ from torch.autograd import Variable
 from torch.nn.parameter import Parameter
 
 class LARC(object):
-    def __init__(self, optimizer, trust_coefficient=0.02, epsilon=1e-8):
+    """
+    :class:`LARC` is a pytorch implementation of both the scaling and clipping varients of LARC,
+    in which the ratio between gradient and parameter magnitudes is used to calculate an adaptive 
+    local learning rate for each individual parameter. The algorithm is designed to improve
+    convergence of large batch training.
+     
+    See https://arxiv.org/abs/1708.03888 for calculation of the local learning rate.
+
+    In practice it modifies the gradients of parameters as a proxy for modifying the learning rate
+    of the parameters. This design allows it to be used as a wrapper around any torch.optim Optimizer.
+
+    ```
+    model = ...
+    optim = torch.optim.Adam(model.parameters(), lr=...)
+    optim = LARC(optim)
+    ```
+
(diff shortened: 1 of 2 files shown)


apex/parallel/LARC.py
```diff
@@ -4,11 +4,45 @@
...
+    optim = torch.optim.Adam(model.parameters(), lr=...)
+    optim = LARC(optim)
+    optim = apex.fp16_utils.FP16_Optimizer(optim)
+    ```
+
+    Args:
+        optimizer: Pytorch optimizer to wrap and modify learning rate for.
+        trust_coefficient: Trust coefficient for calculating the lr. See https://arxiv.org/abs/1708.03888
+        clip: Decides between clipping or scaling mode of LARC. If `clip=True` the learning rate is set to `min(optimizer_lr, local_lr)` for each parameter. If `clip=False` the learning rate is set to `local_lr*optimizer_lr`.
+        eps: epsilon kludge to help with numerical stability while calculating adaotive_lr
```
brettkoonce: sp: adaotive_lr --> adaptive_lr

apex/parallel/LARC.py
```diff
@@ -4,11 +4,45 @@
 from torch.nn.parameter import Parameter
 
 class LARC(object):
-    def __init__(self, optimizer, trust_coefficient=0.02, epsilon=1e-8):
+    """
+    :class:`LARC` is a pytorch implementation of both the scaling and clipping varients of LARC,
```
brettkoonce: sp: varients --> variants

mcarilli:
My rationale for creating a new instance of the passed optimizer's class within FP16_Optimizer's constructor was that if the passed optimizer had been used earlier, it might have created some momentum or other ancillary buffers in FP16.  I would have to trace through the optimizer's param_groups and cast all these ancillary buffers to FP32 as well.  This is doable (it's similar to what torch.Optimizer.load_state_dict does) but seemed more brittle than FP32-ifying the param_groups then using them to create a fresh optimizer instance.

Your proposed change (using the passed optimizer directly) does impose the additional restriction that the passed optimizer has not been used beforehand/does not contain any ancillary buffers (aside from its owned parameters) that might be FP16.  All my documentation and examples work this way, although the requirement is not yet stated explicitly, so I suppose it's fine to accept.

PR #15, opened 2018-06-22
Adding changes due to https://github.com/pytorch/pytorch/pull/8660

syed-ahmed:
Build was breaking against upstream PyTorch. Could have just changed the header to the local `CUDATypeConversion.cuh` but decided to keep things consistent with the ustream PR.


ngimel:
But this will break builds with earlier pytorch versions?

mcarilli:
In some cases, I locked in local copies of upstream code to make us less vulnerable to upstream changes.  The type conversion templates are one of those cases.  I will make the sure build works tomorrow in a backward-and-current compatible way.

syed-ahmed:
Gotcha. Sounds good :)

PR #17, opened 2018-06-24
change Dockerfile FROM to use nvcr.io image

moconnor725:


mcarilli:
Dockerfile has been updated to use nvcr.io/nvidia/pytorch:18.04-py3 as default.

PR #18, opened 2018-06-26
Fixed deadlock issue

rotabulo:
This fixes an issue in `flat_dist_call`, which leads to a deadlock of processes. The issue is due to processes cycling over tensor types in different orders and calling non matching collective operations in the body of the cycle.


csarofeen:
I'm surprised I haven't hit this deadlock, wonder if different Python versions are more sensitive than others.

Thanks for the contribution!

PR #20, opened 2018-06-26
More stringent check for parameter changes to trigger refresh of distributed

mcarilli:
Previously, the check would fail to catch if a parameter was removed from the end of module.parameters(), but no other ordering changes occurred.
@csarofeen 

Also removed unnecessary if statement in closure handling for dynamic loss scaling.


csarofeen:
LGTM

PR #21, opened 2018-06-28
Fix to imagenet main.py data normalization.

romerojosh:
In the main.py script, the fast_collate function returns image data as uint8 in range [0,255]. This data is converted directly to float without range scaling, i.e. it keeps the range [0.0, 255.0]. As such, the normalization mean and stddev need to be scaled by 255 to work correctly.


mcarilli:
lgtm.  @csarofeen ?

PR #22, opened 2018-06-29
Fixes to validation in imagenet example scripts.

romerojosh:
In the validate routine, the reduced precision values (reduced_prec1 and reduced_prec5) are not reported. Instead, the unreduced prec1 and prec5 are reported. Additionally, there are no guards against reduction calls when not running distributed (as there are in the train routine) which causes an error.

I have changed the validate routine to match the handling of these variables from train.


csarofeen:
Thanks Josh.

PR #40, opened 2018-09-10
amp unit tests

cbcase:
Main change here is the first set of unit tests for amp. Tests all the different casting cases. Tests for things like loss scaling or the optimizer wrapper are still TODO.

Also adds experimental support for calling `handle._deactivate()` on an initialized amp handle -- it unwinds all monkey patching. This makes test running easier.

Run `python -m unittest discover -v` in the apex root directory to run tests.


cbcase:
Merging after offline conversation with @mcarilli

PR #42, opened 2018-09-13
load mode before creating master_params when fp16

wangdongxuking61:
When using fp16 and args.resume is True, the model should be loaded before creating master_params variable. Otherwise, the value in master_params is just random and not same as the loaded file because tensor in master_params is cloned before loaded model.

### DIFF
--- a/examples/imagenet/main.py
+++ b/examples/imagenet/main.py
@@ -123,7 +123,20 @@ def main():
     if args.distributed:
         # shared param turns off bucketing in DDP, for lower latency runs this can improve perf
         model = DDP(model)
-
+    
+    # optionally resume model from a checkpoint
+    if args.resume:
+        if os.path.isfile(args.resume):
+            print("=> loading checkpoint '{}'".format(args.resume))
+            checkpoint = torch.load(args.resume, map_location = lambda storage, loc: storage.cuda(args.gpu))
+            args.start_epoch = checkpoint['epoch']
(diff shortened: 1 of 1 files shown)


mcarilli:
You're right, good catch.  However, looking at this more carefully, the best thing to do is save and restore the fp32 master params directly, and copy those into the model params when resuming.  I'll implement that myself.

mcarilli:
Hopefully handled via https://github.com/NVIDIA/apex/commit/327b24469236e467ce3a91fac9fea6a3540af56d.  The examples now resume from checkpoints with the expected loss.

Let me know if this works for you.

wangdongxuking61:
> Hopefully handled via [327b244](https://github.com/NVIDIA/apex/commit/327b24469236e467ce3a91fac9fea6a3540af56d). The examples now resume from checkpoints with the expected loss.
> 
> Let me know if this works for you.

well-done. I reviewed the code, it looks like right.
BTW, i think the following code 
```
saved_master_params = checkpoint['master_params']
for master, saved in zip(master_params, saved_master_params):
    master.data.copy_(saved.data)
```
can be replaced by 
```
if args.fp16:
    saved_master_params = checkpoint['master_params']
    for master, saved in zip(master_params, saved_master_params):
        master.data.copy_(saved.data)
```
because when using fp32, `master_params` is just reference to model's Parameters, calling `model.load_state_dict(checkpoint['state_dict'])` is enough.

PR #48, opened 2018-09-21
[syncBN]

jjsjann123:
added syncBN in native pure python apex
  added fused cuda kernels used for sync BN. Using welford for mean/var
    optional installation using 'python setup.py install --cuda_ext'
  added unit test with side to side comparison between apex sync BN with
    PyTorch BN. Notice that for pytorch BN implementation, because of
    numerical issue for mean/var, the output will be slightly off.


apex/parallel/__init__.py
```diff
@@ -1 +1,10 @@
 from .distributed import DistributedDataParallel, Reducer
+try:
+    import syncbn
+    print("using fused syncBN")
+    from .optimized_sync_batchnorm import SyncBatchNorm
+    from .optimized_sync_batchnorm import replace_with_SYNCBN
+except:
```
mcarilli: Can we make this `except ImportError:` for a bit more safety?

csrc/welford.cu
```diff
@@ -0,0 +1,415 @@
...
+  // load data; 
+  mean_l[thread_id] = mean[input_base];
+  m2n_l[thread_id] = var_biased[input_base] * numel;
+  num_item_l[thread_id] = numel;
+
+  __syncthreads();
+  
+  // parallel reduce with interleaved threads
+  // TODO(jie): unroll this?
+  for (int offset = powf(2.0, ceil(log2f(block_size))); offset > 0; offset>>=1) {
```
mcarilli: Instead of int offset = powf(2.0, ceil(log2f(block_size))), which is a float->int conversion, three [heavyweight](https://docs.nvidia.com/cuda/cuda-c-best-practices-guide/index.html#math-libraries) math library calls, then a float->int coversion, this gadget I wrote might be helpful:
https://github.com/pytorch/pytorch/blob/66b3bae47c61cfc6c345974c60fcd734ffe84365/aten/src/THC/THCReduce.cuh#L35-L41

I have some questions about your launch configurations though, which would have to interact with this.  Let's talk offline.

mcarilli:
Jie, this is a heroic effort.  I have some questions/suggestions, but given that you sit right next to me, we can work through them in person.

csrc/welford.cu
```diff
@@ -0,0 +1,429 @@
...
+      mean_l[thread_id] = (val * count + val2 * count2) / (count + count2);
+      val = val - val2;
+      m2n_l[thread_id] += m2n_l[thread_id + offset] + val*val*count*count2/(count+count2);
+      num_item_l[thread_id] = count + count2;
+    }
+    __syncthreads();
+  }
+
+  if (thread_id == 0) {
+    out_mean[blockIdx.x] = static_cast<scalar_t>(mean_l[0]);
```
ngimel: intermediates for half data are written out as fp32 in cudnn. It may well be that it's not needed, but would be good to keep your options open may be with another template parameter.

csrc/welford.cu
```diff
@@ -0,0 +1,429 @@
...
+      const int ss,
+      const float eps) {
+  int address_base = blockIdx.x*ss + blockIdx.y*gridDim.x*ss;
+
+  auto m_c = static_cast<accscalar_t>(mean[blockIdx.x]);
+  auto var_c = sqrt(static_cast<accscalar_t>(var[blockIdx.x]) + eps);
+  auto w_c = static_cast<accscalar_t>(weight[blockIdx.x]);
+  auto s_c = static_cast<accscalar_t>(shift[blockIdx.x]);
+
+  // sequential welford
```
ngimel: misleading comment

csrc/welford.cu
```diff
@@ -0,0 +1,429 @@
...
+    const at::Tensor mean,
+    const at::Tensor var,
+    const at::Tensor weight,
+    const at::Tensor shift,
+    const float eps) {
+  const auto batch_size = input.size(0);
+  const auto feature_size = input.size(1);
+  at::Tensor out = at::empty_like(input);
+
+  auto space_size = input.size(2);
```
ngimel: This setup code is in every function, should be refactored into a separate function

apex/parallel/optimized_sync_batchnorm_kernel.py
```diff
@@ -0,0 +1,68 @@
...
+from torch.autograd.function import Function
+
+import syncbn
+
+class SyncBatchnormFunction(Function):
+
+    @staticmethod
+    def forward(ctx, input, weight, bias, running_mean, running_variance, eps, track_running_stats = False, momentum = 1.0):
+        torch.cuda.nvtx.range_push("sync_BN_fw")
+        mean, var, var_biased = syncbn.welford_mean_var(input)
```
ngimel: it would be good to have comments on what all these functions do (welford_mean_var, welford_parallel etc), either here or in the backend.

ngimel: Also, I don't see you calling .contiguous on input tensors anywhere, and our optimized kernels depend on input/gradOutput being contiguous.

csrc/welford.cu
jjsjann123: I agree it would be much better to leave it as an option.
I originally leave it as fp16 since bn running_mean/var uses fp16. But in distributed syncBN, intermediate mean/var should be all scattered to compute global mean/var, so it might be safer to output fp32 when running with distributed.

jjsjann123: finished most other parts. Have some questions with the intermediate precision. Will take it offline tomorrow :)

apex/parallel/optimized_sync_batchnorm_kernel.py
jjsjann123: I'll comment it at the syncbn.cpp.
Still intend to clean this up a little bit more

csrc/welford.cu
```diff
@@ -0,0 +1,445 @@
...
+
+  at::Tensor grad_input = at::empty_like(input);
+
+  auto space_size = get_tensor_spatial_size(input);
+
+  const dim3 block(512);
+  // TODO(jie): should I do 1 block per feature?
+  const dim3 grid(feature_size, batch_size);
+
+  AT_DISPATCH_FLOATING_TYPES_AND_HALF(input.type(), "batchnorm_forward", ([&] {
```
t-vi: batchnorm_backward, probably

csrc/welford.cu
```diff
@@ -0,0 +1,469 @@
...
+  
+
+std::vector<at::Tensor> welford_mean_var_CUDA(const at::Tensor input) {
+  const auto batch_size = input.size(0);
+  const auto feature_size = input.size(1);
+
+  auto space_size = get_tensor_spatial_size(input);
+  auto scalar_type = promote_scalartype(input);
+
+  at::TensorOptions t_op(input);
```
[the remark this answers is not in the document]
mcarilli: The most recent syntax I'm aware of (which works in our 18.10 container) is e.g.
```
at::Tensor new_tensor = at::empty({x, y, z}, input.options());
```
Is this still correct to your knowledge?

PR #69, opened 2018-10-30
[WIP] Fused layer norm cuda

thorjohnsen:
Draft release of fused layer norm extension.


apex/normalization/csrc/layer_norm_cuda.cpp
```diff
@@ -0,0 +1,266 @@
...
+}
+
+void check_args(
+    at::IntList normalized_shape,
+    at::Tensor gamma,
+    at::Tensor beta
+    )
+{
+    if (gamma.defined() && !gamma.sizes().equals(normalized_shape)) {
+      std::stringstream ss;
```
ngimel: nit: this can be done with AT_CHECK macro

apex/normalization/csrc/layer_norm_cuda_kernel.cu
```diff
@@ -0,0 +1,975 @@
...
+    at::Tensor* input,
+    int n1,
+    int n2,
+    at::IntList normalized_shape,
+    at::Tensor* gamma,
+    at::Tensor* beta,
+    double epsilon)
+{
+    AT_DISPATCH_ALL_TYPES_AND_HALF(input->type(), "layer_norm_cuda_kernel", ([&] {
+        HostApplyLayerNorm(
```
ngimel: there's no need to dispatch for integer types (normalization operation is ill-defined on integer types anyway), and to avoid specializations, you can use accscalar_t inside dispatch macro (`using accscalar_t = acc_type<scalar_t, true>;`), it defines accscalar_t to float when scalar_t is at::Half, and to scalar_t otherwise (when scalar_t is float or double). Same applies to backward.

setup.py
```diff
@@ -44,7 +44,15 @@
         CUDAExtension(name='syncbn',
                       sources=['csrc/syncbn.cpp',
                                'csrc/welford.cu']))
-
+    ext_modules.append(
+        CUDAExtension(name='fused_layer_norm_cuda',
+                      sources=['apex/normalization/csrc/layer_norm_cuda.cpp',
+                               'apex/normalization/csrc/layer_norm_cuda_kernel.cu'],
+                      extra_compile_args={'cxx': ['-O3',],
+                                          'nvcc':['--gpu-architecture=sm_70',
```
ngimel: Do you know how much perf gain hard-coding to sm_70 brings? I'd rather either not specify it, if it's not a lot, or compile for a few architectures (60,61,70) because apex is a general purpose package.

ngimel:
@Mcarilli do you think cu and cpp files should go into apex/csrc or into apex/normalization/csrc? My vote is for apex/csrc (and I should probably move adam files there also).

setup.py
mcarilli: I'll change it to compile for a few architectures.

apex/normalization/csrc/layer_norm_cuda.cpp
```diff
@@ -0,0 +1,266 @@
...
+    )
+{
+    check_args(input,normalized_shape,n1,n2);
+    check_args(normalized_shape,gamma,beta);
+}
+
+template<typename T> 
+void allocate_layer_norm_output_tensors(
+    at::Tensor input,
+    const T* input_data,
```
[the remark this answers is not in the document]
thorjohnsen: Ok, that's a better way to do it.

apex/normalization/csrc/layer_norm_cuda.cpp
```diff
@@ -0,0 +1,266 @@
...
+    )
+{
+  output = at::empty_like(input);
+  mean = at::empty({n1}, input.options().dtype(at::kFloat));
+  invvar = at::empty_like(mean);
+}
+template<> 
+void allocate_layer_norm_output_tensors(
+    at::Tensor input,
+    const int64_t* input_data,
```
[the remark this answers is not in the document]
thorjohnsen: No, not really. You can't have a gradient for integer types, something I realized after I finished the code and tried to test it. I'll remove it.

apex/normalization/csrc/layer_norm_cuda.cpp
```diff
@@ -0,0 +1,266 @@
...
+    at::Tensor* invvar,
+    at::Tensor* input,
+    int n1,
+    int n2,
+    at::IntList normalized_shape,
+    at::Tensor* gamma,
+    at::Tensor* beta,
+    double epsilon);
+
+#define CHECK_CUDA(x) AT_ASSERTM(x.type().is_cuda(), #x " must be a CUDA tensor")
```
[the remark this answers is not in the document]
thorjohnsen: Ok.

apex/normalization/csrc/layer_norm_cuda_kernel.cu
```diff
@@ -0,0 +1,975 @@
...
+      }
+    }
+    for (;  l < n2;  ++l) {
+      U curr = static_cast<U>(lvals[l]);
+      cuWelfordOnlineSum<U>(curr,mu,sigma2,count);
+    }
+    // intra-warp reductions
+    for (int l = 0;  l <= 4;  ++l) {
+      int srcLaneB = (threadIdx.x+(1<<l))&31;
+      U muB = __shfl_sync(0xffffffff, mu, srcLaneB);
```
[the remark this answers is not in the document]
thorjohnsen: I'll replace it.

apex/normalization/csrc/layer_norm_cuda_kernel.cu
thorjohnsen: I like that, it will simplify the code a little bit.

mcarilli:
> @mcarilli do you think cu and cpp files should go into apex/csrc or into apex/normalization/csrc? My vote is for apex/csrc (and I should probably move adam files there also).

I lean towards subdirectories but not strongly.

apex/normalization/csrc/layer_norm_cuda_kernel.cu
ngimel: You need `#include "ATen/AccumulateType.h"` for this

ngimel:
>I lean towards subdirectories but not strongly.
Ok let's leave it as is then.

thorjohnsen:
I ran a Transformer convergence test with batch_sizes 5120 and 1280. With bs=5120, Transformer converged after 4 epochs and with bs=1280 Transformer converged after 3 epochs. This is same or better than the code it replaces (torch.nn.LayerNorm).

PR #77, opened 2018-11-06
[syncBN]

jjsjann123:
adjusted kernel config for better perf.
removed divergence in welford warp reduction.


jjsjann123:
Forgot to push this here :) @mcarilli

PR #89, opened 2018-12-02
Update examples to PyTorch >=0.4.0

ptrblck:
This PR updates the examples to PyTorch versions >= 0.4.0.

Minor cleanups were performed in all examples:

- Remove `Variables` as they are deprecated
- Remove `torch.cuda.manual_seed`, since `torch.manual_seed` also seeds all CUDA devices if available ([source](https://pytorch.org/docs/master/notes/randomness.html#pytorch))
- Add `with torch.no_grad()` statements where they were missing or `Variable(, volatile=True)` was used

Minor potential bugs fixed:

- [`examples/distributed/main.py`](https://github.com/ptrblck/apex/blob/28bdc04eec8304c4985bb6625828f3c8fdaf321e/examples/distributed/main.py#L178): Convert `correct` to `float`, since `torch.tensor.eq` op returns a `torch.uint8` tensor, which might overflow for large batch sizes
- [`examples/word_language_model/generate.py`](https://github.com/ptrblck/apex/blob/28bdc04eec8304c4985bb6625828f3c8fdaf321e/examples/word_language_model/generate.py#L64): Convert `word_weights` to `float` before calling `torch.multinominal`, since this function is not implemented for `torch.HalfTensors` and yields an error

I've tested these changes on my local machine as far as possible. However, since I only have a Single-GPU system, I couldn't test the distributed examples. Let me know, how to proceed with these examples.

Best,
ptrblck

### DIFF
--- a/examples/distributed/main.py
+++ b/examples/distributed/main.py
@@ -6,7 +6,6 @@ import torch.nn as nn
 import torch.nn.functional as F
 import torch.optim as optim
 from torchvision import datasets, transforms
-from torch.autograd import Variable
 from apex.fp16_utils import to_python_float
 
 #=====START: ADDED FOR DISTRIBUTED======
@@ -82,9 +81,6 @@ if args.distributed:
 #=====END:   ADDED FOR DISTRIBUTED======
 
 torch.manual_seed(args.seed)
-if args.cuda:
-    torch.cuda.manual_seed(args.seed)
-
 
 kwargs = {'num_workers': 1, 'pin_memory': True} if args.cuda else {}
 
@@ -158,7 +154,6 @@ def train(epoch):
     for batch_idx, (data, target) in enumerate(train_loader):
         if args.cuda:
             data, target = data.cuda(), target.cuda()
-        data, target = Variable(data), Variable(target)
         optimizer.zero_grad()
         output = model(data)
         loss = F.nll_loss(output, target)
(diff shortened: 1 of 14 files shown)


examples/distributed/main.py
```diff
@@ -177,11 +172,10 @@ def test():
         with torch.no_grad():
             if args.cuda:
                 data, target = data.cuda(), target.cuda()
-            data, target = Variable(data), Variable(target)
             output = model(data)
             test_loss += to_python_float(F.nll_loss(output, target, size_average=False).data) # sum up batch loss
             pred = output.data.max(1, keepdim=True)[1] # get the index of the max log-probability
-            correct += pred.eq(target.data.view_as(pred)).cpu().sum()
+            correct += pred.eq(target.data.view_as(pred)).cpu().float().sum()
```
mcarilli: Half<->float conversions are generally super slow on the CPU.  Can we change this to `.float().cpu()`?

mcarilli:
This is great, and long since overdue haha...it's been a low priority for me because everything works fine.  But if we intend these to be examples of best practices, using the most up-to-date syntax is definitely important.

examples/distributed/main.py
ptrblck: Sure! Good to know it's slower on the CPU.

examples/imagenet/main_fp16_optimizer.py
```diff
@@ -371,8 +371,6 @@ def validate(val_loader, model, criterion):
     while input is not None:
         i += 1
 
-        target = target.cuda(async=True)
-
         # compute output
```
ptrblck: Remove the `target.cuda()` call, since `data_prefetcher` makes sure to push the data and target onto the GPU.

mcarilli:
Looks good to me

PR #92, opened 2018-12-06
WIP: improve fused adam

FDecaYed:
Adding fp16_optimizer so fused adam can be used in a similar way to apex.fp16_utils
Adding support for parameter groups
Adding support for weight decay( adamW, https://www.fast.ai/2018/07/02/adam-weight-decay/)
Adding support for fused-in clip_grad_norm
Adding automatic parameter flattening
Adding NaN checking for static loss scale and support dynamic loss scaling now

### DIFF
--- a/apex/optimizers/fp16_optimizer.py
+++ b/apex/optimizers/fp16_optimizer.py
@@ -1,0 +1,183 @@
+import torch
+from torch._utils import _flatten_dense_tensors, _unflatten_dense_tensors
+
+class FP16_Optimizer(object):
+    """
+    :class:`FP16_Optimizer` A cutdown version of apex.fp16_utils.FP16_Optimizer.
+    Design to be used in the same way but support only fused optimizers in apex.
+    Refer to apex.fp16_utils documents for more information.
+
+    Example::
+
+        model = torch.nn.Linear(D_in, D_out).cuda().half()
+        optimizer = apex.optimizers.FusedAdam(model.parameters())
+        # Name the FP16_Optimizer instance to replace the existing optimizer
+        # (recommended but not required):
+        optimizer = FP16_Optimizer(optimizer, static_loss_scale = 128.0)
+        ...
+        # loss.backward() becomes:
+        optimizer.backward(loss)
+        ...
+
+    Example with dynamic loss scaling::
+
+        ...
+        optimizer = FP16_Optimizer(optimizer, dynamic_loss_scale=True)
+                                   # optional arg to control dynamic loss scaling behavior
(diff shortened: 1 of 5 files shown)


apex/optimizers/fp16_optimizer.py
```diff
@@ -0,0 +1,183 @@
...
+            fp16_grads_flat (tensor): fp16 grad flattened
+            norm_type (float or int): type of the used p-norm. Can be ``'inf'`` for
+                infinity norm.
+
+        Returns:
+            Total norm of the current fp16 gradients (viewed as a single vector).
+            Returns -1 if the most recently computed fp16 gradients overflowed
+        """
+        # TODO: currently using pre-1.0 api, and not most efficient with copy to cpu and sync
+        norm = float(torch.norm(fp16_grads_flat, p=norm_type))
```
ngimel: This will likely overflow, you need to use the hacky way of outputting fp32 norm until we fix this in the core.

FDecaYed: That is a concern. @mcarilli tested and it seems the internal accumulation is in fp32. But the result is cast back down to fp16 and that could overflow.

apex/optimizers/fused_adam.py
```diff
@@ -31,44 +53,64 @@ class FusedAdam(torch.optim.Adam):
...
-        for group in self.param_groups:
+
+        if grads_group is None:
+            grads_group = [None]*len(self.param_groups)
+        if output_params_group is None:
+            output_params_group = [None]*len(self.param_groups)
+        if grad_norms is None:
+            grad_norms = [None]*len(self.param_groups)
+
+        for group, grads, output_params, grad_norm in zip(self.param_groups, grads_group, output_params_group, grad_norms):
```
ngimel: would this break our current usage, where we pass grads as a list for an (implicit) single group? It would need to be passed as a nested list, right?

ngimel: To make it backwards-compatible, we could check the type of grads_group[0], and wrap it in another list if it's not a list already.

apex/optimizers/fused_adam.py
```diff
@@ -117,6 +95,11 @@ def step(self, closure=None, grads=None, output_params=None, scale=1., grad_norm
                 if clip > 1:
                     combined_scale = clip * scale
 
+            # set bias correction for this group
+            bias_correction = 0
```
ngimel: can you please make it 
```
bias_correction = 1 if group['bias_correction'] else 0
```
?

PR #110, opened 2018-12-22
Update __init__.py

rxy1212:
torch.distributed.new_group and torch.distributed.reduce_op are deprecated on pytorch 1.0.0, this fix can avoid some errors for now.


mcarilli:
The wording of https://github.com/pytorch/pytorch/pull/14767 makes me think that new_group() will be a standard non-deprecated feature of pytorch 1.0 moving forward.  This PR is still helpful though, thanks!

Issue #105, opened 2018-12-19
Training gets stuck when using SyncBN


heilaw:
`DistributedDataParallel` works great for me. But when I use it together with the synchronized batch normalization, either the Python version or the optimized version, the training will get stuck after a few iterations and the code gives the following warning:

> /home/heilaw/.conda/envs/CornerNet/lib/python3.6/multiprocessing/semaphore_tracker.py:143: UserWarning: semaphore_tracker: There appear to be 1 leaked semaphores to clean up at shutdown
  len(cache)) 

Any idea how I should debug it?

heilaw:
Yes, this occurs when I change from non-sync BN to sync BN with no other changes.

mcarilli:
@jjsjann123 have you ever observed this behavior?

jjsjann123:
No I haven't seen this before. @heilaw any chance that you can share a repro script?

heilaw:
I will see if I can share a script that reproduces this issue.

I just found that if I remove `tqdm` from my code, the code won't give the warning but the training still gets stuck after a few iterations.

jjsjann123:
Good catch!
I remember the discussion we had between Teng & Carilli on torch.distributed APIs last week when they were updating the API. I haven't got back and update apex SyncBatchNorm yet.

Thanks for sharing your fix. I'll push a fix to apex soon.

wangxiaodong1021:
> I think this issue is related to `process_group = group_creator()` in `optimized_sync_batchnorm_kernel.py`. In `parallel/__init__.py`, you set `group_creator` to `new_group` if `get_default_group` is not available. However, I don't think that's a good idea. `get_default_group` is not available in PyTorch 1.0, so that creates a new group every time we call the sync BN forward function! It looks like we are using the default group anyway. We may not need that line.
> 
> After I removed that line and `process_group` in both `torch.distributed.all_reduce` and `torch.distributed.all_gather`, the training now works, even with `tqdm`.

I still have the same problem with the latest version. Then I used this method and solved it.
Very good

donnyyou:
Met the same problem with the latest version of apex! How can I fix torch.distributed.all_reduce and torch.distributed.all_gather? @heilaw @jjsjann123 @mcarilli Thanks for your repley!!!

kkjh0723:
@donnyyou I also got the same problem. Did you find a way to fix it?

jjsjann123:
[...]

PR #113, opened 2019-01-02
[syncBN]

jjsjann123:
replacing new_group with torch.distributed.group.WORLD, avoids creating new
group in every iteration.

This should resolve the issue in Training gets stuck when using SyncBN #105

### DIFF
--- a/apex/parallel/__init__.py
+++ b/apex/parallel/__init__.py
@@ -1,14 +1,5 @@
 import torch
 
-# Backward compatibility hack around
-# https://github.com/pytorch/pytorch/pull/14767
-if hasattr(torch.distributed, 'get_default_group'):
-    group_creator = torch.distributed.get_default_group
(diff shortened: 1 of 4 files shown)


jjsjann123:
According to https://github.com/pytorch/pytorch/pull/14767/

> All collectives are supposed to either use group.WORLD, or anything that comes out of new_group. That was the initial design.
We need to make a TODO on removing group.WORLD one day. It exists for backward compatibility reasons and adds lots of complexity.

Until they remove group.WORLD, we can fallback to new_group during our init and cache the new_group so we are not creating a new one each iteration.

mcarilli:
Yeah, I agree this is a better fallback in the interim than the ugly hasattr checks.

donnyyou:
Training gets stuck when using SyncBN with the latest version of apex! Is there something I have missed?

PR #123, opened 2019-01-11
apex.optimizers.FP16_Optimizer: add state_dict() and load_state_dict()

donglixp:
Add state_dict() and load_state_dict() for apex.optimizers.FP16_Optimizer to store and recover optimizer states. These two methods are important to recover from a stored training checkpoint.

### DIFF
--- a/apex/optimizers/fp16_optimizer.py
+++ b/apex/optimizers/fp16_optimizer.py
@@ -214,3 +214,69 @@ class FP16_Optimizer(object):
         self.optimizer.param_groups = value
 
     param_groups = property(_get_param_groups, _set_param_groups)
+    
+    def state_dict(self):
+        """
+        Returns a dict containing the current state of this :class:`FP16_Optimizer` instance.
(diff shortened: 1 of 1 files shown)


vince62s:
@ngimel @mcarilli  as you can see this is really wanted by at least two projects (AllenNLP OpenNMT).
I would fully understand if this is not the direction you would like to take, but could please give us a bit of heads up just to see if we can count on optimzers.FP16_Optimizer or not for our stuff.
Great work anyway, thanks.

mcarilli:
Apex came together as a collection of researchy tools from different sources.  Right now I'm working on creating a unified API for the various optimizers, as well as Amp, that will support saving and loading in a more general way.  I plan to have this new API in master by the end of February.  

If this PR is something that's helpful immediately, I can merge it as a stopgap.

mcarilli:
It looks like the code corresponds directly to what's in apex.fp16_utils.FP16_Optimizer (I know, different tool that came from a different source, working on unifying the interface).  So i'm fine merging this for now.

PR #127, opened 2019-01-20
Added more details for installation in README.MD

bcbcbcbcbcl:
Specify gcc version not later than 6 is required to install with --cuda_ext --cpp_ext and added steps to install gcc-5 & change symlink to point to gcc 5 and g++ 5.

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,10 +94,30 @@ import apex
 ```
 
 ### CUDA/C++ extension
-Apex contains optional CUDA/C++ extensions, installable via
+Apex contains optional CUDA/C++ extensions, installable via (Note: gcc version not later than 6 is required)
 ```
-python setup.py install [--cuda_ext] [--cpp_ext]
+python setup.py install --cuda_ext --cpp_ext
 ```
+
+Install gcc and g++
+```
+sudo apt install g++-5
+sudo apt install gcc-5
+```
+
+Change the symlink to point to gcc 5 and g++ 5
+```
+sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-5 10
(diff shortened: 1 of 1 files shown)


mcarilli:
Is there an associated issue this is meant to resolve?  If cpp and cuda extensions don't work with gcc 6, that is a larger (not Apex-specific) issue worthy of upstream's attention.

Also how did you discover this?  Is gcc 6 the default in ubuntu 18.04?  Our containers, upstream containers, and my bare metal install all have gcc 5.4 so I can't test this immediately.

mcarilli:
Can you build pytorch extensions at all in this environment?  If not, as I said, the problem is bigger than just apex, and worthy of upstream's attention.

Try 
```
cd pytorch/test/cpp_extensions
python setup.py install
```
and see if that works.  You should observe output like https://gist.github.com/mcarilli/05a94a92cd8118fdfec5e168805f74fe (check for the presence of both nvcc and gcc/g++ compilation lines, so you know it's actually seeing cuda on the system).

bcbcbcbcbcl:
I could not locate `pytorch/test/cpp_extensions`.

`python setup.py install` works without --cuda_ext & --cpp_ext.

drscotthawley:
More details would be helpful.  PyTorch tends to install its own version of CUDA (e.g., `conda install pytorch torchvision cudatoolkit=10.0 -c pytorch`), and if instead you install a standalone version of CUDA (e.g. in /usr/local/cuda) then it breaks PyTorch.  
  
But if you want to install Apex with the C++ extensions then you need nvcc which doesn't come with PyTorch. 
Catch-22.

PR #132, opened 2019-01-25
Fix + tests for the eval->training caching issue

mcarilli:
The caching logic is still vulnerable to the user manually modifying weights between eval and training:
```
# eval
with torch.no_grad():
    forward # validation pass, may cache some casts

...user manually modifies weights in some way that the cache cannot detect...

# resume training
forward # uh oh, this will use stale cached values that do not reflect the manual changes...
backward # this will drop the cache
forward # this will refresh the weights, and everything will be ok from here
```
However, this vulnerability was also present with the old caching logic.

I think the only permanent solution is to drop the cache at the end of forward instead of the end of backward (which can be done if we wrap the model like you said, which gives us some leeway to wrap forward).

### DIFF
--- a/apex/amp/utils.py
+++ b/apex/amp/utils.py
@@ -82,7 +82,6 @@ def casted_args(cast_fn, args, kwargs):
     return new_args
 
 def cached_cast(cast_fn, x, cache):
-    # print("Calling cached_cast")
     if is_nested(x):
         return type(x)([cached_cast(y) for y in x])
     if x in cache:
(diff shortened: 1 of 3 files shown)


apex/amp/utils.py
```diff
@@ -86,13 +86,20 @@ def cached_cast(cast_fn, x, cache):
...
+            # Make sure x is actually cached_x's autograd parent.
+            if cached_x.grad_fn.next_functions[1][0].variable is not x:
+                raise RuntimeError("x and cache[x] both require grad, but x is not "
+                                   "cache[x]'s parent.  This is likely an error.")
+        # During eval, it's possible to end up caching casted weights with
+        # requires_grad=False.  On the next training iter, if cached_x is found
+        # and reused from the cache, it will not actually have x as its parent.
+        # Therefore, we choose to invalidate the cache (and force refreshing the cast)
+        # if x.requires_grad and cached_x.requires_grad do not match.
+        if torch.is_grad_enabled() and x.requires_grad != cached_x.requires_grad:
```
cbcase: Re: our slack convo, could you add one sentence explaining why the `torch.is_grad_enabled()` check matters?

cbcase:
Added one comment -- otherwise, LGTM! This is subtle and good

Issue #133, opened 2019-01-26
FP16 BatchNorm with affine=False error


h6197627:
Model converted with network_to_half function fails during forward pass if it has batch normalization layers without affine transformation:

> Traceback (most recent call last):
  File "/home/user/PyTorch/classification.py", line 308, in step
    output = self.model(batch)
  File "/usr/local/lib/python3.5/dist-packages/torch/nn/modules/module.py", line 489, in __call__
    result = self.forward(*input, **kwargs)
  File "/usr/local/lib/python3.5/dist-packages/torch/nn/modules/container.py", line 92, in forward
    input = module(input)
  File "/usr/local/lib/python3.5/dist-packages/torch/nn/modules/module.py", line 489, in __call__
    result = self.forward(*input, **kwargs)
  File "/home/user/PyTorch/models/ResNet.py", line 124, in forward
    x = self.bn2(x)
  File "/usr/local/lib/python3.5/dist-packages/torch/nn/modules/module.py", line 489, in __call__
    result = self.forward(*input, **kwargs)
  File "/usr/local/lib/python3.5/dist-packages/torch/nn/modules/batchnorm.py", line 76, in forward
    exponential_average_factor, self.eps)
  File "/usr/local/lib/python3.5/dist-packages/torch/nn/functional.py", line 1623, in batch_norm
    training, momentum, eps, torch.backends.cudnn.enabled
RuntimeError: expected scalar type Half but found Float

Currently I fix it with such hack applied between network_to_half and forward calls:
```
for module in self.model.modules():
    if isinstance(module, torch.nn.modules.batchnorm._BatchNorm):
        # Hack to fix BN fprop without affine transformation
        if module.weight is None:
            module.weight = nn.Parameter(torch.ones(module.running_var.shape, dtype=module.running_var.dtype, device=module.running_var.device), requires_grad=False)
        if module.bias is None:
            module.bias = nn.Parameter(torch.zeros(module.running_var.shape, dtype=module.running_var.dtype, device=module.running_var.device), requires_grad=False)
```

mcarilli:
Should be fixed by https://github.com/NVIDIA/apex/pull/137, reopen if it doesn't work

h6197627:
@mcarilli, yes, it works now, thanks!
[...]

PR #137, opened 2019-01-28
don't convert to float bn with affine=False

ngimel:
Fix for #133


h6197627:
@ngimel, it works now, thanks!
However, considering [guide](https://docs.nvidia.com/deeplearning/sdk/mixed-precision-training/index.html): "values computed by large reductions should be left in FP32. Examples of this include statistics (mean and variance) computed by batch-normalization", I am interested does applying hack from the issue description is better in terms of training stability than this fix?
Thanks for info!

Issue #134, opened 2019-01-27
Correct Learning Settings of SyncBn layer 


wangjingbo1219:
Hi, recently I use the syncbn layer in semantic segmentation task and replace the syncbn layer from other's repo. But I find that the network with syncbn in this repo can not achieve the same performance as before in the same experiment settings.

In the test code, I find that the optimizer of sycnbn in this repo has the double lr than the origin bn layer. **So I want to know the learning settings of the network with this syncbn layer**.

mcarilli:
Syncbn is intended to help with scaled-out runs, where the minibatch in each process is not big enough to get good batchnorm statistics within each process alone.

Since the only intention of syncbn is to help with the stability of batchnorm statistics, I expect that it shouldn't require changing the learning rate or other hyperparameters.

The LR difference in the tests is a coincidence caused by how they were written:  The distributed test case **averages** gradients from 2 minibatches (this averaging is done by DistributedDataParallel under the hood). The single-process control case handles double the minibatch size (2 effective minibatches) but then **does not average**.  This means that the single-process control case is effectively creating gradients that are twice as large, so @jjsjann123 gave the distributed case a 2X learning rate to compensate and give a fair comparison.  

@jjsjann123 is updating the tests so that the single-process control case also averages over the two effective minibatches.  The learning rate will then be the same for both the distributed and non-distributed test cases.

mcarilli:
^ updated tests to hopefully reduce confusion

PR #138, opened 2019-01-28
[syncBN]

jjsjann123:
test update to resolve
  https://github.com/NVIDIA/apex/issues/134#issue-403525480

Using identical learning rate for both DDP with sync BN and single process BN.
The previous configure leaves the impression that sync BN requires adjusting lr
in the script, which is not true.


tests/synced_batchnorm/two_gpu_unit_test.py
```diff
@@ -92,6 +92,8 @@ def compare(desc, inp1, inp2, error):
 grad_bn = grad_output_t.clone().detach()
 out_bn = bn(inp_bn)
 out_bn.backward(grad_bn)
+for param in bn.parameters():
+    param.grad = param.grad / args.world_size
```
mcarilli: Maybe add a comment here explaining that the division by world_size is included to make the single-process control case equivalent to the averaging over processes that is done by DDP.

PR #144, opened 2019-02-01
Better FP16 support in pytorch fp16 utils.

jma127:
This commit adds an FP16Model class as a successor to `network_to_half()`.

The benefits of this class are:

- Preservation of single-precision for BatchNorm layers. The models
  generated by `network_to_half()` convert BatchNorm moment tensors to
  half-precision, then back to single-precision, which hurts the
  accuracy of the moment estimators and occasionally results in NaNs.
- Support for multi-argument `nn.Module`s (self-explanatory from code).

### DIFF
--- a/apex/fp16_utils/__init__.py
+++ b/apex/fp16_utils/__init__.py
@@ -3,10 +3,13 @@ from .fp16util import (
     network_to_half,
     prep_param_lists,
     model_grads_to_master_grads,
-    master_params_to_model_params, 
+    master_params_to_model_params,
     tofp16,
     to_python_float,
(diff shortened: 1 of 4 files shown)


jma127:
cc @ngimel @mcarilli for review

ngimel:
Ah I see, makes sense.

jma127:
@ngimel Any concerns or does this look good to merge?

jma127:
Added a barebones test with a custom dummy net. Probably makes sense to avoid torchvision dependencies if it can be avoided.

PR #148, opened 2019-02-08
adding a test and example for sync (group) bn with group_size<world_size

ekrimer:


mcarilli:
Thank you!  Can you rig this to run through unit_test.sh like the other tests?

ekrimer:
@mcarilli I added the test into unit_test.sh, though it would require a system with 4 GPUs that would be the minimum to test group_size = 2 < world_size = 4.
I also moved process group creation away from the test into apex, so users can now call it given desired group_size rather than dealing with process group creation nuances.

https://github.com/NVIDIA/apex/blob/ekrimer_add_groupbn_test/apex/parallel/__init__.py#L55

Probably worth moving outside of `__init__.py` at some point

mcarilli:
@jjsjann123 I know you're currently trying to figure out the right way to create and assign process groups to SyncBN instances.  What do you think of the `create_syncbn_proces_group` free function?

jjsjann123:
This is definitely usable.
It is somewhat awkward as it does not give user control on how the group is split.
You can require user to pass all groups to `create_syncbn_process_group2`, but this is conflicting with how DDP API accepts group information (each process is only aware of the group info which it belongs to).

ekrimer:
I guess I didn't think of a scenario user will want a custom grouping. We could enhance this current function later by passing a groups list argument and fallback to this predefined grouping if that argument is None (default).

PR #165, opened 2019-02-20
Use args.batch_size instead of float in main.py

jykzel:
args.batch_size instead of 256

### DIFF
--- a/examples/imagenet/main.py
+++ b/examples/imagenet/main.py
@@ -156,7 +156,7 @@ def main():
     criterion = nn.CrossEntropyLoss().cuda()
 
     # Scale learning rate based on global batch size
-    args.lr = args.lr*float(args.batch_size*args.world_size)/256. 
+    args.lr = args.lr*float(args.batch_size*args.world_size)/float(args.batch_size)
     optimizer = torch.optim.SGD(master_params, args.lr,
                                 momentum=args.momentum,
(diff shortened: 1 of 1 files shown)


examples/imagenet/main.py
```diff
@@ -156,7 +156,7 @@ def main():
     criterion = nn.CrossEntropyLoss().cuda()
 
     # Scale learning rate based on global batch size
-    args.lr = args.lr*float(args.batch_size*args.world_size)/256. 
+    args.lr = args.lr*float(args.batch_size*args.world_size)/float(args.batch_size)
```
mcarilli: I think the existing formula (using a fixed denominator of 256) actually does provide the correct scaling.   I believe learning rate should be scaled based on the global batch size; notice that `args.batch_size` is in the numerator.

Why do you think the existing formula is wrong?

jykzel: Could you explain where the fixed denominator of 256 comes from?

mcarilli: There's nothing special about 256 in particular.  Originally, we tested end-to-end convergence with a batch size of 256 per process, so for convenience, when I added logic to handle arbitrary batch and world sizes, I decided to scale the learning rate relative to the batch size we had just tested (256).  I could move the factor elsewhere instead (for example, divide the default learning rate by 256).  I don't think there is any harm keeping it where it is though.

PR #173, opened 2019-03-01
Unified mixed precision API + backend performance improvements

mcarilli:
PRing for comparison.  Single-GPU and distributed tests pass for the full cross product of supported pure and mixed precision options.  Any code using the old API should also continue to work properly for now, although it will be discouraged and eventually removed.

Documentation still needs some work.


apex/fp16_utils/__init__.py
```diff
@@ -14,3 +14,5 @@
 
 from .fp16_optimizer import FP16_Optimizer
 from .loss_scaler import LossScaler, DynamicLossScaler
+
+test = 1
```
mcarilli: delete

dave-epstein:
Once this is merged, is there a recommended process for porting over old code that uses FP16_Optimizer?

mcarilli:
@dave-epstein of course!  Working on that now.  It‘ll be really easy.

dave-epstein:
> @dave-epstein of course! Working on that now. It‘ll be really easy.

Great! Thanks so much, this is a fantastic library :)

mcarilli:
@dave-epstein I'm still working on making the documentation more comprehensive but in the meantime you can check out the reworked Imagenet example https://github.com/NVIDIA/apex/tree/api_refactor/examples/imagenet.

`--opt-level O2` is equivalent to FP16_Optimizer with dynamic loss scaling. (In fact, it calls directly into FP16_Optimizer under the hood.  Once I get everyone everyone on the new API I can stop maintaining FP16_Optimizer as a separate entity.)

PR #176, opened 2019-03-03
Bug fix in next power of 2

mkolod:
The existing "next power of 2" implementation produces incorrect results in some cases. The one identified so far is that for every value less than a power of 2 by 1 keeps value the same, e.g. h_next_pow2(7) produces 7, etc. For numbers farther away, the output is correct, e.g. h_next_pow2(26) gives 32.

The fix is identical to the "popular" way of doing this, i.e. subtracting 1 before the bit shifts, doing bitwise ORs with the bit-shifted values, and then adding 1 again, as in [here](https://graphics.stanford.edu/~seander/bithacks.html#RoundUpPowerOf2) for instance.

Note that the [earlier commit](https://github.com/NVIDIA/apex/blame/8bd382fadd42f77922fbc3dfacc3e4cc2f89ece5/csrc/welford.cu#L22:L29) was also wrong because it didn't subtract 1 at the beginning.

### DIFF
--- a/csrc/welford.cu
+++ b/csrc/welford.cu
@@ -20,13 +20,13 @@ __device__ __forceinline__ int lastpow2(int n)
 }
 
 __host__ __forceinline__ int h_next_pow2(unsigned int n) {
-    unsigned int old = n;
+    n--;
     n |= (n >>  1);
     n |= (n >>  2);
     n |= (n >>  4);
     n |= (n >>  8);
     n |= (n >> 16);
-    return n == old? n : n + 1;
+    return ++n;
 }
 
 __host__ __forceinline__ int h_last_pow2(unsigned int n) {


mcarilli:
You can also use the __clz intrinsic, like I did to obtain the last (as opposed the next) power of 2 in [THCReduce](https://github.com/pytorch/pytorch/blob/2336f0ba062d3c00117ccefc49c07bf2626284f1/aten/src/THC/THCReduce.cuh#L36).  @jjsjann123, what do you want to do here?

mkolod:
@mcarilli Yep, `__clz` works too (and I'd argue better), I guess my main point was to fix the existing bug. :)

ngimel:
Interestingly, the version in aten doesn't use __clz, and also returns last power of 2 less than or equal to n, as opposed to similarly called version in THC that returns last power of 2 less than n. https://github.com/pytorch/pytorch/blob/d313b582fd9ab83329fd5e4efcf2fb6374e7d068/aten/src/ATen/native/cuda/Reduce.cuh#L28-L36

jjsjann123:
The Aten one is calling last_pow2, which is equivalent to https://github.com/NVIDIA/apex/blame/8bd382fadd42f77922fbc3dfacc3e4cc2f89ece5/csrc/welford.cu#L31-L38

I'm cool with intrinsic function. Will update tomorrow. Thanks for pointing out the issue @mkolod

jjsjann123:
Wait, this is host code...
I'm pulling Marek's update here @mcarilli

PR #182, opened 2019-03-06
Remove LoadLibrary half norm hack

FDecaYed:
Let's remove it since we now have great torch.norm.
Also currently version will fail in multi-gpu case since context will end up on GPU 0. this change also fix that.


ngimel:
It suddenly became a very large diff. I assume the only thing that is really happening is deletion of norm?

FDecaYed:
I merged with master using in github web page and somehow this happens. I think it is just white spaces?
https://github.com/NVIDIA/apex/pull/182/files?utf8=%E2%9C%93&diff=split&w=1

ngimel:
Thanks!

PR #184, opened 2019-03-07
Add "dynamic" argument to loss scaler

### LINKED ISSUE #183: Failing optim_wrapper due to missing Scaler argument
When creating an optimizer and wrapping it via ``amp_handle.wrap_optimizer(optim)``, the handle [`calls the OptimWrapper`](https://github.com/NVIDIA/apex/blob/master/apex/amp/handle.py#L148), who wraps the optimizer and tries to instantiate a loss scaler per loss. 

The `OptimWrapper` tries to [instantiate the loss-scaler without an argument](https://github.com/NVIDIA/apex/blob/master/apex/amp/opt.py#L16), but the loss-scaler [needs an argument `loss_scale`](https://github.com/NVIDIA/apex/blob/master/apex/amp/scaler.py#L28), which causes the whole wrapping process to fail.

justusschock:
Adds missing argument to `LossScaler`.

Would fix #183 

Alternative would be to make this a default argument in the scaler.

### DIFF
--- a/apex/amp/opt.py
+++ b/apex/amp/opt.py
@@ -13,7 +13,7 @@ class OptimWrapper(object):
         self._num_loss = num_loss
         self._loss_idx = 0
         self._skip_next = [False] * num_loss
-        self._loss_scaler = [LossScaler() for _ in range(num_loss)]
+        self._loss_scaler = [LossScaler('dynamic') for _ in range(num_loss)]
 
     @contextlib.contextmanager
     def scale_loss(self, loss):


mcarilli:
I just merged a massive API refactor that dramatically improves versatility and also (in many cases) moderately improves performance:
https://nvidia.github.io/apex/amp.html#apex-amp
The OptimWrapper API is deprecated.  I can merge this to give people time to refactor, but everyone is strongly recommended to move to the new unified API.  It should be easy:
https://nvidia.github.io/apex/amp.html#transition-guide-for-old-api-users

Multiple optimizers are supported:
https://nvidia.github.io/apex/amp.html#module-apex.amp

Issue #185, opened 2019-03-07
Apex fails to build due to "no matching function" and "cannot convert"


jah3xc:
I'm using Apex within the nvidia/cuda docker container. I was using commit 500939d04a863f78b2af3c72c123deee3609469d, and everything was working. Today I pulled and tried building it and it fails. 

Steps to reproduce:
1. Run `nvidia/cuda:9.0-cudnn7-devel-ubuntu16.04`
2. Install needed stuff:
```
apt update && apt install -y python3 python3-pip git
```
3. Install prereqs
```
pip3 install torch numpy
```
4. Clone and try to install
```
git clone https://github.com/nvidia/apex
cd apex
python3 setup.py install --cuda_ext --cpp_ext
```

The build fails:
```
/usr/local/lib/python3.5/dist-packages/torch/lib/include/pybind11/pytypes.h:1165:206: error: expansion pattern 'pybind11::detail::negation<std::is_same<pybind11::detail::bools<pybind11::detail::negation<std::is_base_of<pybind11::arg, Args> >::value ..., pybind11::detail::negation<std::is_same<pybind11::detail::kwargs_proxy, Args> >::value ..., true>, pybind11::detail::bools<true, pybind11::detail::negation<std::is_base_of<pybind11::arg, Args> >::value ..., pybind11::detail::negation<std::is_same<pybind11::detail::kwargs_proxy, Args> >::value ...> > >::value' contains no argument packs
/usr/local/lib/python3.5/dist-packages/torch/lib/include/pybind11/pytypes.h:1165:215: error: template argument 1 is invalid
/usr/local/lib/python3.5/dist-packages/torch/lib/include/pybind11/pytypes.h:1165:392: error: expansion pattern 'pybind11::detail::negation<std::is_same<pybind11::detail::bools<pybind11::detail::negation<std::is_base_of<pybind11::arg, Args> >::value ..., pybind11::detail::negation<std::is_same<pybind11::detail::kwargs_proxy, Args> >::value ..., true>, pybind11::detail::bools<true, pybind11::detail::negation<std::is_base_of<pybind11::arg, Args> >::value ..., pybind11::detail::negation<std::is_same<pybind11::detail::kwargs_proxy, Args> >::value ...> > >::value' contains no argument packs
/usr/local/lib/python3.5/dist-packages/torch/lib/include/pybind11/pytypes.h:1165:395: error: template argument 2 is invalid
[...]

PR #190, opened 2019-03-10
fix includes

ngimel:
To compile with cuda 9.


ngimel:
Should fix #185 and #150

PR #197, opened 2019-03-13
Wkong/api refactor

alpha0422:
FusedAdam with multi-tensor apply support.


mcarilli:
This will be folded into master via the multi_tensor_sgd branch, which is still WIP.

PR #205, opened 2019-03-15
Anticipating upstream #17996

mcarilli:
Effectively reverts https://github.com/NVIDIA/apex/commit/42180bd929639c137706be99c93626da93a36f2e, because once upstream merges https://github.com/pytorch/pytorch/pull/17996, the shim hack for type-based dispatch in extensions will no longer be necessary (and in fact will break).  However, with current upstream master, the shim hack is still necessary, so I can't merge this into Apex master until 17996 is merged upstream.


cliffwoolley:
I confirm that this builds successfully with the now-merged 17996.  (Build 5866676.)

PR #207, opened 2019-03-17
More permissive inputs to forward function

gpapan:
Allow more permissive inputs to forward function:
(1) When some input is a CPU pytroch tensor, show a one-time warning instead of printing it every time the forward function is called.
(2) Allow numpy arrays as (nested) inputs. Currently this fails.

### DIFF
--- a/apex/amp/_initialize.py
+++ b/apex/amp/_initialize.py
@@ -1,6 +1,8 @@
 import torch
 from torch._six import string_classes
 import functools
+import numpy as np
+import warnings
 from ._amp_state import _amp_state, warn_or_err, container_abcs
 from .handle import disable_casts
 from .scaler import LossScaler
@@ -15,7 +17,7 @@ def to_type(dtype, t):
     if isinstance(t, torch.Tensor):
         if not t.is_cuda:
             # This should not be a hard error, since it may be legitimate.
-            print("Warning:  An input tensor was not cuda. ")
+            warnings.warn("An input tensor was not cuda.")
         # GANs require this.
         # if t.requires_grad:
         #     warn_or_err("input data requires grad.  Since input data is not a model parameter,\n"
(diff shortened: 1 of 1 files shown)


mcarilli:
The warnings.warn is a great change, I should have gone that route to begin with.

Is the ndarray something that you intend to cast manually within your model?  That could be a problem.  You might have to change your manual cast depending on the opt_level.

gpapan:
In my setup the np.ndarray carries instance-specific parameters about the loss, so I do not need to cast it differently based on the opt_level. Right now this code path fails if any of the inputs is an np.ndarray, so I propose to just pass its value unchanged similar to how strings or primitives (int, float etc) are handled. What do you think?

mcarilli:
The input-casting's final case is already set up to return any types it doesn't recognize without attempting to cast them (https://github.com/NVIDIA/apex/blob/master/apex/amp/_initialize.py#L43-L52)
so your ndarray should be getting through.  What failures were you observing?

gpapan:
Unfortunately it never reaches https://github.com/NVIDIA/apex/blob/master/apex/amp/_initialize.py#L43-L52. The error happens at https://github.com/NVIDIA/apex/blob/master/apex/amp/_initialize.py#L39-L40
because ndarray is an iterable and it fails with a TypeError:

> import numpy as np
> xs = np.zeros((3, 4))
> type(xs)(x for x in xs)
  File "<stdin>", line 1, in <module>
TypeError: expected sequence object with len >= 0 or a single integer

mcarilli:
Oh i see, it fails because ndarray IS an iterable.  I'm still not entirely comfortable with forwarding ndarrays straight through, though.  Your use case does not require casting the input depending on `opt_level`, but I can imagine someone else might pass an ndarray as input, which they later (within forward) convert to a Tensor.  The new Tensor's type would then be the same regardless of the surrounding `opt_level`, and may or may not be correct.  I guess forwarding the ndarrays is philosophically no different from forwarding any type I can't recognize, or that does not have a user-supplied `to` method, and I can't stop people from shooting themselves in the foot that way...

PR #217, opened 2019-03-22
Add prelu to list of torch overrides

henrymai:
This is to fix the following error:

  File "/opt/conda/lib/python3.6/site-packages/torch/nn/modules/module.py", line 489, in __call__
    result = self.forward(*input, **kwargs)
  File "/opt/conda/lib/python3.6/site-packages/torch/nn/modules/container.py", line 92, in forward
    input = module(input)
  File "/opt/conda/lib/python3.6/site-packages/torch/nn/modules/module.py", line 489, in __call__
    result = self.forward(*input, **kwargs)
  File "/opt/conda/lib/python3.6/site-packages/torch/nn/modules/activation.py", line 722, in forward
    return F.prelu(input, self.weight)
  File "/opt/conda/lib/python3.6/site-packages/torch/nn/functional.py", line 1040, in prelu
    return torch.prelu(input, weight)
RuntimeError: expected scalar type Half but found Float

### DIFF
--- a/apex/amp/lists/torch_overrides.py
+++ b/apex/amp/lists/torch_overrides.py
@@ -12,7 +12,6 @@ FP16_FUNCS = [
     'conv_transpose2d',
     'conv_transpose3d',
     'conv_tbc',
-    'prelu',
 
     # BLAS
     'addmm',
@@ -71,6 +70,7 @@ CASTS = [
     'addcmul',
     'atan2',
     'cross',
+    'prelu',
 
     # Element-wise _or_ tensor-wise math
     'add',
--- a/apex/amp/lists/torch_overrides.py
+++ b/apex/amp/lists/torch_overrides.py
@@ -12,6 +12,7 @@ FP16_FUNCS = [
     'conv_transpose2d',
     'conv_transpose3d',
     'conv_tbc',
+    'prelu',
 
     # BLAS
     'addmm',


mcarilli:
Adding things to the various lists is partially based on expectation, but also based on experience, so thank you for this.

Prelu should be safe in either fp16 or fp32, the only tricky thing is that it has two arguments, and they must match.  Amp has a third list for that, the list called CASTS, which casts multiple inputs to the widest incoming type.  Therefore, if both incoming arguments are fp16, the function will still run in fp16, but if either incoming argument is fp32, both will be promoted to fp32.

I will move prelu to the CASTS list.

henrymai:
Yeah, I initially did think about putting it under CASTS.
I decided against it since PReLU as you said is fine in fp16 and it also worked well in testing, so I thought it would be fine to just always cast to fp16.
Thanks for merging the change.

mcarilli:
Amp's policy in general for "neutral" functions that are fine in either precision is to just take the inputs as they come.  Having prelu as a CASTS function is more in keeping with that philosophy.  Does putting it under CASTS also resolve your issue?

henrymai:
Yeah, having it under CASTS works fine for me too.

henrymai:
Actually, sorry, I spoke too soon last time.

I just pulled and installed the latest version of Apex. Having prelu under CASTS resolves the exception that I got, but now I'm seeing a major slow down with my model.

The problem here is that in order for Apex amp to provide a speedup for the normal use of torch.prelu (via the torch.nn.PReLU layer), one would have to manually cast the internal `.weight` parameter inside of the torch.nn.PReLU layer to fp16 or define a custom PReLU layer that has a fp16 `.weight` parameter.

Philosophically, I think this is the same as the torch.conv* functions so I think it makes more sense for prelu to be in the FP16_FUNCS list rather than CASTS, since amp relies on casting the parameters rather than the weights of a model.

henrymai:
I just made a new pull request and updated the comment inside the FP16_CASTS list.
https://github.com/NVIDIA/apex/pull/257

PR #218, opened 2019-03-22
Fix 'local variable 'optimizers_was_list' referenced before assignmen…

enricoschroeder:
…t' when amp.initialize() is called with optimizers=None


mcarilli:
Thanks!  I implemented the optimizers=None option on the fly at GTC as it was requested by someone in person.  I was going to try it today but it looks like you got there first...

PR #232, opened 2019-03-31
WIP:  Handle arbitrary combinations of optimizers/models/losses

mcarilli:
Handle arbitrary combinations of optimizers/models/losses.  Do not maintain a loss scale per optimizer.  Instead, maintain either a single global loss scale (by default), or a loss scale per-loss (if the user supplies the `num_loss` argument to amp.initialize and the `loss_id` argument to each invokation of amp.scale_loss).

Current status:  L0 tests have been added and are passing (maybe need more).  L1 single-GPU tests pass.  Verified that it fixes Szymon's issue with master params getting out of sync across processes for JoC GNMT.  L1 multi-GPU tests underway.

### DIFF
--- a/apex/amp/_initialize.py
+++ b/apex/amp/_initialize.py
@@ -6,6 +6,7 @@ import warnings
 from ._amp_state import _amp_state, warn_or_err, container_abcs
 from .handle import disable_casts
 from .scaler import LossScaler
+from ._process_optimizer import _process_optimizer
 from apex.fp16_utils import convert_network
 from ..fp16_utils import FP16_Optimizer as FP16_Optimizer_general
 from ..optimizers import FP16_Optimizer as FP16_Optimizer_for_fused
@@ -122,7 +123,7 @@ def wrap_fused_adam(optimizer, properties):
         return FP16_Optimizer_for_fused(optimizer, static_loss_scale=properties.loss_scale)
 
 
-def _initialize(models, optimizers, properties):
+def _initialize(models, optimizers, properties, num_losses=1):
     from apex.parallel import DistributedDataParallel as apex_DDP
     from .amp import init as amp_init
 
@@ -146,7 +147,8 @@ def _initialize(models, optimizers, properties):
 
     check_models(models)
 
-    check_params_fp32(models)
+    if not _amp_state.allow_incoming_model_not_fp32:
+        check_params_fp32(models)
     
     check_optimizers(optimizers)
 
@@ -181,21 +183,16 @@ def _initialize(models, optimizers, properties):
         for optimizer in optimizers:
             optimizer.load_state_dict(optimizer.state_dict())
 
-    if properties.master_weights:
-        for i, optimizer in enumerate(optimizers):
-            if isinstance(optimizer, FusedAdam):
-                optimizers[i] = wrap_fused_adam(optimizer, properties)
-            if properties.loss_scale == "dynamic":
-                optimizers[i] = FP16_Optimizer_general(optimizer,
-                                                       dynamic_loss_scale=True,
-                                                       verbose=False)
-            else:
-                optimizers[i] = FP16_Optimizer_general(optimizer,
-                                                       static_loss_scale=properties.loss_scale,
-                                                       verbose=False)
-    else:
-        for optimizer in optimizers:
-            optimizer.loss_scaler = LossScaler(properties.loss_scale)
+    for i, optimizer in enumerate(optimizers):
+        # Still need to special case this for the first pass
+        if isinstance(optimizer, FusedAdam):
+            optimizers[i] = wrap_fused_adam(optimizer, properties)
+        else:
+            optimizers[i] = _process_optimizer(optimizer, properties)
+
+    _amp_state.loss_scalers = []
+    for _ in range(num_losses):
+        _amp_state.loss_scalers.append(LossScaler(properties.loss_scale))
 
     if properties.patch_torch_functions:
         # handle is unused here. It's accessible later through a global value anyway.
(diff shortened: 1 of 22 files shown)


hukkelas:
I tried this branch earlier today to include different loss scales for each loss in my progressive growing GAN, but I had issues getting it working correctly. What I ended up with was something similar to this:

```python
d_optimizer.zero_grad()
with amp.scale_loss(wasserstein_distance, self.d_optimizer, loss_id=0) as scaled_loss:
                    scaled_loss.backward(retain_graph=True)
with amp.scale_loss(epsilon_penalty, self.d_optimizer, loss_id=1) as scaled_loss:
                    scaled_loss.backward(retain_graph=True)
with amp.scale_loss(gradient_penalty, self.d_optimizer, loss_id=2) as scaled_loss:
                    scaled_loss.backward()
d_optimzier.step()
```

This seems to be working with a single GPU, but when I use apex DistributedDataParallel (2 GPU's) it is very unstable, and I have no idea why :/ 


I also took a look at the [advanced AMP usage](https://nvidia.github.io/apex/advanced.html#multiple-backward-passes-per-iteration) for multiple backward passes for a single optimizer, where it says that you should use `delay_unscale=True`. I tried this (by having `delay_unscale=True` for the first two backward passes), but it seems like it was using the last loss's loss_scaler to unscale all the grads. So the model was diverging immediately.

mcarilli:
Bold of you to test it already...That usage looks correct.  I'm glad to hear it's working on single-GPU and that the new syntax enables your use case.  I'm still trying it with multi-GPU.

I haven't updated documentation for the new branch, but guidance will be that using `delay_unscale` is no longer necessary and should only be used as a minor performance optimization if you're really sure what you're doing.

mcarilli:
@hukkelas I notice you're using gradient penalty.  I don't think Apex DDP supports gradient penalty properly, so your instability may not be the fault of the new branch.  I still haven't finished testing the new branch with distributed, but it seems to work ok so far.  Can you try `torch.nn.parallel.DistributedDataParallel` instead?  Replace
```
model, optimizer = amp.initialize(model, optimizer, opt_level=XX, num_losses=YY)
model = apex.parallel.DistributedDataParallel(model)
```
with
```
model, optimizer = amp.initialize(model, optimizer, opt_level=XX, num_losses=YY)
model = torch.nn.parallel.DistributedDataParallel(model,
    device_ids=[args.local_rank],
    output_device=args.local_rank)
```

hukkelas:
Hmm, didn't know that, thanks! 

I was using Apex DDP, switched to `torch.nn.parallel.DistributedDataParallel` now. From some quick tests it looks better.

I just recently started using multi-gpu training for my GAN. I think most of my previous instability, with a single GPU, has come from the gradient penalty calculation. I made a hacky solution to introduce independent loss scaling for it previously, but this branch would remove a lot of ugly code :)

mcarilli:
Glad it appears to be working now.  My local distributed tests/tests on local models have also succeeded so far so I'll probably merge this branch today.  

When you say you are scaling the gradient penalty calculation, is this the line you are referring to:
```
with amp.scale_loss(gradient_penalty, self.d_optimizer, loss_id=2) as scaled_loss:
                    scaled_loss.backward()
```
? Scaling the loss for this call^ is fine.

Or are you referring to an earlier call to `torch.autograd.grad(...., create_graph=True, only_inputs=True)` to create out-of-place gradients used to construct the `gradient_penalty` scalar?  Scaling the loss for that particular call would not be supported by the new branch either (Amp has no visibility of the newly created out-of-place gradients, so it has no way to check them for infs).

hukkelas:
Before this branch I had independent scaling on the gradient penalty like this:
```python
```

hukkelas:
Before this branch, I independently scaled my gradient penalty like this: 
```python
logits = discriminator(x_hat, condition, landmarks)
logits = logits.sum() * loss_scaler.loss_scale()
grad = torch.autograd.grad(
    outputs=logits,
    inputs=x_hat,
    grad_outputs=torch.ones(logits.shape).to(fake_data.dtype).to(fake_data.device),
    create_graph=True
)[0] 
grad = grad.view(x_hat.shape[0], -1)
if check_overflow(grad):
    print("Overflow in gradient penalty calculation.")
    loss_scaler._loss_scale /= 2
    print("Scaling down loss to:", loss_scaler._loss_scale)
    return None
grad = grad / loss_scaler.loss_scale()

grad_penalty = ((grad.norm(p=2, dim=1) - 1)**2)
```
`check_overfow` is doing the same checks as `scale_check_overflow_python` in [scaler.py](https://github.com/NVIDIA/apex/blob/8402878650eeca3cad3d01c1c7ecb80e9a0cb1fc/apex/amp/scaler.py#L6). If there is an overflow, I just skip the current batch for all GPUs. 

From my understanding, since AMP has no visibility of the `torch.autograd.grad` call, I should keep this independent scaling of the gradient penalty, and also include the independent loss scaling as I showed earlier?

apex/amp/_initialize.py
```diff
@@ -181,21 +183,16 @@ def new_fwd(*args, **kwargs):
...
-        for optimizer in optimizers:
-            optimizer.loss_scaler = LossScaler(properties.loss_scale)
+    for i, optimizer in enumerate(optimizers):
+        # Still need to special case this for the first pass
+        if isinstance(optimizer, FusedAdam):
+            optimizers[i] = wrap_fused_adam(optimizer, properties)
+        else:
+            optimizers[i] = _process_optimizer(optimizer, properties)
+
+    _amp_state.loss_scalers = []
```
hukkelas: Should there be a assert / check that _amp_state.loss_scalers is None? Got a very confusing error message in backward if I called `amp.initialize` a second time for a network without an optimizer.

Found  a work around for it, but its still confusing to find the error for this.

mcarilli: My intention is that amp.initialize should be called only once.  Why do you need to call it multiple times?

Also, the scaling of the out-of-place backward to create the grads for the gradient penalty is always going to be tricky...can you start an issue where we can track that and figure out best practices?

PR #234, opened 2019-03-31
fix: minor print typo

hukkelas:
Prints
```bash
Gradient overflow.  Skipping step, loss scaler 3 reducing loss scale to 32768.0
````
instead of 
```bash
Gradient overflow.  Skipping step, loss scaler {} reducing loss scale to 3
```


mcarilli:
Good catch, thanks!

PR #241, opened 2019-04-04
Run interpolation in fp32 because it's faster

mkolod:
Running interpolation (nearest neigbor and bilinear) in fp16 is slower than fp32, because the backprop depends on individual half atomicAdds (as opposed to half2 atomicAdds), which actually translates to atomicCAS under the hood. For networks that heavily depend on upsampling (any kind of generative CV models, say GANs), the overall impact of using fp16 upsampling results in up to 30% end-to-end training slowdowns. 

The long-term fix is to fix upsampling (I'm planning to make PRs to this effect), but for now I think it's better to run upsampling in fp32 instead.

### DIFF
--- a/apex/amp/lists/functional_overrides.py
+++ b/apex/amp/lists/functional_overrides.py
@@ -27,6 +27,10 @@ FP16_FUNCS = [
 ]
 
 FP32_FUNCS = [
+
+    # Interpolation/Upsampling
+    'interpolate',
+
     # Pointwise
     'softplus',
     'softmin',


mcarilli:
I was planning to do this, but I wasn't prioritizing it yet because I thought it might take a long time to track down the various Python exposures that might trigger the offending kernel(s).  After seeing your PR, I can verify that upsampling Modules, upsampling functional calls, and interpolation calls all appear to route through torch.nn.functional.interpolate, so patching that should be sufficient.  Thanks!

PR #252, opened 2019-04-08
Add support for CPU and sparse tensors

azgo14:
1) Added CPU tensor support (all cases with cuda.FloatTensor and cuda.HalfTensor are replaced with also allowing FloatTensor and HalfTensor. Whether tensor is on CPU vs GPU should make no difference algorithmically).

2) Added sparse tensor support (overflow checks need special case for Sparse tensors since previous logic of `float(model_grad.float().sum())` does not work for sparse). For additional safety since sparse tensors cannot be distributed, I added an option to explicitly `all_reduce_overflow` to ensure all nodes receive the same overflow state per update. Since some parameters are not distributed for sparse use-cases (Pytorch doesn't support DistributedDataParallel with sparse tensors), it's possible some nodes can have inconsistent overflow states with other nodes which can un-sync their parameter updates after loss scaling.

### DIFF
--- a/apex/amp/_initialize.py
+++ b/apex/amp/_initialize.py
@@ -75,15 +75,17 @@ def check_models(models):
 def check_params_fp32(models):
     for model in models:
         for name, param in model.named_parameters():
-            if param.is_floating_point() and param.type() != "torch.cuda.FloatTensor":
-                warn_or_err("Found param {} with type {}, expected torch.cuda.FloatTensor.\n"
+            if param.is_floating_point() and (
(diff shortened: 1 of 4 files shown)


mcarilli:
This is great!  The main shortcoming right now is that I think it will only work for a Python-only install of Apex.

If Apex is installed with the cpp extensions, it performs gradient unscaling using my custom multi-tensor kernels, which can only handle contiguous tensors.  I can and will hack that logic to accept sparse tensors as well, but I need to clean up handling of fused optimizers and come up with a unified strategy for checkpointing first.

PR #257, opened 2019-04-11
prelu belongs in FP16_CASTS

henrymai:
prelu belongs in FP16_CASTS

The main use of these functions (e.g.: `torch.{conv*, prelu}`) is via their `torch.nn`
wrapping layers.

The `torch.nn` layers are what contain the weights and call into these lower level
functions with the weights as a parameter in their `forward()` method.

The `torch.conv*` functions are already in the `FP16_CASTS` list due to amp's philosophy of
casting the parameters rather than the model/layer weights.

Conceptually `torch.prelu` is the same as the `torch.conv*` case, where its weight parameter
is passed in from its wrapper layer `torch.nn.PReLU`.

### DIFF
--- a/apex/amp/lists/torch_overrides.py
+++ b/apex/amp/lists/torch_overrides.py
@@ -5,8 +5,9 @@ from .. import utils
 MODULE = torch
 
 FP16_FUNCS = [
-    # Math
-    # TODO: why are these in top-level torch namespace?
+    # Low level functions wrapped by torch.nn layers.
+    # The wrapper layers contain the weights which are then passed in as a parameter
+    # to these functions.
     'conv1d',
     'conv2d',
     'conv3d',
@@ -14,6 +15,7 @@ FP16_FUNCS = [
     'conv_transpose2d',
     'conv_transpose3d',
     'conv_tbc',
+    'prelu',
 
     # BLAS
     'addmm',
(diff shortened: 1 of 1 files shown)


mcarilli:
Splitting hairs here:  I tend to use the terms "parameters" and "weights" interchangeably, since the "Parameters" attributes of a module are the weights.

Amp's philosophy with O1 is to cast the incoming arguments to torch functions, rather than the weight (Parameter) attributes of the wrapping Modules.  I think I see what you mean.  Since torch.prelu takes 2 arguments, and these argument types must be made to match, I originally registered it as a [CASTS function](https://github.com/NVIDIA/apex/blob/master/apex/amp/lists/torch_overrides.py#L73-L79) meaning that it casts both inputs to match the widest incoming type.  In this case, because (with O1) the weight is FP32, the input data will always be cast to FP32 to match the weight, even though the operation is probably perfectly safe in FP16.

I'm surprised the upcasting of input data here is having such a large performance impact on your model as a whole, but I'm fine with this change.  FYI putting torch.prelu in FP16_FUNCS will not eliminate casting entirely (if the incoming data is FP16, the FP32 weight will have to be cast to FP16 on entrance to the patched prelu).  If you supplied a large value for num_parameters and therefore have a large weight tensor, this may also be problematic.

henrymai:
Prior to pulling master down and installing it, I was using my own version of amp which still had prelu in FP16_CASTS.
Once I installed master, I noticed the memory usage go up and my iterations taking longer than before.
I modified master locally again to have prelu under FP16_CASTS and the memory consumption went down and the iterations were faster again.

PR #268, opened 2019-04-23
move and fix check_optimizers

ptrblck:
Fix `check_optimizers` using `bad_optim_type` and move the check up so that it the appropriate error will be thrown.
Currently passing an instance of `FP16_Optimizer` will trigger the `else` statement [here](https://github.com/NVIDIA/apex/blob/28097c9999d86bd889a1f03c963e29e3384f3996/apex/amp/_initialize.py#L151), so that the error might be confusing stating "optimizers must be either a single optimizer or a list of optimizers."

### DIFF
--- a/apex/amp/_initialize.py
+++ b/apex/amp/_initialize.py
@@ -111,7 +111,7 @@ def check_optimizers(optimizers):
         if isinstance(optim, FP16_Optimizer_for_fused):
             bad_optim_type = "apex.optimizers.FP16_Optimizer"
         if bad_optim_type is not None:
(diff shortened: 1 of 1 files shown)


ptrblck:
Code to trigger the right warning in this PR:
```python
import torch
import torch.nn as nn
import torch.optim as optim

from apex import amp
from apex.fp16_utils import FP16_Optimizer


model = nn.Linear(1, 1).to('cuda')
optimizer = optim.SGD(model.parameters(), lr=1e-3)
optimizer = FP16_Optimizer(optimizer)

model, optimizer = amp.initialize(model, optimizer, opt_level='O1')
```

mcarilli:
Thanks!

PR #272, opened 2019-04-24
Replace type().ScalarType() with scalar_type()

### LINKED ISSUE #267: error: class "at::Type" has no member "scalarType"
I can't setup this, error is:
    csrc/type_shim.h(13): error: class "at::Type" has no member "scalarType"
    1 error detected in the compilation of "/tmp/tmpxft_00004318_00000000-6_multi_tensor_scale_kernel.cpp1.ii".
    error: command '/usr/local/cuda/bin/nvcc' failed with exit status 1
who know how to fix it?

ptrblck:
Fixes #267 

PyTorch PR https://github.com/pytorch/pytorch/pull/17991 introduced some breaking changes which broke building `apex` using PyTorch master.

This PR replaces all deprecated `.type().ScalarType()` calls with `.scalar_type()` as well as `at::ScalarType::Type` with `at::kType`.

### DIFF
--- a/csrc/type_shim.h
+++ b/csrc/type_shim.h
@@ -10,19 +10,19 @@ struct TypeShim
   // Enable trivial conversion to a const at::Type& for pre-3aeb78
   operator const at::Type&(){ return payload; };
   // Enable dispatch switch statements to take *this directly for  post-3aeb78
-  operator at::ScalarType(){ return payload.scalarType(); };
+  //operator at::ScalarType(){ return payload.; };
 };
 
 #define DISPATCH_FLOAT_AND_HALF(TYPE, LEVEL, NAME, ...) \
   switch(TYPE) \
   { \
-    case at::ScalarType::Float: \
+    case at::kFloat: \
     { \
       using scalar_t_##LEVEL = float; \
       __VA_ARGS__; \
       break; \
     } \
-    case at::ScalarType::Half: \
+    case at::kHalf: \
     { \
       using scalar_t_##LEVEL = at::Half; \
       __VA_ARGS__; \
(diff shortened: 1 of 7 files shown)


Stonesjtu:
Hi, I think we should tag the current master as a compatible version for the latest pytorch stable release.

csrc/type_shim.h
```diff
@@ -3,26 +3,26 @@
...
+//   operator const at::Type&(){ return payload; };
+//   // Enable dispatch switch statements to take *this directly for  post-3aeb78
+//   //operator at::ScalarType(){ return payload.; };
+// };
 
 #define DISPATCH_FLOAT_AND_HALF(TYPE, LEVEL, NAME, ...) \
   switch(TYPE) \
   { \
-    case at::ScalarType::Float: \
+    case at::kFloat: \
```
mcarilli: I feel more comfortable with the existing `at::ScalarType::Float` syntax because that's what the upstream dispatch macros currently use https://github.com/pytorch/pytorch/blob/master/aten/src/ATen/Dispatch.h

ptrblck: I've changed it back to the `at::ScalarType::Type` usage for most of the code. I only left the usage of `at::kType` for creating new tensors (in case `tensor.scalar_type()` wasn't used), as it seems to be the recommended way ([link](https://github.com/pytorch/pytorch/blob/c42f3f9055b1ead5253fb99980bfaf46ecafcb0f/c10/core/TensorOptions.h#L49)).

Would this be alright or should I revert all usages of `at::kType`?

mcarilli: I feel most comfortable mirroring what upstream does, so this seems fine

PR #274, opened 2019-04-26
Persistent group batch normalization

jjsjann123:
Persistent group batchnorm added

Added persistent grouped batch norm for performance run on strong scaling case:
currently only supporting:

      1. nhwc layout
      2. fp16
      3. synchronization only within a node!

Potential issue: relying on unstable internal `_share_cuda_()` to get ipc mem handle.

Environment variable is used to tune LAUNCH_MARGIN that limits the CTAs usage
by the persistent kernel.

Documentation and examples will follow.

* Also fused BNAddRelu for apex::parallel::SyncBatchNorm nhwc kernel.


mcarilli:
This will be folded in from the multi_tensor_sgd branch once that's merged.

Issue #254, opened 2019-04-10
sync_batchnorm (python version) doesn't work in FP16


xuyuan:
RuntimeError: expected type torch.cuda.FloatTensor but got torch.cuda.HalfTensor at line
https://github.com/NVIDIA/apex/blob/03100f466145a0de73b79c32d11d58c917b10393/apex/parallel/sync_batchnorm.py#L108

mcarilli:
@jjsjann123 can you follow up when you get a chance?

jjsjann123:
@xuyuan Any reason why you are not using the optimized kernel implementation?
You could try that by installing your apex with `--cuda_ext` attached to the end (`python setup.py install --cuda_ext`). That implementation has much better perf and more test coverage :)

I'll take a look later this weekend. Thanks for pinging me on this @mcarilli

xuyuan:
@jjsjann123 thanks for your suggestion. But the cpp_ext/cuda_ext kernel converge slower than pure python version in my experiments. I am using apex.dist + mixed precision (amp O2) training UNet like model for segmentation.  

The error is due to BN is kept in FP32, inputs are FP16.

jjsjann123:
Hmmm, the convergence issue sounds strange.
When you say pure python version converges faster, are you talking about torch.nn.BatchNormXd or the pure python implementation in apex?
Are you running the two comparison test in similar setup (same global_batch_size and local mini-batch_size for BN stats)?

jjsjann123:
So the issue is that pure python implementation doesn't support fp16 input for fp32 node, which is consistent with PyTorch convention, with the exception that cudnn batchnorm kernels are fine with it.

amp O2 casts everything to fp16 while leaving BN in fp32, but it's expecting CUDNN kernel that could accept fp16 input.
Our cpp kernel does handle the cast during data load, but pure python implementation doesn't do that.

Let me talk to Carilli to figure out a way to support this.

denru01:
I encountered the same issue when using 'O2' opt-level with `mapillary/inplace_abn'.

jjsjann123:
Sorry for the delayed response. Finally got some bandwidth and I'm looking at this right now.

jjsjann123:
Tried it with our image net example. Python version should work fine with amp o2 now.

PR #285, opened 2019-05-02
[SyncBatchNorm update]

jjsjann123:
resolves issue #254

Added input casting for pure python implementation, this supports mismatched
input and layer dtype.


jjsjann123:
FYI @mcarilli @ptrblck

minizon:
Hello, I think you forgot to "return" out in the forward function.

jjsjann123:
No idea how this passed unit test. Thanks for pointing it out @minizon

Issue #262, opened 2019-04-18
SyncBN raise ERROR when count = 1


EthanZhangYi:
It is right that BN layer in `torch.nn` raises error when `count = 1` (only 1 value per channel).

However, for SyncBN layer, even that `count = 1` for one GPU, it should NOT raise the error when `mean` and `var` are synchronized across multiple GPUs.

https://github.com/NVIDIA/apex/blob/cd2708ccdb66e15c94564e3f608a44aaa5d6a4ab/apex/parallel/optimized_sync_batchnorm_kernel.py#L29-L30

mcarilli:
@jjsjann123 can you take a look at this when you get a chance?

jjsjann123:
Thanks for reporting this.
Finally got some spare bandwidth and I'm looking at this right now.

PR #287, opened 2019-05-02
[syncbn update]

jjsjann123:
update input size check to fix github issue #262

update SyncBatchNorm count check so that size 1 input with cross GPU
synchronization runs fine.


jjsjann123:
FYI @mcarilli @ptrblck

PR #300, opened 2019-05-10
use verbose parameter to control print of grad overflow

michaelrglass:
Since gradient overflow is a normal part of dynamic loss scaling until the right loss scaling is established, it seems appropriate to gate the print statements with the 'verbose' parameter.

In distributed training, with dozens of processes all printing 'Grad overflow...' to the log, it can be tedious to scroll past. In this setting the verbose=True can be reserved for the process with world rank 0.


mcarilli:
This seems fine for now, although soon, FP16_Optimizer won't be used at all anymore by the backend.

PR #306, opened 2019-05-14
Enable LARC for use with amp

blisc:


mcarilli:
It looks like this requires a module "apex.parallel.LARC" but I don't see this among the diffs...am I missing something?

blisc:
LARC is already implemented in apex. It currently does not work with amp.
https://github.com/NVIDIA/apex/blob/master/apex/parallel/LARC.py
It works if you use the old method of
```
if args.fp16:
    optimizer = FP16_Optimizer(optimizer, dynamic_loss_scale=True)
optimizer = LARC(optimizer, trust_coefficient=0.001)
...
optimizer.optim.backward(loss)
```

mcarilli:
Oh gotcha, apex/parallel/LARC.py was "last edited July 3, 2018," and not added by me, and not part of Amp's core functionality that I've been working on, so I totally forgot about it.  It looks fairly agnostic to the structure of the optimizer and I think it will work with anything other than the current incarnation of FusedAdam.  I have a WIP branch which should enable FusedAdam to work with LARC as well.

PR #309, opened 2019-05-15
fix URLs in docs of apex.parallel

ptrblck:
Fixes minor errors in the URLs for [apex.parellel](https://nvidia.github.io/apex/parallel.html) pointing to old source code locations.

### DIFF
--- a/apex/parallel/distributed.py
+++ b/apex/parallel/distributed.py
@@ -104,8 +104,6 @@ class Reducer(object):
     When used with this launcher, :class:`Reducer` assumes 1:1 mapping of processes to GPUs.
     It also assumes that your script calls ``torch.cuda.set_device(args.rank)`` before creating the model.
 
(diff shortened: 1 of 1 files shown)


apex/parallel/distributed.py
```diff
@@ -104,7 +104,7 @@ class Reducer(object):
     When used with this launcher, :class:`Reducer` assumes 1:1 mapping of processes to GPUs.
     It also assumes that your script calls ``torch.cuda.set_device(args.rank)`` before creating the model.
 
-    main_reducer.py in https://github.com/NVIDIA/apex/tree/master/examples/imagenet shows example usage.
+    main_amp.py in https://github.com/NVIDIA/apex/tree/master/examples/imagenet shows example usage.
```
mcarilli: main_amp.py only shows DDP usage.  We don't have an example for Reducer, it's more of a tool for internal edge cases, and we don't intend external people to use it.

Issue #292, opened 2019-05-06
Adding new parameter groups to an optimizer


Tetratrio:
Hey,

I am currently training a growing model that requires me to add new param groups to the optimizer at each growth. That is, I train the model for X iterations, then add a new layer. This is repeated around 5-10 times. Preferably I keep the state of the optimizer for the older param groups when adding a new one.

Now I want to run this with mixed precision.
When using amp, there seem to be two options:
1. Add param group to optimizer where amp will take care of the operation. However, from my understanding, amp does not dynamically handle new parameter groups added to the optimizer?
2. Disable amp and restore the "pre-amp" state, add the param group and then reinitialize amp. I have only found a way to disable casts. If I try to reinitialize without first disabling amp, an error will be raised at https://github.com/NVIDIA/apex/blob/39e153a3159724432257a8fc118807b359f4d1c8/apex/amp/_process_optimizer.py#L257

I have not had time to go through all the code of this repository, but if I do I should probably be able to figure out how to manually reset amp. Although that would easily break with new apex releases and is not very elegant. So I thought I would check here first if there is an easy solution to my issue.

Thanks in advance!

P.S. I will be traveling for 2 days so I may not be able to respond to this issue.

mcarilli:
Are you adding layers by unfreezing them (ie by changing their requires_grad attributes from False to True, so that they begin to receive gradients) or via an `optimizer.add_param_group` call?  

Amp should properly treat unfreezing or unfreezing (switching from requires_grad=True to False or the other way around) for any parameters as long as those parameters were seen by the optimizer's constructor (i.e., before the call to amp.initialize).

Adding param groups via `add_param_groups` after the call to amp.initialize is not currently supported but I could add support for it.

Finally, when you say you are "adding new layers" does that mean you are altering the model itself by adding layers to it on the fly, or is the model fully constructed from the beginning, and you are simply allowing more of the model's parameters to be seen/updated by the optimizer (by sending these params to the optimizer via `add_param_group` or by changing their requires_grad attributes to True)?

Tetratrio:
I am adding layers on the fly, but I should be able to construct the entire network if I make some changes in my code. This is the only time I have ever had the need for `add_param_group` and I can work around it anyway.

It is a feature I can live without but if you feel like adding it I would still use it. Maybe there are other types of more dynamically growing networks that need it?

mcarilli:
[...]

PR #310, opened 2019-05-15
Support add_param_group

mcarilli:
Possible fix for #292

### DIFF
--- a/apex/amp/_process_optimizer.py
+++ b/apex/amp/_process_optimizer.py
@@ -106,7 +106,7 @@ def post_backward_with_master_weights(self, scaler):
         if fp16_param.grad is None and fp32_param.grad is not None:
             continue
         elif fp16_param.grad is not None and fp32_param.grad is None:
-            fp32_param.grad = torch.empty_like(fp32_param) 
+            fp32_param.grad = torch.empty_like(fp32_param)
             fp16_grads_needing_unscale.append(fp16_param.grad)
             new_fp32_grads.append(fp32_param.grad)
         elif fp16_param.grad is not None and fp32_param.grad is not None:
@@ -176,7 +176,7 @@ def lazy_init_no_master_weights(self):
                 raise TypeError("Optimizer's parameters must be either "
                                 "torch.cuda.FloatTensor or torch.cuda.HalfTensor. "
                                 "Received {}".format(param.type()))
-    
+
     stash.all_fp16_grad_stash = [None for _ in stash.all_fp16_params]
     stash.all_fp32_grad_stash = [None for _ in stash.all_fp32_params]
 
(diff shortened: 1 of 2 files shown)


vfdev-5:
@mcarilli does this PR alternate internal optimizer's `param_groups` ? In Ignite contrib lr schedulers take directly references to groups instead of optimizer. So, once user `amp.initialize` after defining an lr scheduler it wont work as groups seem to be redefined. Any chance to keep the references to groups the same ?

mcarilli:
Amp patches the [original optimizer.add_param_group](https://pytorch.org/docs/stable/_modules/torch/optim/optimizer.html#Optimizer.add_param_group) to mutate the incoming new group in place (creating FP32 master params from FP16 model params if necessary), then [calls the original optimizer.add_param_group](https://github.com/NVIDIA/apex/blob/master/apex/amp/_process_optimizer.py#L405) to add the new group.

When you call `optimizer.add_param_group(new_group)`, it should mutate the incoming `new_group` dict in place.  In other words if you (or some LR scheduler) stash some external reference to `new_group`, then call `optimizer.add_param_group(new_group)`, the external references should still point to the same dict object that ends up being saved in `optimizer.param_groups` after the call to `add_param_group`.  Some of the params owned within the dict may have been changed (for example, some may have been cast from fp16 to fp32).  But saying e.g. `external_reference['lr'] += 1.0` later in the script should be visible/picked up by the optimizer.  Is this not what you're seeing?

vfdev-5:
@mcarilli thanks for the answer. I see more or less what you mean. But I can not say whether it is related to my problem. More precisely my problem is the following :
```python
import torch
print(torch.__version__)

model = torch.nn.Linear(1, 1).to('cuda')
optim = torch.optim.SGD(model.parameters(), lr=0.1)

param_groups = optim.param_groups
param_group = optim.param_groups[0]
print(id(param_groups), id(param_group), param_group['lr'])

from apex import amp
model, optim = amp.initialize(model, optim, opt_level="O2", num_losses=1)

new_param_groups = optim.param_groups
new_param_group = optim.param_groups[0]
print(id(new_param_groups), id(new_param_group), new_param_group['lr'])

# As ignite's classes work directly on param_groups passed from the previous optimizer
# internally it sets for example lr to another value:
param_group['lr'] = 0.01
# but current optimizer is not updated
print(optim.param_groups[0]['lr'])
```
with the output
```
1.1.0
140687541229960 140687957656704 0.1
Selected optimization level O2:  FP16 training with FP32 batchnorm and FP32 master weights.

Defaults for this optimization level are:
enabled                : True
opt_level              : O2
cast_model_type        : torch.float16
patch_torch_functions  : False
keep_batchnorm_fp32    : True
master_weights         : True
loss_scale             : dynamic
Processing user overrides (additional kwargs that are not None)...
After processing overrides, optimization options are:
enabled                : True
opt_level              : O2
cast_model_type        : torch.float16
patch_torch_functions  : False
keep_batchnorm_fp32    : True
master_weights         : True
loss_scale             : dynamic
140689438827528 140687957665472 0.1
0.1
```

As you can see, once I initialized model and optimizer with amp. Optimizer's param_groups are not the same as the previous ones. Any ideas why it is so and whether we could keep them the same ?

mcarilli:
So, you're worried about the effects of the amp.initialize call, not the effects of calling add_param_group explicitly?   The add_param_group PR should not have altered the behavior of amp.initialize.

vfdev-5:
Yes, seems like that. However, I do not see where in `_process_optimizer` function param_groups are altered. Actually, why apex need to alter `param_groups` instead of maybe replacing "params" ?

PR #322, opened 2019-05-22
New PR for #321

ptrblck:
Last commit wasn't merged last time.


mcarilli:
The diffs here were weird.  They showed that both lines 73 and 81 were being changed by the PR, but line 73 had already been changed in master.  I decided to make the change via a direct push to master instead.

PR #334, opened 2019-05-31
Multi tensor lamb optimizer

thorjohnsen:
Two stage implementation of LAMB optimizer.


mcarilli:
Looks fine but compile time is getting a little high.  Eventually we should reorganize this to be compiled optionally.

francoishernandez:
Hey guys,
Is there some WIP going on to go all the way and add the `fused_lamb_cuda` C++ wrapper and the `apex.optimizers.FusedLamb` python class?
If not, we might have a look, as we're interested in adding this for [OpenNMT-py](https://github.com/OpenNMT/OpenNMT-py).

mcarilli:
@francoishernandez Yes, we definitely plan to expose this via a `apex.optimizers.FusedLamb` Python class.  @sharatht has a local implementation, which we are still testing (not present on Github yet).

francoishernandez:
Great, thanks for your prompt reply ans the good news!

PR #346, opened 2019-06-07
return model only, if no optimizer was specified

### LINKED ISSUE #344: amp.initialize
This code takes advantage of the optional optimizers initializer from #199 

When optimizers are not present, the return type from amp.initialize is a scalar model when enable is True but a tuple when enable is False

```
import torch
import torchvision

from apex import amp

model = torchvision.models.segmentation.deeplabv3_resnet101(pretrained = True)

model = model.cuda().eval()

# Different behavior when enabled vs disabled with no optimizer
# One commented out to only call initialize once - uncomment to repro

# When enabled, returns a single item
#model = amp.initialize(model, enabled=True)

# When disabled, returns a tuple
model, _ = amp.initialize(model, enabled=False)

# Touch-test the model to show the failure when disabled writing to a single item - model(x) fails when model is a tuple
x = torch.zeros([1, 3,480,640]).cuda()
y = model(x)
```

Tested at 1dca16cc9c5 with pytorch 1.0 and torchvision 0.3

ptrblck:
Fixed #344

### DIFF
--- a/apex/amp/frontend.py
+++ b/apex/amp/frontend.py
@@ -308,7 +308,10 @@ def initialize(
     _amp_state.verbosity = verbosity
 
     if not enabled:
-        return models, optimizers
+        if optimizers == None:
+            return models
+        else:
+            return models, optimizers
 
     if not torch.backends.cudnn.enabled:
         raise RuntimeError(


mcarilli:
Heh, I already did the same thing independently.

PR #356, opened 2019-06-13
Add option to turn on/off allreduce in DDP (useful for gradient accum…

thorjohnsen:
Here is my attempt at adding an option to turn on/off allreduce for DDP. By default, it's on, so existing code should still work.

### DIFF
--- a/apex/parallel/distributed.py
+++ b/apex/parallel/distributed.py
@@ -215,6 +215,8 @@ class DistributedDataParallel(Module):
         self.reduction_event = torch.cuda.Event(enable_timing=False, blocking=False) 
         
         self.module = module
+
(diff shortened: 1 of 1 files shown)


apex/parallel/distributed.py
```diff
@@ -249,6 +251,12 @@ def __getstate__(self):
             del attrs['self.reduction_stream']
             del attrs['self.reduction_event']
             return attrs
+
+    def turn_on_allreduce(self):
```
mcarilli: Can we call this `enable_allreduce`?

ngimel:
For the record, here's upstream PR implementing similar functionality in upstream ddp https://github.com/pytorch/pytorch/pull/21736

PR #359, opened 2019-06-13
Separate LDG/STG from compute loop

thorjohnsen:
This implementation of LAMB is a few % faster than the first one.

### DIFF
--- a/csrc/multi_tensor_lamb_stage_1.cu
+++ b/csrc/multi_tensor_lamb_stage_1.cu
@@ -62,19 +62,47 @@ struct LAMBStage1Functor
             i_start < n && i_start < chunk_size;
             i_start += blockDim.x*ILP)
     {
+      GRAD_T r_g[ILP];
+      T r_p[ILP];
+      T r_m[ILP];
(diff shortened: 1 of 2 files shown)


mcarilli:
Good catch.  Sometimes the compiler's not smart enough to reorder all the ILP-ed loads before the ILP-ed stores (in fact if the pointers are not restrict, i think it must conservatively not reorder them).  Without reordering, each STG stalls waiting for the previous LDG, so the unrolled ILP loop confers no benefit in terms of the number of transactions it can put in flight at once.  In my other kernels I try to make sure I [fire off the ILP-ed load loop before the ILP-ed store loop](https://github.com/NVIDIA/apex/blob/master/csrc/multi_tensor_scale_kernel.cu#L49-L74).  I also skimmed the SASS to double check that loads weren't going to stall.  I love this stuff :)

While we're at it:  When you are unpacking pointers from TensorListMetadata and casting them to `T` and `GRAD_T`, will nvcc also let you add `__restrict__` or `const __restrict__` decorations?

PR #360, opened 2019-06-14
update gbn

ekrimer:
* fixed add_relu mem alloc issue
* extended support to bn_group=8
* optimized BN launch times
* switched to a more efficient fence-less group bn communication implementation


apex/contrib/csrc/groupbn/nhwc_batch_norm_kernel.h
```diff
@@ -426,69 +425,52 @@ DEVICE_FUNCTION void parallel_sums_16x2(float *smem, float (&x)[4], int nhw, voi
...
-#endif
+            // now each CTA (on each GPU) reads the data written by CTA 0 of the other GPU
+            volatile float * read_data =
+                &((reinterpret_cast<float*>(params_my_data))[data_offset]);
 
             float other[4];
-                    asm volatile ("ld.global.cv.v4.f32 {%0, %1, %2, %3}, [%4];" 
-        : "=f"(other[0]), "=f"(other[1]), "=f"(other[2]), "=f"(other[3]) : "l"(read_data_));
+            uint32_t other_flag_a, other_flag_b;
+            do {
```
jjsjann123: So we removed the timeout from https://github.com/NVIDIA/apex/pull/360/files#diff-8a2b4400b4fad5e25aede628e56699c4L475
I'm totally OK with it if we think it's not necessary.

ekrimer: that was dead debug code surrounded with `#ifdef BNDEBUG`

jjsjann123: Kinda unrelated question.
Is it necessary to have 1 magic flag per number? Would 1 flag per PTX instruction suffice? In this case it's not saving us anything, but I'm just curious.

ekrimer: Previous version of the kernel used a single flag per 4 data elements and required a system fence to guarantee data arrival prior to flag update. By packing a flag+data into 64bit chunks here I am able to avoid the need for a fence. A different flag to data ratio given the current sizes would not allow a fence-less implementation.

apex/contrib/groupbn/batch_norm.py
```diff
@@ -102,13 +116,29 @@ def __init__(self, num_features, fuse_relu=False, bn_group=1, max_cta_per_sm=2,
...
+        #calculate grid dimentions based on occupancy numbers
+        mp_count = torch.cuda.get_device_properties(None).multi_processor_count
+        self.fwd_grid_dim_x = max(mp_count*self.fwd_occupancy - cta_launch_margin , 1)
+        self.bwd_grid_dim_x = max(mp_count*self.bwd_occupancy - cta_launch_margin , 1)
+        self.addrelu_fwd_grid_dim_x = max(mp_count*self.addrelu_fwd_occupancy - cta_launch_margin , 1)
+        self.addrelu_bwd_grid_dim_x = max(mp_count*self.addrelu_bwd_occupancy - cta_launch_margin , 1)
+        self.grid_dim_y = (num_features + 63) // 64
+
+        # allocate scratch space used by implementation
+        self.ret_cta = torch.cuda.ByteTensor(8192).fill_(0)
```
jjsjann123: As @slayton58 kindly pointed out, we don't need to expose our scratch space.
We can instead do
> `auto retired_ctas = at::zeros({shape}, ...)`
> and let the caching allocator do its job.

apex/contrib/groupbn/batch_norm.py
```diff
@@ -159,21 +196,27 @@ def forward(self, x, z=None):
             return bn_addrelu_NHWC_impl.apply(x, z,
                                   self.weight, self.bias,
                                   self.running_mean, self.running_var,
-                                  self.minibatch_mean, self.minibatch_riv,
+                                  self.minibatch_mean, self.minibatch_riv, self.grid_dim_y, self.ret_cta,
                                   self.momentum,
-                                  self.eps, self.training, self.bn_group, self.my_data, self.pair_data, (self.magic), self.pair_data2,
-                                  self.max_cta_per_sm, self.cta_launch_margin)
+                                  self.eps, self.training, self.bn_group, self.my_data, self.pair_data, (self.magic), self.pair_data2, self.pair_data3,
+                                  self.addrelu_fwd_occupancy, self.addrelu_fwd_grid_dim_x,
```
slayton58: Is there a reason for needed these (occupancy, grid dim) to be exposed & stored in python-land? The implementations look cheap enough to do each time and it'd clean up the interface (along with not needing to expose them through pybind)

jjsjann123:
LGTM. Should be safe to go.
@mcarilli

apex/contrib/groupbn/batch_norm.py
ekrimer: Occupancy and grid_dim_x are determined by querying device properties. Moving these to pytorch allowed reducing to a single device query in construction time instead of on every layer execution. Given it is theoretically possible one will use a single process driving more than one GPU, a simple static variable caching device properties will not be a sufficiently general solution. As such I preferred moving these to pytorch rather than implementing more complex caching scheme.

slayton58: :+1:

apex/contrib/groupbn/batch_norm.py
ekrimer: This would require zeroing out this buffer on every call which will come at cost and will be redundant. Functionally it is sufficient to zero the buffer only before the fist use, but then we have to expose it as done here to hold the pointer for reuse.

jjsjann123: Now I remember that part.
I'm OK with the trade off between perf / cleanliness. What do you think @slayton58 ?

slayton58: I don't think it's friendly, and is exposing complexity to the front-end when I think it could be solved in other ways, but it's your code :)

jjsjann123: I definitely agree that it's not clean to expose it here.
Considering the purpose of this module, I think it's OK to keep it as is for now. Let's leave a comment here just so we have more context for future discussion. @ekrimer

Issue #363, opened 2019-06-17
Prioritize Custom Batch Type "to" method


felix-schneider:
`amp._initialize.applier` checks for a `to()` method of a custom batch type *last*, after checking for Mapping, Iterable, etc. This is a problem for custom batch types that are also iterable, such as those created by torchtext, as it will cause a new object to be created (in the case of torchtext resulting in a crash) rather than the custom `to()` method being called.

Solution: The check for the `to` method should be the first, not the last.

mcarilli:
That's a good point but I don't think the `to` check should be first.  I think the 'to' check should be moved before the iterables, so it doesn't interfere with processing of vanilla Tensors.  For example, we don't want to accidentally convert incoming int tensors to half.

PR #364, opened 2019-06-17
Give custom to method a higher priority

mcarilli:
Fix for https://github.com/NVIDIA/apex/issues/363.

There's a tradeoff here:

This version (giving the `to` method lower priority than Tensors, string_classes, and numpy.ndarrays) will ignore a custom `to` method if the custom batch is an instance of string_classes or numpy.ndarray.  

However, if we give the custom `to` method higher priority, it might be spuriously called if one of the string_classes or numpy.ndarray has a `.to` method already.

A third option is to force user-defined `to` method to have a more unique name, like `to_amp_type`.  But this would require changing the existing code.

@felix-schneider do you see a strong reason to prefer one of these options?

### DIFF
--- a/apex/amp/_initialize.py
+++ b/apex/amp/_initialize.py
@@ -40,12 +40,12 @@ def applier(value, fn):
         return value
     elif isinstance(value, np.ndarray):
         return value
+    elif hasattr(value, "to"): # Allow handling of custom batch classes
+        return fn(value)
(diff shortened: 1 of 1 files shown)


felix-schneider:
I think option 1 (custom `to` is lower than string_classes and ndarrays) is fine. I am not aware of any numpy or string methods with that name, but it's probably best to be on the safe side.

PR #390, opened 2019-07-12
Add missing semicolon.

ezyang:
We plan to remove the implicit semicolon from one of our macros, this will prevent breakage in that case.

Signed-off-by: Edward Z. Yang <ezyang@fb.com>


mcarilli:
Thanks!

PR #405, opened 2019-07-25
Bug fix for non-affine layer-norm + add backward unit test

thorjohnsen:
Some NULL checks were missing for the CPU driver code in the backward call chain. These went undiscovered because the unit test missed a backward test. Added the NULL checks and added backward unit tests.

### DIFF
--- a/tests/L0/run_fused_layer_norm/test_fused_layer_norm.py
+++ b/tests/L0/run_fused_layer_norm/test_fused_layer_norm.py
@@ -4,38 +4,40 @@ import random
 
 import torch
 import apex
+from torch.autograd import Variable
 
         
 class TestFusedLayerNorm(unittest.TestCase):
     def setUp(self):
-        self.module = apex.normalization.FusedLayerNorm(normalized_shape=[32, 64], elementwise_affine=False)
-        self.input_ = torch.randn(16, 32, 64)
+        self.module_cpu_ = apex.normalization.FusedLayerNorm(normalized_shape=[32, 64], elementwise_affine=False).cpu()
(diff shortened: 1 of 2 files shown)


tests/L0/run_fused_layer_norm/test_fused_layer_norm.py
```diff
@@ -4,38 +4,40 @@
...
-        out_ = self.forward_cuda(self.input_)
-        assert out_.is_cuda == True
-        
-    def test_forward_cpu(self):
-        out_ = self.forward_cpu(self.input_)
-        assert out_.is_cuda == False
-        
+        self.input_ = torch.randn(16, 32, 64)
+        self.loss_ = torch.randn(16, 32, 64)
+        self.input_cpu_ = Variable(self.input_.detach().clone(), requires_grad=True)
```
ngimel: `Variable` is an old syntax that's going to be deprecated, can you please replace it with
```
self.input_cpu_ = self.input_.clone().detach().requires_grad_(True)
```
similar for input_cuda_ below.

tests/L0/run_fused_layer_norm/test_fused_layer_norm.py
```diff
@@ -4,38 +4,40 @@
...
+        self.loss_cpu_ = self.loss_.detach().clone()
```
ngimel: I think it's better to create gradOutput directly in the test rather than have it as a class attribute, it would also reduce code duplication between affine and non-affine tests

mcarilli:
Superseded by PR #418

PR #424, opened 2019-08-09
make fused optimizers compatible with AMP

FDecaYed:
Since amp has unified api and deprecated fp16_optimizer, it is confusing for user when fusedadam optimizer still on that api.
In this change:
- optimizers in apex.optimizers can be used with AMP unified API now. when running with AMP, loss scaling is handled by amp and no longer fused.
- parameter flatenning and ops within optimizer are still fused to provide speed up over torch native
- added FusedNovoGrad (described in [https://arxiv.org/abs/1904.03288](url))
- kept FusedSGD unchanged, and still function the old way
- refactored FusedAdam and greatly simplified backend handling
- updated unit tests accordingly


FDecaYed:
@mcarilli 
I have updated the PR and merged in multi_tensor_sgd branch. Now things works like this:
FusedSGD and FusedAdam remains unchanged. Only a message is added to encourage switching.

Added 3 new optimizers in apex.optimizers, SGD, Adam and NovoGrad. These optimizers does not have `Fused` prefix, and are suppose to be used as regular optimizers.
They will work with both O1 and O2, with amp handling loss scaling.

FDecaYed:
After discussion with @mcarilli, here is the final plan:
- FusedSGD now work the same as before
- FusedAdam no longer fuse scaling and casting, but work with O1 now
- Removed special backend handling for FusedAdam
- Print deprecation message for legacy usage of FusedAdam and Fp16_Optimizer
- Added FusedNovoGrad, all 3 fused optimizers can be used alone, or with AMP O1/O2

apex/optimizers/fused_adam.py
```diff
@@ -32,116 +37,75 @@ class FusedAdam(torch.optim.Optimizer):
...
         defaults = dict(lr=lr, bias_correction=bias_correction,
-                        betas=betas, eps=eps, weight_decay=weight_decay,
-                        max_grad_norm=max_grad_norm)
+                        betas=betas, eps=eps, weight_decay=weight_decay)
         super(FusedAdam, self).__init__(params, defaults)
         self.eps_mode = 0 if  eps_inside_sqrt else 1
+        self.dummy_overflow_buf = torch.cuda.IntTensor([0])
 
-    def step(self, closure=None, grads=None, output_params=None, scale=1., grad_norms=None):
+    def step(self, closure=None, grads=None, output_params=None, scale=None, grad_norms=None):
```
mcarilli: We should get rid of this argument list so it doesn't show up in the Sphinx documentation.  I can handle that myself post-merge though.

PR #459, opened 2019-08-30
[novograd] move exp_avg_sq to param device in load_state_dict

FDecaYed:


mcarilli:
Given that the existing behavior of `Optimizer.load_state_dict()` _does_ already cast incoming params to() the device of the preexisting params, i think this is ok.

PR #485, opened 2019-09-10
removing nvtx range used for debugging

jjsjann123:


jjsjann123:
@ptrblck 
Removing nvtx stuff. Not sure if this guy needs to be pushed to some other branch at all :/

Issue #475, opened 2019-09-05
New FusedAdam (since #424 commit 8 Aug 2019) memory Footprint


vince62s:
Team,

This huge PR #424 was not squashed and commits in a long period of time trigger lots of side effects.

As an example, if I use FusedAdam as of this commit 4a8c4ac then the memory footprint in GPUs enables to train with a certain batch size (eg 3072 tokens per batch).

On master, I have to reduce significantly the batch size not to get a CUDA OOM.
(use case is OpenNMT-py)

Not easy to give you numbers, but maybe with some test cases you can better reproduce this issue.

FDecaYed:
In terms of accuracy, updated fusedadam incoporated this upstream fix
https://github.com/pytorch/pytorch/commit/fed5ca192c98619947cca2fb9491ac9624b787d2
Result will be slightly different compare to before, but should be correct and same as upstream adamW now. Do you observe any regression in term of final accuracy? This fix does affect initial steps more.

In terms of memory
updated fused adam no longer fuses grad unscaling and update, thus now fp32 grad is instantiated. This could be the main source of difference if you are comparing against old fusedadam.

There could also be other change affect how memory get reused, but those should not be strict more use. With the dynamic nature of pytorch's allocator, any change on order and life cycle of tensor could cause OOM if you are very close to limit, even though you are not using more memory in total.

vince62s:
I'll go for a full run and check if there is some regression, thanks for the explanation.


> In terms of memory
> updated fused adam no longer fuses grad unscaling and update, thus now fp32 grad is instantiated. This could be the main source of difference if you are comparing against old fusedadam.

What is the rationale for this change ?

FDecaYed:
We want fused optimizers to work with AMP O1, while not doing it by special casing in both optimizer and amp backend.
We expect speed benefit to be marginal(in your case ~1%) and memory benefit also small(most of the case optimizer is not peak of caching allocator). So trading it for expandibility and maintainability is seem a good deal.

@mcarilli speaking of optimizer not being peak. I believe on 4a8c4ac FusedAdam is still handled by FP16_optimizer path. Which means fp32 grad never exist and fp16 grad become None after zero_grad()
In latest master, only fp32 becomes None after zero_grad(), fp16 grad will be zeroed but keep existing through forward/backward(higher potential causing OOM compare to before)
any thought on changing this as well to improve that?
https://github.com/NVIDIA/apex/blob/master/apex/amp/_process_optimizer.py#L366

[...]

PR #517, opened 2019-10-01
move previous fused_adam and fp16_optimizer to contrib

FDecaYed:
add back old fused_adam to contrib, which handle type casting, unscaling and clipping
move fp16_optimizer wrapper with it, which handle flattening, nan ckecking and loss scaling

User should be advised to use AMP with optimizers in apex.optimizers.


setup.py
```diff
@@ -182,6 +175,25 @@ def check_cuda_torch_binary_vs_bare_metal(cuda_dir):
                           extra_compile_args={'cxx': ['-O3'] + version_dependent_macros,
                                               'nvcc':['-O3'] + version_dependent_macros}))
 
+if "--fusedadam" in sys.argv:
```
mcarilli: Can we call this `deprecated_fused_adam` instead?

setup.py
```diff
@@ -182,6 +175,25 @@ def check_cuda_torch_binary_vs_bare_metal(cuda_dir):
...
+    from torch.utils.cpp_extension import CUDAExtension
+    sys.argv.remove("--fusedadam")
```
mcarilli: --deprecated_fused_adam

setup.py
```diff
@@ -182,6 +175,25 @@ def check_cuda_torch_binary_vs_bare_metal(cuda_dir):
...
+
+    from torch.utils.cpp_extension import BuildExtension
+    cmdclass['build_ext'] = BuildExtension
+
+    if torch.utils.cpp_extension.CUDA_HOME is None:
+        raise RuntimeError("--fusedadam was requested, but nvcc was not found.  Are you sure your environment has nvcc available?  If you're installing within a container from https://hub.docker.com/r/pytorch/pytorch, only images whose names contain 'devel' will provide nvcc.")
```
mcarilli: --deprecated_fused_adam

vince62s:
is this somehow a workaround to #475 ?

FDecaYed:
> is this somehow a workaround to #475 ?

We are adding this code path to help investigate the issue and we don't recommend switching back to the deprecated path.

PR #520, opened 2019-10-03
Disable tests for mixed opt_levels, add bitwise accurate test of parameters

ptrblck:
Increased `atol` to `1.5e-4` due to failing `state_dict` comparison for `'O0'` and `'O2'`:
```python
x = torch.tensor(0.8839)
torch.allclose(x.half().float(), x, atol=1e-4)
> False
torch.allclose(x.half().float(), x, atol=1.5e-4)
> True
```


ptrblck:
First comment is deprecated.
The `state_dict` check now compares for bitwise accurate results and only for matching `opt_level` and `restore_opt_level`.

PR #523, opened 2019-10-04
Include loss scaling in README code example

f0k:
The code example in the README, although meant to illustrate checkpointing, is the first `amp` usage example seen by a new user, and it looks so complete that it's tempting to just copy what's written there to adapt one's training script. Unfortunately, it misses the loss scaling, so a new user doing so will find that checkpointing works great, but training doesn't work at all. (Been there, done that.)

I'd propose to just add the two lines needed for loss scaling (this PR), or remove the code example completely so people will be forced to look at the documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,6 +67,9 @@ model, optimizer = amp.initialize(model, optimizer, opt_level=opt_level)
 
 # Train your model
 ...
+with amp.scale_loss(loss, optimizer) as scaled_loss:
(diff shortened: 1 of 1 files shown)


ptrblck:
Thanks @f0k for the suggestion and PR!
I would prefer to add the loss scaling code into the example.

CC @mcarilli what do you think?

mcarilli:
Looks good to me, thanks!

PR #531, opened 2019-10-08
allow for non-distributed envs (Windows)

### LINKED ISSUE #429: AttributeError: module 'torch.distributed' has no attribute 'deprecated'
Hi! 

I get this error on Windows 10 and torch=1.2.0 when just `import apex`

I think my system just not supports this, but it is not good behavior.

BramVanroy:
closes #429

### DIFF
--- a/apex/amp/_initialize.py
+++ b/apex/amp/_initialize.py
@@ -2,6 +2,7 @@ import torch
 from torch._six import string_classes
 import functools
 import numpy as np
+import sys
 import warnings
 from ._amp_state import _amp_state, warn_or_err, container_abcs
 from .handle import disable_casts
@@ -10,8 +11,10 @@ from ._process_optimizer import _process_optimizer
 from apex.fp16_utils import convert_network
 from ..fp16_utils import FP16_Optimizer as FP16_Optimizer_general
 from ..contrib.optimizers import FP16_Optimizer as FP16_Optimizer_for_fused
-from ..parallel import DistributedDataParallel as apex_DDP
-from ..parallel.LARC import LARC
+
+if torch.distributed.is_available():
+    from ..parallel import DistributedDataParallel as apex_DDP
+    from ..parallel.LARC import LARC
 
 
 def to_type(dtype, t):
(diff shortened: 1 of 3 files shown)


mcarilli:
Looks good to me, thanks!  Note that Windows is technically not officially supported...

BramVanroy:
> Looks good to me, thanks! Note that Windows is technically not officially supported...

I am aware. Luckily I was able to build Apex on Windows 10 including all extensions (64bit, CUDA 10.1) and so the only thing stopping me from using it was this small issue. If I recall correctly, DDP will still not work on Windows (?), but at least AMP should work well out-of-the-box.

PS: credit where credit is due: I did nothing more than push the changes that @ptrblck made to the current head. His fork is [here](https://github.com/ptrblck/apex/tree/apex_no_distributed).

apex/amp/_initialize.py
```diff
@@ -139,11 +142,10 @@ def __call__(self, module, state_dict, prefix, local_metadata):
 
 
 def _initialize(models, optimizers, properties, num_losses=1, cast_model_outputs=None):
-    from apex.parallel import DistributedDataParallel as apex_DDP
     from .amp import init as amp_init
 
     optimizers_was_list = False
-    if isinstance(optimizers, torch.optim.Optimizer) or isinstance(optimizers, LARC):
+    if isinstance(optimizers, torch.optim.Optimizer) or ('LARC' in sys.modules and isinstance(optimizers, LARC)):
```
xuzhenqi: This line does not work in linux, because 'LARC' is always not in sys.modules. Please revert this commit. 

```
>>> import sys
>>> import apex
>>> 'LARC' in sys.modules
False
>>> 'apex.parallel.LARC' in sys.modules
True
```
```

BramVanroy: Hm, perhaps @ptrblck can comment on this. I am not sure what that change is supposed to signify. But from the looks of it, you're right that it should be `'apex.parallel.LARC' in sys.modules`.

PR #553, opened 2019-10-19
Make the patched optimizer step function a full method

### LINKED ISSUE #552: 'function' object has no attribute '__self__' in Pytorch 1.3
After upgrading to Pytorch 1.3.0, my training script crashes with `AttributeError`. Relevant lines from the stack:
```
...
  File "/usr/lib/python3.7/site-packages/torch/optim/lr_scheduler.py", line 299, in __init__
    super(CosineAnnealingLR, self).__init__(optimizer, last_epoch)
  File "/usr/lib/python3.7/site-packages/torch/optim/lr_scheduler.py", line 58, in __init__
    self.optimizer.step = with_counter(self.optimizer.step)
  File "/usr/lib/python3.7/site-packages/torch/optim/lr_scheduler.py", line 40, in with_counter
    instance_ref = weakref.ref(method.__self__)
AttributeError: 'function' object has no attribute '__self__'
```

The `step` method of the Amp optimizer is simply patched with a function, but is not made a method again in https://github.com/NVIDIA/apex/blob/088985936518be7e25795a30d8ab33affa9db6ed/apex/amp/_initialize.py#L235-L245, therefore it does not have a `__self__`.

aronhoff:
Rather than simply being a function stored as an instance member.

Fixes #552

### DIFF
--- a/apex/amp/_initialize.py
+++ b/apex/amp/_initialize.py
@@ -3,6 +3,7 @@ from torch._six import string_classes
 import functools
 import numpy as np
 import sys
+from types import MethodType
 import warnings
 from ._amp_state import _amp_state, warn_or_err, container_abcs
 from .handle import disable_casts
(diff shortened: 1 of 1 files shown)


mcarilli:
Good catch, thanks!

PR #594, opened 2019-11-07
Fwd/bwd compatibility around pip 19.3

mcarilli:
Fixes https://github.com/NVIDIA/apex/issues/548#issuecomment-550486441.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -2,7 +2,6 @@ import torch
 from setuptools import setup, find_packages
 import subprocess
 
-from pip._internal import main as pipmain
 import sys
 import warnings
 import os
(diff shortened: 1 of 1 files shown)


adityaiitb:
LGTM.

setup.py
```diff
@@ -32,6 +31,10 @@
 ext_modules = []
 
 if "--pyprof" in sys.argv:
+    try:
+        from pip._internal import main as pipmain
```
ptrblck: `19.3.1` will be able to import it without an `ImportError`, so we need to call `pipmain` and catch the `TypeError` instead.

ptrblck: This should work:
```python
if "--pyprof" in sys.argv:
    try:
        from pip._internal import main as pipmain
        pipmain() # dummy call to raise the TypeError for pip >= 19.
    except TypeError:
        from pip._internal.main import main as pipmain
```

Issue #601, opened 2019-11-12
apex.pyprof.prof throwing exception when handling bool datatype


riskybacon:
While executing:
```
python -m apex.pyprof.prof -w 100 -c kernel,op,sil,tc,flops,bytes,device,stream,block,grid net.dict
```

I get this exception:

```
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/local/lib/python3.6/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/usr/local/lib/python3.6/site-packages/apex-0.1-py3.6-linux-x86_64.egg/apex/pyprof/prof/__main__.py", line 10, in <module>
    main()
  File "/usr/local/lib/python3.6/site-packages/apex-0.1-py3.6-linux-x86_64.egg/apex/pyprof/prof/prof.py", line 224, in main
    bytes = xx.bytes()
  File "/usr/local/lib/python3.6/site-packages/apex-0.1-py3.6-linux-x86_64.egg/apex/pyprof/prof/convert.py", line 61, in bytes
    b = self.elems() * (Utility.typeToBytes(self.stype) + Utility.typeToBytes(self.dtype))
  File "/usr/local/lib/python3.6/site-packages/apex-0.1-py3.6-linux-x86_64.egg/apex/pyprof/prof/utility.py", line 22, in typeToBytes
    assert False
AssertionError
```

While digging into the problem a little bit, I found that `utility.py:Utility.typeToBytes` is being fed a boolean for `t`.

This is the line in `net.dict` that causes the failure:

```
[...]

PR #603, opened 2019-11-13
Add support for bool datatype (#601)

riskybacon:
This seems like a possibly naive fix for #601. This unblocked us when using pyprof, but it's not clear to me if it breaks other use cases.

### DIFF
--- a/apex/pyprof/prof/utility.py
+++ b/apex/pyprof/prof/utility.py
@@ -9,10 +9,10 @@ class Utility(object):
 
 	@staticmethod
 	def typeToBytes(t):
-		if (t in ["uint8", "int8", "byte", "char"]):
+		if (t in ["uint8", "int8", "byte", "char", "bool"]):
 			return 1
 		elif (t in ["float16", "half", "int16", "short"]):
-			return 2 
+			return 2
 		elif (t in ["float32", "float", "int32", "int"]):
 			return 4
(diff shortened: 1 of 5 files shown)


mcarilli:
@mkolod @adityaiitb as pyprof authors, I don't know anything about the internals

apex/pyprof/prof/utility.py
```diff
@@ -1,4 +1,5 @@
 from functools import reduce
+import pdb
```
ptrblck: Is the `pdb` import necessary here or could it be removed?

ptrblck:
@riskybacon Could you add a test case please, so that we can make sure it'll work for this new `dtype`?

apex/pyprof/prof/utility.py
riskybacon: Sorry about that, thanks for the catch, I'll remove it.

riskybacon:
I've added unit tests for both a bool and non-bool test case, but I'm not too happy with how I added the test.

To test the code, I exposed `data` and `prof` from `apex/pyprof/__init__.py` and `apex/pyprof/__init__.py`. This doesn't feel right to me, but it seemed like the only way to expose those functions to a unit test.

Also, the test data inside of `TestPyProfData` seems like a brittle solution.

I'm open to suggestions :)

PR #620, opened 2019-11-22
Don't define distributed variable on module import

### LINKED ISSUE #619: _amp_state determines whether running in distributed at import
It looks like the `_amp_state.py` module determines whether pytorch is running in distributed mode at the import level. The `distributed` only seems to be used in `maybe_print`. See code snippet:

https://github.com/NVIDIA/apex/blob/37cdaf4ad57ab4e7dd9ef13dbed7b29aa939d061/apex/amp/_amp_state.py#L38-L52

This causes a couple issues:

1. It will only support the `env://` initialization of torch distributed
2. It will fail if amp is imported before launching the distributed training

Neither of these is an issue for most, since most people launch via `torch.distributed.launch`. However, it can be an issue if you define your own distributed launch function or use `torch.multiprocessing.spawn`. I can't see a good reason to do it this way anyway, as it appears this variable is only used in the `maybe_print` function. I'll submit a pull request to fix this. Let me know if I'm missing something though.

rmrao:
* Fixes #619 
* Determine whether in distributed mode at function call, not at module import
* Support other initializations of torch.distributed other than `env://`

### DIFF
--- a/apex/amp/_amp_state.py
+++ b/apex/amp/_amp_state.py
@@ -1,5 +1,5 @@
 # This is a "header object" that allows different amp modules to communicate.
-# I'm a C++ guy, not a python guy.  I decided this approach because it seemed most C++-like.  
+# I'm a C++ guy, not a python guy.  I decided this approach because it seemed most C++-like.
 # But apparently it's ok:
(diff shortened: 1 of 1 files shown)


mcarilli:
This is helpful, thanks!

PR #644, opened 2019-12-04
Don't use pip.__internal

trivedigaurav:
Not sure why main was imported this way before

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -2,7 +2,7 @@ import torch
 from setuptools import setup, find_packages
 import subprocess
 
-from pip._internal import main as pipmain
+from pip import main as pipmain
 import sys
 import warnings
 import os


ptrblck:
I tested it on `pip==18.0` and `==19.3.1`, which doesn't seem to work.

Raised error:
```python
    Running command python setup.py egg_info
      File "<string>", line 1, in <module>
      File "/tmp/pip-req-build-34u6umbx/setup.py", line 5, in <module>
        from pip import main as pipmain
    ImportError: cannot import name 'main' from 'pip' (.../python3.7/site-packages/pip/__init__.py)
Cleaning up...
```

@trivedigaurav Which `pip` version were you using?

trivedigaurav:
I believe this maybe be the culprit here: https://github.com/pypa/pip/commit/95bcf8c5f6394298035a7332c441868f3b0169f4

How about the following to be safe?

```python
try
    from pip import main as pipmain
except ImportError:
    from pip._internal import main as pipmain
```

PR #652, opened 2019-12-14
add WAR for pip>=19.3.1

ptrblck:


ptrblck:
This PR removes the usage of `pipmain`, as there are some conflicts in `pip==18.0` vs. `pip==19.3.1`.
Tested with both versions locally.

Instead it uses the `extras_require` option as described [here](https://setuptools.readthedocs.io/en/latest/setuptools.html#declaring-extras-optional-features-with-their-own-dependencies).

PR #661, opened 2019-12-18
Fix typo

bonlime:
There is a little typo in docs for Novograd optimizer. It should be beta1 not beta2. Here is an algorithm from the paper. 
image
Also in C sources it is https://github.com/NVIDIA/apex/blob/4ad9b3bd4fab0c8f0e7811b369f59c234ee72ac7/csrc/multi_tensor_novograd.cu#L156


mcarilli:
Thanks!

Issue #680, opened 2020-01-08
pyprof: pipmain TypeError: 'module' object is not callable


hartb:
Building APEX with `--pyprof` in an environment with current pip may fail with:           

```                                          
Traceback (most recent call last):           
  File "setup.py", line 37, in <module>      
    pipmain(["install"] + required_packages) 
TypeError: 'module' object is not callable   
```                                          

This is because pip moved the `main` function into a new `main` module under `pip._internal` between 19.2.3 and 19.3. It looks like they'll (temporarily) re-instate `pip._internal.main` as a function post-19.3.1, but the existing APEX `setup.py` will fail for `--pyprof` installs with pip 19.3 or 19.3.1.                                                        

See:                                         

- https://github.com/pypa/pip/commit/09fd200c599de4fadf2ff814a1bef855bc6d77e8             
- https://github.com/pypa/pip/issues/7498    
- https://github.com/pypa/pip/pull/7499

hartb:
Will submit a PR

hartb:
Fixed by https://github.com/NVIDIA/apex/pull/652

Thank you! Will close.

PR #681, opened 2020-01-08
tolerate pip 19.3 for --pyprof builds

hartb:
The 'main' function was moved between pip 19.2.3 and 19.3, which
causes '--pyprof' build to fail with (at least) pip 19.3 and 19.3.1:

      File "setup.py", line 37, in <module>
        pipmain(["install"] + required_packages)
    TypeError: 'module' object is not callable

A permanent solution for this might require rewriting the '--pyprof'
install section to use 'python -m pip', but this change just adjusts
the import to tolerate the 19.3 change.

See:

- https://github.com/pypa/pip/commit/09fd200c599de4fadf2ff814a1bef855bc6d77e8
- https://github.com/pypa/pip/issues/7498
- https://github.com/pypa/pip/pull/7499

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -3,6 +3,9 @@ from setuptools import setup, find_packages
 import subprocess
 
 from pip._internal import main as pipmain
+if not callable(pipmain):
+    from pip._internal.main import main as pipmain
+
 import sys
 import warnings
 import os


hartb:
See https://github.com/NVIDIA/apex/issues/680

mcarilli:
Should be fixed now via https://github.com/NVIDIA/apex/pull/652, I thought we merged that already.

Issue #703, opened 2020-02-05
PR #657 breaks contrib FP16_Optimizer


francoishernandez:
Hi there,
It looks like #657 is breaking `apex.contrib.optimizers.FP16_Optimizer`. The `fp32_groups_flat` attribute is removed but is still used in the definition of `state_dict()`:
https://github.com/NVIDIA/apex/blob/2ca894da7be755711cbbdf56c74bb7904bfd8417/apex/contrib/optimizers/fp16_optimizer.py#L199
and `load_state_dict()`:
https://github.com/NVIDIA/apex/blob/2ca894da7be755711cbbdf56c74bb7904bfd8417/apex/contrib/optimizers/fp16_optimizer.py#L241

cc @vince62s @kexinyu @mcarilli

vince62s:
Just changing these fp32_groups_flat to fp32_groups seems to fix this.

kexinyu:
Thanks for bringing this up! Just created a PR for this https://github.com/NVIDIA/apex/pull/704.

PR #704, opened 2020-02-05
Fix attribute name mismatch in state_dict() and load_state_dict()

kexinyu:
This fixes https://github.com/NVIDIA/apex/issues/703.
`fp32_groups_flat` should be changed to `fp32_groups` due to changes from https://github.com/NVIDIA/apex/pull/657.


francoishernandez:
@kexinyu
When there are lists of params in the `state_dict["fp32_groups"] there needs to be an additional loop.
e.g. at the end of `load_state_dict`
``` python
for current, saved in zip(self.fp32_groups, state_dict['fp32_groups']):
    for _current, _saved in zip(current, saved):
        _current.data.copy_(_saved.data)
```

PR #725, opened 2020-02-19
Changes to make xentropysoftmax load/store vectorized when possible:

FDecaYed:
Increase default ILP so that each thread handle 16 Bytes data in one step
Make thread load/store longest vector possible
Support shift for blocks that are not aligned. Greatly improve performance for cases where `classes % 8 > 0`

### DIFF
--- a/apex/contrib/csrc/xentropy/xentropy_kernel.cu
+++ b/apex/contrib/csrc/xentropy/xentropy_kernel.cu
@@ -70,7 +70,6 @@
  * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
  * POSSIBILITY OF SUCH DAMAGE.
  */
-
 #include <ATen/ATen.h>
 #include <ATen/cuda/CUDAContext.h>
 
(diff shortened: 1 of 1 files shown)


apex/contrib/csrc/xentropy/xentropy_kernel.cu
```diff
@@ -123,7 +124,7 @@ const int max_threads = 1024;
 inline dim3 SoftMax_getBlockSize(int ILP, uint64_t dim_size) {
   uint64_t block_size = 1;
   uint64_t max_block_size = std::min(dim_size / ILP, static_cast<uint64_t>(max_threads));
-  while (block_size < max_block_size) block_size *= 2;
+  while (block_size < (max_block_size/2)) block_size *= 2;
```
slayton58, line 127: Why is this change needed? Should it be `max_block_size / ILP`?

FDecaYed: resolved with simon:
`/n` here means `(n-1)/n` portion of the block will be vectorized in case of shifting due to not multiple/aligned. Too high of `n` makes block size too small, so I chose 2 here.

PR #732, opened 2020-02-26
NHWC support for multi tensor apply

mcarilli:
@ngimel @VitalyFedyunin

This is only to unblock tests until the native integrations of [autocasting](https://github.com/pytorch/pytorch/pull/32140) and [gradient scaling](https://github.com/pytorch/pytorch/pull/33366) are merged.

As far as I can tell, grabbing data_ptr()s + numel()s from a list of nhwc-contiguous tensors and using them to launch the multi tensor apply kernel just works, as long as all the tensors that interact across lists are the same format.  For example, if you're adding two lists, this will be fine:
[nhwc0, nhwc1, nchw0, nchw1] +
[nhwc2, nhwc3, nchw2, nchw3] =
[result0, result1, result2, result3]
but this won't:
[nhwc0, nhwc1, nhwc2, nhwc3] +
[nchw0, nchw1, nchw2, nchw3] =
[result0, result1, result2, result3]

### DIFF
--- a/csrc/multi_tensor_apply.cuh
+++ b/csrc/multi_tensor_apply.cuh
@@ -56,7 +56,11 @@ void multi_tensor_apply(
     for(int t = 0; t < tensor_lists[l].size(); t++)
     {
       // TODO:  Print which tensor fails.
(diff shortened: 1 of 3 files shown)


VitalyFedyunin:
Works fine to propagate channels last through the network.

PR #740, opened 2020-03-04
disable ninja for multihead_attn

ptrblck:
Ninja fails to copy submodule files at the moment. Disable it to build fast_multihead_attn


ptrblck:
https://github.com/NVIDIA/apex/pull/746 will include this fix, so closing.

PR #752, opened 2020-03-11
Fix contrib fused_adam to work correctly with multi-GPU

msbaines:
The cuda kernel used by fused-adam was using the default stream
on the default device. The kernel needs use the same device as
the parameter tensor.

Fixed by using context manager to set correct default device. For
the use_mt case, raised an error. Alternatively, the use_mt
case could launch one kernel per cuda device.

Before this change, my model would crash.

The non-contrib version will also need to be fixed.


FDecaYed:
This fix is great. For mt case, user can separate gradient into groups as work around before we have more general fix.

@mcarilli I'm a little worried this might be a wider spread problem, especially as model parallel become more of a thing. For extensions in this repo, none of them check current device against operand tensors after things got into C++. 
How does pytorch native handle this kind of dispatch and is there a standard practice for making extension?

PR #753, opened 2020-03-12
Reversible fused adam with mt support

thorjohnsen:
Add undo feature to fused adam cuda kernel in contrib area.


apex/contrib/csrc/optimizers/fused_adam_cuda_kernel.cu
```diff
@@ -384,3 +384,652 @@ void fused_adam_cuda_mt(
...
+    const int threadIdInBlock = threadIdx.y * blockDim.x + threadIdx.x;
+    const int i = (blockId * threadsPerBlock + threadIdInBlock) * stride;
+    const int totThreads = gridDim.x*gridDim.y*threadsPerBlock*stride;
+
+    if (clear_overflow_first) {
+        if (i == 0) {
+            *noop_gmem = 0;
+        }
+        __syncthreads();
+    }
```
cbcase, line 484: Could this have a race condition, as `__syncthreads()` is only within a block? If blockId zero gets stalled, then some other block will run ahead and start setting `*noop_gmem=1`, but then it will get clobbered back to 0 by the first thread bloock

thorjohnsen: You are right. The reason why that hasn't been a problem is because, in practice, we only launch a single thread block for all the cases we use that kernel for. The code does not guarantee this, though, if there are more shards than there are threads in a single block, multiple blocks will be launched.

thorjohnsen: Number of blocks is now hard-wired at 1 when clear_overflow_flag is set, thus preventing a race condition from ever happening. This has no performance impact for any of the mlperf benchmark configs, since none of them will ever launch more than 1 block.

cbcase:
Putting an actual LGTM here :). As discussed offline, we'll want to converge on a single DistributedAdam version, but we can keep the multiples for further testing

PR #763, opened 2020-03-18
enable wider load/store for multi_tensor_apply kernels

FDecaYed:


apex/contrib/csrc/optimizers/fused_adam_cuda_kernel.cu
```diff
@@ -14,6 +14,17 @@
 #define BLOCK_SIZE 512
 #define ILP 4
 
+template<typename T>
```
mcarilli, line 17: Can these go in multi_tensor_apply.cuh?

FDecaYed: yeah that make sense. Original plan was just modify one of the kernel, so I decided to keep it in the same file just to be more clear what's the value.

FDecaYed: There are other code added separately not in the PR also doing it. I'll create a new PR to clean them all up

PR #791, opened 2020-04-11
reimplement data_prefetecher in a more pyhonic way

gzygzy9211:
```python
for input, target in data_prefetcher(loader):
    # do what you want
```

### DIFF
--- a/examples/imagenet/main_amp.py
+++ b/examples/imagenet/main_amp.py
@@ -261,35 +261,32 @@ def main():
                 'optimizer' : optimizer.state_dict(),
             }, is_best)
 
-class data_prefetcher():
-    def __init__(self, loader):
-        self.loader = iter(loader)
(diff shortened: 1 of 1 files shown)


gzygzy9211:
any news about review?

mcarilli:
The idea of a generator expression is fine (although you misspelled `yield`) and I've implemented prefetchers in other models using generator expressions.

However, the example is not intended to show the "most pythonic" way of prefetching.  It's supposed to demonstrate an idea that's sometimes useful.  Once people have the idea, they can implement it in their own models however they want.

Generator expressions are a cool way to write it, but I think they're less intuitive to read than a class with explicit `next` and `preload` methods.  I appreciate you showing how the generator expression version is done, and I'm glad it works for your code, but I don't think it's appropriate for the example here.

PR #792, opened 2020-04-13
[sync BN]

jjsjann123:
support non-uniform batch size across process group.

TODO: test should be added later as well.


jjsjann123:
Tests added, let's merge this guy! @ptrblck

PR #819, opened 2020-05-01
Use global gradient clipping in FusedLAMB & add option for using NVLAMB

kexinyu:
Changes include:
For `apex.optimizers.FusedLAMB`

- use global norm rather than per group norm in gradient clipping 

- add option for using "NVLAMB" (by setting `use_nvlamb`=True), which applies adaptive learning rate to all parameters (the default is to only apply to those with non-zero weight decay)

### DIFF
--- a/apex/optimizers/fused_lamb.py
+++ b/apex/optimizers/fused_lamb.py
@@ -1,4 +1,5 @@
 import torch
+import math
 from apex.multi_tensor_apply import multi_tensor_applier
 
 class FusedLAMB(torch.optim.Optimizer):
--- a/apex/optimizers/fused_lamb.py
+++ b/apex/optimizers/fused_lamb.py
@@ -51,6 +51,8 @@ class FusedLAMB(torch.optim.Optimizer):
             method is called. (default: True)
         max_grad_norm (float, optional): value used to clip global grad norm
             (default: 1.0)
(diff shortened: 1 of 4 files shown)


apex/optimizers/fused_lamb.py
```diff
@@ -100,6 +105,34 @@ def step(self, closure=None):
...
+            g_norm_32 = multi_tensor_applier(self.multi_tensor_l2norm,
+                                             self._dummy_overflow_buf,
+                                             [g_all_32], False)[0].item()
+        if len(g_all_16) > 0:
+            g_norm_16 = multi_tensor_applier(self.multi_tensor_l2norm,
+                                             self._dummy_overflow_buf,
+                                             [g_all_16], False)[0].item()
+
+        # blend two grad norms to get global grad norm
+        global_grad_norm = math.sqrt(g_norm_32 * g_norm_32 + g_norm_16 * g_norm_16)
```
FDecaYed, line 133: maybe it is better to keep data on device here and save a sync? you can use another multi_tensor_l2norm to do it

kexinyu: Call `multi_tensor_l2norm` on `[g_norm_32, g_norm_16]` to replace that `math.sqrt()` so that we only need one `.item()` at the end?

PR #822, opened 2020-05-04
[Optimizers] apex.optimizers.FusedAdagrad implementation

ajtulloch:
This is an implementation of the standard Adagrad optimizer in Apex, mirroring the [`torch.optim.Adagrad`](https://pytorch.org/docs/stable/optim.html#torch.optim.Adagrad) baseline implementation.

Performance is substantially improved, especially for small models. Here, we compare the effective bandwidth achieved of `optim.step()` on a simple MLP as we vary the hidden dimension `N`

```.py
       nn.Sequential(
            nn.Linear(N, N),
            nn.Linear(N, N),
            nn.Linear(N, N),
            nn.Linear(N, N),
            nn.Linear(N, N),
            nn.Linear(N, N),
            nn.Linear(N, 1),
        )
```

V100 perf

This supports the standard weight decay implementation, as well as an AdamW-style decoupled weight decay (disabled by default).

### DIFF
--- a/apex/optimizers/__init__.py
+++ b/apex/optimizers/__init__.py
@@ -2,3 +2,4 @@ from .fused_sgd import FusedSGD
 from .fused_adam import FusedAdam
 from .fused_novograd import FusedNovoGrad
 from .fused_lamb import FusedLAMB
+from .fused_adagrad import FusedAdagrad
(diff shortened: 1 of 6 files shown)


mcarilli:
Impressive work!  Looks like you're using multi tensor apply correctly, correctly handling None gradients and params with different types, and the integration looks fine.  I'm going to trust you that the kernel math is correct.  @ptrblck is testing the build, if it succeeds in our CI we can merge.

ptrblck:
Thanks for the PR @ajtulloch!
The number look really good. I've imported your branch to a few setups and all pass, so merging it. :)

ajtulloch:
Thanks @mcarilli  and @ptrblck.

PR #851, opened 2020-05-26
make FusedLAMB async

kexinyu:
Making FusedLAMB async: pass global norm by tensor directly into later kernel and get rid of `.item()`


FDecaYed:
LGTM. @ptrblck

PR #854, opened 2020-05-27
Fuses dropout and softmax in backward pass, add bias support to CPP MHA, add additive mask support, separate Q/K/V parameters

seryilmaz:
at::fused_dropout is to generate the same random mask as python version to make testing easier and to keep it consistent (otherwise they generate different random mask).
This PR builds on PRs 852 and 853.


apex/contrib/multihead_attn/self_multihead_attn.py
```diff
@@ -67,10 +88,20 @@ def __init__(self, embed_dim, num_heads, dropout=0., bias=False, include_norm_ad
             else :                   assert False, "Unsupported impl: {} !".format(impl)
 
     def reset_parameters(self):
-        nn.init.xavier_uniform_(self.in_proj_weight)
+        if self.separate_qkv_params:
```
kevinstephano, line 91: Is the separate_qkv_params just for initialization purposes?  I see it concated but never addressed separated, I think.

seryilmaz: Separation is handled by autograd directly.

PR #856, opened 2020-05-27
Fixed a typo in LAMB optimizer with O3 mode

xeonvs:
Simple fix a typo in LAMB optimizer when use O3 optimization level


mcarilli:
Good catch, thanks!

PR #862, opened 2020-05-30
Add Pyprof removal warnings that point to new repo

mcarilli:
Pyprof now has its own repo (https://github.com/NVIDIA/PyProf) and code should be removed from Apex soon.  In the meantime, add warnings that point to new repo.


mcarilli:
elias approved.

PR #885, opened 2020-06-15
2d masking and sparsity

a-maci:
Adding comments for 2d pruning: importance and usage and other minor cleanups.
No functional change.


a-maci:
Filing PR

PR #892, opened 2020-06-23
add unit tests for FusedLAMB optimizer

kexinyu:
Add reference LAMB implementation, which is a PyTorch equivalent to this TensorFlow addon
https://github.com/tensorflow/addons/blob/master/tensorflow_addons/optimizers/lamb.py + global grad norm clipping,
to test `FusedLAMB(use_nvlamb=True)`, i.e. apply adaptive learning rate to all parameters

### DIFF
--- a/tests/L0/run_optimizers/test_lamb.py
+++ b/tests/L0/run_optimizers/test_lamb.py
@@ -1,0 +1,253 @@
+import unittest
+import os
+
+import torch
+from torch.optim import Optimizer
+import apex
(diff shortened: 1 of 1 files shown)


FDecaYed:
@kexinyu could you also add a test for weight_decay != 0 case?

kexinyu:
> @kexinyu could you also add a test for weight_decay != 0 case?

Non-zero weight decay test cases added.

PR #924, opened 2020-07-30
fix additive mask option failure with base MHA in default mode

seryilmaz:
Currently base MHA (no layer norm) fails in default mode as additive mask option broke it. This PR fixes the bug. Fix is verified by running /apex/contrib/test/multihead_attn/test_self_multihead_attn.py before and after the fix.


kevinstephano:
BERT mlperf reference uses additive mask to begin with, it adds -10000 (it also makes softmax output 0, just like -inf does). Also, my theory is -inf input to softmax causes problem when all numbers in a row are -inf, because softmax substracts the max from all elements for numerical robustness before exponential and -inf - -inf is NaN.

Just a comment on why additive masking was used.

Looks good!

PR #926, opened 2020-08-01
Make sgd fused optimizer support setting grads to None

ngimel:
Port of @ajtulloch's diff

### DIFF
--- a/apex/optimizers/fused_sgd.py
+++ b/apex/optimizers/fused_sgd.py
@@ -76,7 +76,8 @@ class FusedSGD(Optimizer):
     def __init__(self, params, lr=required, momentum=0, dampening=0,
                  weight_decay=0, nesterov=False,
                  wd_after_momentum=False,
-                 materialize_master_grads=True):
+                 materialize_master_grads=True,
+                 set_grad_none=False):
         if lr is not required and lr < 0.0:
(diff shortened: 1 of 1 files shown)


mcarilli:
@ajtulloch I don't see any harm in this, but I also don't see much need for it.  Does the [documented way ](https://pytorch.org/docs/stable/autograd.html#default-gradient-layouts) (manually setting grads to None)
```
    for param in model.parameters():
        param.grad = None
```
not work?

ajtulloch:
Yes, it's mostly for consistency with other Apex optimizers - `FusedSGD` is the only optimizer that doesn't support this, c.f.

- https://github.com/NVIDIA/apex/blob/274cc0630286d598e14860cfda91752d0840f712/apex/optimizers/fused_adagrad.py#L59-L63
- https://github.com/NVIDIA/apex/blob/274cc0630286d598e14860cfda91752d0840f712/apex/optimizers/fused_adam.py#L82-L85
- https://github.com/NVIDIA/apex/blob/274cc0630286d598e14860cfda91752d0840f712/apex/optimizers/fused_lamb.py#L89-L92
- https://github.com/NVIDIA/apex/blob/274cc0630286d598e14860cfda91752d0840f712/apex/optimizers/fused_novograd.py#L91-L94

PR #927, opened 2020-08-02
set device guard for multi tensor optimizer implementations

ngimel:
Per title, in pytorch core usually device guard handles situations where device of the tensors is different from the current device. Extensions should do it manually. 
3 optimizers fixed: adam, adagrad and sgd. 
Added tests for this behavior.


csrc/multi_tensor_sgd_kernel.cu
```diff
@@ -160,6 +160,9 @@ void multi_tensor_sgd_cuda(
         TORCH_CHECK(tensor_lists[3][i].scalar_type() == at::ScalarType::Half,
                  "Additional output tensors should always be fp16.");
 
+  noop_flag=noop_flag.to(tensor_lists[0][0].device());
```
mcarilli, line 163: Moving this to multi_tensor_apply.cuh would silently avoid the same crash for all other multi_tensor_apply clients.

However, maybe we want a crash.  Maybe we want to inform calling code noop_flag should be on the same device as the tensors, and give calling code the chance to put it there efficiently/permanently, rather than silently concealing the issue with `.to(tensor_lists[0][0].device());` every time the C++ backend is called.  If so, we should remove the above line and initialize
```python
self._dummy_overflow_buf = torch.cuda.IntTensor([0], device=self.param_groups[0]["params"][0].device)
```
In fact it's crazy ^ is not already present, I think we should add that regardless of C++ diffs.

csrc/multi_tensor_apply.cuh
```diff
@@ -61,7 +63,7 @@ void multi_tensor_apply(
       contiguous_memory = (contiguous_memory || tensor_lists[l][t].is_contiguous(at::MemoryFormat::ChannelsLast));
 #endif
       TORCH_CHECK(contiguous_memory, "A tensor was not contiguous.");
-      TORCH_CHECK(tensor_lists[l][t].is_cuda(), "A tensor was not cuda.");
+      TORCH_CHECK(tensor_lists[l][t].device()==ref_device, "A tensor was not on the same device as the first tensor");
```
mcarilli, line 66: nit:  spaces around ==

csrc/multi_tensor_sgd_kernel.cu
mcarilli: Sorry, forgot link to the Python line.  I think using `TORCH_CHECK(noop_flag.device() == tensor_lists[0][0].device(), "message of your choice")` here, together with changing [this line](https://github.com/NVIDIA/apex/blob/5b53121a2124d70dad0fd4f462d1392d06573ca9/apex/optimizers/fused_sgd.py#L101) in `FusedSGD.__init__` to use initialize it with the right device
```python
self._dummy_overflow_buf = torch.cuda.IntTensor([0], device=self.param_groups[0]["params"][0].device)
```
is the best approach if tests pass.

tests/L0/run_optimizers/test_fused_optimizer.py
```diff
@@ -159,6 +168,89 @@ def test_adam_option(self):
...
+    def __init__(self, *args, **kwargs):
+        super(TestFusedSGD, self).__init__(*args, **kwargs)
+        self.options = {"lr": .25, "momentum": .125}
+        self.ref_optim = torch.optim.SGD
+        self.fused_optim = apex.optimizers.FusedSGD
+
+    def test_float(self):
+        self.gen_single_type_test(param_type=torch.float)
+
+    @unittest.skip("PyTorch optimizer is not numerically correct for fp16")
```
mcarilli, line 241: Are you skipping the fp16 test unconditionally?  that bad huh?

tests/L0/run_optimizers/test_fused_optimizer.py
```diff
@@ -159,6 +168,89 @@ def test_adam_option(self):
...
+    def __init__(self, *args, **kwargs):
+        super(TestFusedAdagrad, self).__init__(*args, **kwargs)
+        self.options = {"lr": 5e-4, "eps": 1e-08, "weight_decay": 1.0e-5}
+        self.ref_optim = torch.optim.Adagrad
+        self.fused_optim = apex.optimizers.FusedAdagrad
+
+    def test_float(self):
+        self.gen_single_type_test(param_type=torch.float)
+
+    @unittest.skip("PyTorch optimizer is not numerically correct for fp16")
```
mcarilli, line 181: Are you skipping the fp16 test unconditionally?

tests/L0/run_optimizers/test_fused_optimizer.py
```diff
@@ -159,6 +168,89 @@ def test_adam_option(self):
...
+
+    @unittest.skipIf(torch.cuda.device_count()<2, "more than 1 GPU required")
+    def test_multi_device(self):
+        devices = ("cuda:0", "cuda:1")
+        for current_dev, tensor_dev in product(devices, devices):
+            with torch.cuda.device(current_dev):
+                self.gen_single_type_test(param_type=torch.float, device=tensor_dev)
+
+
+    def test_multi_params(self):
```
mcarilli, line 193: Afaict none of these tests stress dtype splitting logic in the optimizers...but if direct updates applied to FP16 parameters are "not numerically correct" per above comment, there's no point trying, and it's none of this PR's business anyway.

tests/L0/run_optimizers/test_fused_optimizer.py
```diff
@@ -50,41 +51,54 @@ def get_max_diff(self, ref_param, tst_param):
...
 
     def test_half(self):
         self.gen_single_type_test(param_type=torch.float16)
 
+    @unittest.skipIf(torch.cuda.device_count()<2, "more than 1 GPU required")
+    def test_multi_device(self):
+        devices = ("cuda:0", "cuda:1")
+        for current_dev, tensor_dev in product(devices, devices):
+            with torch.cuda.device(current_dev):
+                self.gen_single_type_test(param_type=torch.float, device=tensor_dev)
```
mcarilli, line 90: I like this test.  Can you add it at the end of [test_lamb.py](https://github.com/NVIDIA/apex/blob/5b53121a2124d70dad0fd4f462d1392d06573ca9/tests/L0/run_optimizers/test_lamb.py#L255) as well?  Don't bother merging test_lamb.py into this file, that would take more effort than it's worth.

tests/L0/run_optimizers/test_fused_optimizer.py
```diff
@@ -159,6 +168,89 @@ def test_adam_option(self):
...
+        )
+
+        for _ in range(self.iters):
+            self.gen_grad(ref_param, tst_param)
+            ref_optim.step()
+            tst_optim.step()
+            max_abs_diff, max_rel_diff = self.get_max_diff(ref_param, tst_param)
+            self.assertLessEqual(max_abs_diff, self.max_abs_diff)
+            self.assertLessEqual(max_rel_diff, self.max_rel_diff)
+
```
mcarilli, line 211: for extra credit can you include
```
def test_multi_params_different_devices_throws(self):
        sizes = [[4096, 1024], [4096], [4096, 2048], [32320, 1024], [1]]
        adagrad_option = {"lr": 5e-4, "eps": 1e-08, "weight_decay": 0}

        tensors = []
        for i, size in enumerate(sizes):
            tensors.append(torch.rand(size, dtype=torch.float, device="cuda:"+str(i % 2)))
        ref_param, tst_param, ref_optim, tst_optim = self.gen_param_optim(
            tensors, adagrad_option
        )
        self.gen_grad(ref_param, tst_param)
        with self.assertRaises(i forget what error type torch_check throws):
            tst_optim.step()
```

csrc/multi_tensor_l2norm_kernel.cu
```diff
@@ -369,7 +370,7 @@ void multi_tensor_norm_out_cuda(
   const int norm_type)
 {
   auto float_options = tensor_lists[0][0].options().dtype(at::kFloat);
-
+  TORCH_CHECK(tensor_lists[0][0].device() == noop_flag.device(), "noop flag should be on the smae device as tensors");
```
mcarilli, line 373: same

mcarilli:
alright LGTM, thanks!

Issue #931, opened 2020-08-04
Merged PR #925 breaks CPU-Only installs


aatifjiwani:
While running a Github CD job, it attempted to install Apex, but failed with the following traceback: 

```
Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/tmp/pip-req-build-f1vgc77y/setup.py", line 35, in <module>
        _, bare_metal_major, _ = get_cuda_bare_metal_version(cpp_extension.CUDA_HOME)
      File "/tmp/pip-req-build-f1vgc77y/setup.py", line 14, in get_cuda_bare_metal_version
        raw_output = subprocess.check_output([cuda_dir + "/bin/nvcc", "-V"], universal_newlines=True)
    TypeError: unsupported operand type(s) for +: 'NoneType' and 'str'
```

It seems that cuda_dir is None, which is intuitively expected when installing Apex on a CPU only machine. But even when installing on a CPU only machine, the setup is still looking for the Nvidia Cross-Compilers. 

The most recent commit as of 8/4/2020 (hash: 5b53121a2124d...) adds this line into setup, which breaks the install. 

The corresponding line within the current release of Apex:
https://github.com/NVIDIA/apex/blob/5b53121a2124d70dad0fd4f462d1392d06573ca9/setup.py#L14

matthaeusheer:
I can confirm this issue.

Raghava14:
I am facing the same issue as well but in my case I am facing the issue when trying to install apex in a GPU machine.

wangzifu1994:
same here

aatifjiwani:
> I am facing the same issue as well but in my case I am facing the issue when trying to install apex in a GPU machine. Was anyone able to fix it ?

@Raghava14 The fix that we were able to make was install Apex using the commit before the one that broke the installation. Obviously this is not good in the long run but its a fine temporary fix

ptrblck:
Thanks for reporting this issue. The `get_cuda_bare_metal_version` has to be moved inside the MHA code. 
I'll provide a fix tomorrow for it.

qhaas:
Seeing this with CentOS 8.2 with default python3(.6).

Would be helpful to have tags with snapshots that were known to be relatively stable.  Always get nervous when cloning master.

mcarilli:
[...]

Issue #818, opened 2020-04-30
torch.cuda.amp > apex amp


mcarilli:
For a while now my main focus has been moving mixed precision functionality into Pytorch core.  It was merged about a month ago:
https://pytorch.org/docs/master/amp.html
https://pytorch.org/docs/master/notes/amp_examples.html
and is now usable via master or nightly pip/conda packages.  (Full features did not make the 1.5 release, unfortunately.)

`torch.cuda.amp` is more flexible and intuitive, and the native integration brings more future optimizations into scope.  Also, `torch.cuda.amp` fixes many of `apex.amp`'s known pain points.  Some things native amp can handle that apex amp can't:
- Guaranteed Pytorch version compatibility, because it's part of Pytorch
- No need to build extensions
- Windows support
- Bitwise accurate [saving](https://pytorch.org/docs/master/amp.html#torch.cuda.amp.GradScaler.state_dict)/[restoring](https://pytorch.org/docs/master/amp.html#torch.cuda.amp.GradScaler.load_state_dict)
- [DataParallel](https://pytorch.org/docs/master/notes/amp_examples.html#dataparallel-in-a-single-process) and intra-process model parallelism (although we still recommend [torch.nn.DistributedDataParallel](https://pytorch.org/docs/master/notes/amp_examples.html#distributeddataparallel-one-gpu-per-process) with one GPU per process as the most performant approach)
- [Gradient penalty](https://pytorch.org/docs/master/notes/amp_examples.html#gradient-penalty) (double backward)
- `torch.cuda.amp.autocast()` has no effect outside regions where it's enabled, so it should serve [cases that formerly struggled with multiple calls to `apex.amp.initialize()`](https://github.com/NVIDIA/apex/issues/439) (including [cross-validation](https://github.com/NVIDIA/apex/issues/392#issuecomment-610038073)) without difficulty.
- Sparse gradient support in flight (https://github.com/pytorch/pytorch/pull/36786)

Damiox:
Can torch.cuda.amp be used only for inferences on a FP32 model? See https://github.com/NVIDIA/apex/issues/750 and https://github.com/NVIDIA/apex/issues/809
I couldn't find an example in https://pytorch.org/docs/master/notes/amp_examples.html

mcarilli:
[...]

PR #937, opened 2020-08-10
move sm80 code inside MHA

ptrblck:
Should fix https://github.com/NVIDIA/apex/issues/931

Thanks @mcarilli for the review.


mcarilli:
Before this diff, a CPU-only install (`python3 setup.py install` or `python3 setup.py install --user`) fails with the error reported in https://github.com/NVIDIA/apex/issues/931 when the Cuda toolkit is not available/visible.

After this diff, CPU-only install succeeds even if the Cuda toolkit is not available/visible.

Thanks Piotr!

To anyone reading this, please do not consider it an endorsement of Apex as a future-proof source of mixed precision.  [torch.cuda.amp](https://github.com/NVIDIA/apex/issues/818) is the truth.

PR #967, opened 2020-09-25
allow cuda minor version diff

gregjhogan:
remove the requirement to modify code when a cuda minor version difference is acceptable
```sh
pip install -v --no-cache-dir --global-option="--cpp_ext" --global-option="--cuda_ext" --global-option="--cuda_ignore_minor_version" git+https://github.com/NVIDIA/apex.git
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -90,7 +90,7 @@ def get_cuda_bare_metal_version(cuda_dir):
 
     return raw_output, bare_metal_major, bare_metal_minor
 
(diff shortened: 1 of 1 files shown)


stas00:
Unknowingly I sent a similar solution here https://github.com/NVIDIA/apex/pull/997

ptrblck:
Thanks for the PR!

We should not encourage disabling the valid check with an install argument, as these version mismatches are not tested internally. 

Closing this PR.

PR #972, opened 2020-09-30
Update C++ extensions installation instructions

helboukkouri:
Installing apex with the C++ extensions using:
```
pip install -v --no-cache-dir --global-option="--cpp_ext" --global-option="--cuda_ext" ./
```
seems to only install apex_C and amp_C. For me, installing using:
```
python setup.py install --cuda_ext --cpp_ext
```
did the trick 🙂.
Maybe this should be the default instructions in the README ?

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,7 +124,7 @@ CUDA and C++ extensions via
 ```
 $ git clone https://github.com/NVIDIA/apex
 $ cd apex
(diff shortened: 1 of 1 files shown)


ptrblck:
I'm unsure, what the issue using the first command was and how it was solved by the second one.
I would assume both should install apex with the same options.
Could you explain the issue you were seeing?

PR #980, opened 2020-10-17
Optimize the sync batchnorm by batching the communication

lly-zero-one:
In this PR, we mainly tried to optimize the performance of Syncatchnorm and also fixed one potential issue in the welford_parallel kernel implementation.
1) For performance improvement, we batched the mean/var/count all_gather communication together and sent it once in the forward path
2) We also batch the all_reduce in backward path
3) We add the contiguous call on the input of welford_parallel kernel.

If there is any standard perf benchmark, I would be happy to run it.

### DIFF
--- a/apex/parallel/optimized_sync_batchnorm_kernel.py
+++ b/apex/parallel/optimized_sync_batchnorm_kernel.py
@@ -21,33 +21,31 @@ class SyncBatchnormFunction(Function):
             if channel_last:
                 count = int(input.numel()/input.size(-1))
                 mean, var_biased = syncbn.welford_mean_var_c_last(input)
+                num_channels = input.size(-1)
             else:
                 count = int(input.numel()/input.size(1))
                 mean, var_biased = syncbn.welford_mean_var(input)
+                num_channels = input.size(1)
 
             if torch.distributed.is_initialized():
                 if not process_group:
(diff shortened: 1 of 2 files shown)


lly-zero-one:
cc @ngimel

jjsjann123 (approved):
Don't see any major issues and it's passing local test. I'm merging this.

Timen:
The changes introduced in this PR actually reduced my training speed by about 30%, I'm using a 4 GPU setup and O1 fp16 optimizations. Not sure why but I'm reverting to the previous commit.

lly-zero-one:
> The changes introduced in this PR actually reduced my training speed by about 30%, I'm using a 4 GPU setup and O1 fp16 optimizations. Not sure why but I'm reverting to the previous commit.

Can you share the script for reproducing the regression?

Timen:
I'm unfortunately not at liberty to share my training script. I will say however it is a quite complex model setup training on videos and using [NVIDIA/DALI](https://github.com/NVIDIA/DALI) for data loading. I'm wondering what you used to benchmark your changes and where you saw an improvement in training speed?

lly-zero-one:
What if you check the trace and see which part of SyncBatchNorm got slow down? Let me paste the EfficientNet result here.

PR #981, opened 2020-10-20
DistributedFusedAdam Model Parallelism Support (Megatron)

kexinyu:
New features:
- Multiple parameter groups support (e.g. BERT)
- Model parallelism support (e.g. Megatron)
- New distopt adam kernel in `apex/contrib/csrc/optimizers/multi_tensor_distopt_adam_kernel.cu` with exactly the same update math as `apex/csrc/multi_tensor_adam.cu`
- Checkpoint load/save (including step number, master params, first and second moments)
- Add a unit test in `tests/L0/run_optimizers/test_dist_adam.py`

No longer supported:
- `_pipeline_block_step()` (Call step kernel once per block)
- `revert_step()` (Revert effect of previously calling partial_step) `step()` should only be called when there's no overflow happening.


thorjohnsen:
Hi Kexin, my initial review revealed nothing problematic, but since you are changing the distributed optimizer that we used for mlperf 0.7 I have to run the affected benchmarks to verify that they still converge. Single node transformer looked good, once Selene opens back up, I will queue up the multi-node runs.

slayton58 (approved):
@thorjohnsen was reviewing this PR & convergence, and it looks to be good (he's out this week so I'm registering the approval instead)

slayton58:
@ptrblck This has been reviewed from our side, any issues with merging?

tests/L0/run_optimizers/test_dist_adam.py
```diff
@@ -0,0 +1,182 @@
+import argparse
+import random
+import sys
+
+import torch
+from torch.nn.parallel import DistributedDataParallel as DDP
+
+from apex import amp
+from apex.optimizers import FusedAdam
+from apex.contrib.optimizers.distributed_fused_adam_v4 import DistributedFusedAdam
```
ptrblck, line 10: Could you change this to `distributed_fused_adam` and remove the `_v*` files or are they needed?

kexinyu: Done, good catch!

PR #985, opened 2020-11-04
Seryilmaz/fused dropout softmax

seryilmaz:
Fuses dropout and softmax both in fprop and bprop. This is only done in additive masked case for now, used by BERT. Following changes are made:
1) To avoid writing softmax outputs in fprop, recompute softmax results in dgrad kernel
2) Use pytorch philox random number generator for dropout instead of cuRAND since it is faster
3) Kernels are vectorized for large enough sequence length and when sequences are multiples of 4.


kevinstephano (approved):
This looks okay to me.

Issue #988, opened 2020-11-07
apex not supporting CUDA 11.0? [Help me]


MuruganR96:
My nvcc version is cuda 11.0, but I found the pytorch latest version from this website is 10.2
As a result I can't properly install apex.

# ImportError: cannot import name 'amp'

Software Versions pre-installed:
```
Nvidia Driver: 450.51v
CUDA: 11v
cuDNN: 8.0v
Python: 3.8
Docker: 19.03.12v
Nvidia-docker: 2.0v
NGC(Nvidia GPU Cloud) CLI: 1.15.0v
```
```
Traceback (most recent call last):
  File "train.py", line 188, in <module>
    train(num_gpus, args.rank, args.group_name, **train_config)
  File "train.py", line 83, in train
    from apex import amp
ImportError: cannot import name 'amp'
```
not importing apex module. 

Please help me to solve this issue @ptrblck sir. :)

ptrblck:
> but I found the pytorch latest version from this website is 10.2

The latest PyTorch binaries can be installed with CUDA11.0 as shown in the [install instructions](https://pytorch.org/get-started/locally/).

Note that mixed-precision training is available in PyTorch directly via `torch.cuda.amp` as explained [here](https://github.com/NVIDIA/apex/issues/818) and we recommend to use the native implementation. 

In case you have trouble building apex, you could use a PyTorch NGC container with CUDA11.1, where PyTorch and apex are installed.

sakaia:
@ptrblck CUDA 11.0 supports MIG. Is this feature available on PyTorch?  or any tips?

I met following error

```
[ASP][Warning] torchvision cannot be imported.
/home/zdlsadmin/.local/lib/python3.6/site-packages/torch/cuda/__init__.py:104: UserWarning:
A100-PCIE-40GB MIG 3g.20gb with CUDA capability sm_80 is not compatible with the current PyTorch installation.
The current PyTorch install supports CUDA capabilities sm_37 sm_50 sm_60 sm_70 sm_75.
If you want to use the A100-PCIE-40GB MIG 3g.20gb GPU with PyTorch, please check the instructions at https://pytorch.org/get-started/locally/

  warnings.warn(incompatible_device_warn.format(device_name, capability, " ".join(arch_list), device_name))
Traceback (most recent call last):
  File "toy_problem.py", line 87, in <module>
    main(args)
  File "toy_problem.py", line 41, in main
    optimizer = FusedAdam(model.parameters())
[...]

PR #997, opened 2020-11-15
add an option to skip minor ver check

stas00:
As discussed at https://github.com/NVIDIA/apex/issues/988#issuecomment-726343453 we can build apex with cuda-11.1 while having pytorch built with cuda-11.0, so this PR adds an option to skip the minor version check. I understand the same happens with cuda-10. Docs updated.
`setup.py` will also print a note:
```
--skip-minor-ver-check activated: Attempting to use cuda-11.1, despite pytorch using cuda-11.0 - this may or may not work
```
so that it's in the logs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,6 +127,9 @@ $ cd apex
 $ pip install -v --no-cache-dir --global-option="--cpp_ext" --global-option="--cuda_ext" ./
 ```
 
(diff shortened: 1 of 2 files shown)


stas00:
Oh, found someone else sent a PR to solve this problem a while ago https://github.com/NVIDIA/apex/pull/967 
Either works, but if that other is merged instead, please add this new option to `README.md`

ptrblck:
Thanks for the PR!

> As discussed at #988 (comment) we can build apex with cuda-11.1 while having pytorch built with cuda-11.0

A successful build doesn't indicate no issues while executing the compiled kernels.
We should not encourage disabling the valid check with an install argument, as these version mismatches are not tested internally.

Closing this PR.

PR #1000, opened 2020-11-16
Fix lack of proper loading of best_prec1 from the checkpoint

JanuszL:
- resume() is a nested function and when it loads best_prec1 it creates
  a local variable that hides the one from the parent function. This
  PR adds `nonlocal` to modify the parent function variable as intended

Signed-off-by: Janusz Lisiecki <jlisiecki@nvidia.com>


ptrblck:
Thanks for the PR, Janusz! :)

PR #1007, opened 2020-12-01
Distributed LAMB fixes

kexinyu:
1. Fix `beta2`, `beta3` values in DistributedLAMB kernel
2. Gradients unscaling:  load fp16 gradient to register, cast to fp32, unscale, use unscaled fp32 thereafter inside DistributedLAMB
3. dtypes all consistent with the FusedLAMB kernel


slayton58 (approved):
Looks good to me, tested on e2e model.

PR #1025, opened 2021-01-06
cuda rng changes for graph capture with apex MHA

seryilmaz:
Apex MHA dropout implementation keeps RNG states in GPU for graph capture.


seryilmaz:
@mcarilli  please review.

Issue #101, opened 2018-12-15
AMP supported hardware


amoussawi:
Hello!

It's not clear what hardware apex benefits from the most. I'm using AMP on a Tesla K80 to train my model and actually the training slowed down by ~ 1.3x. In the mixed precision [user guide](https://docs.nvidia.com/deeplearning/sdk/mixed-precision-training/index.html), it says that the framework should support Volta Tensor Core math and it only mentions Tesla V100, so is Tesla V100 the best hardware for mixed precision training?

amoussawi:
Closing after i found this. sorry. https://github.com/NVIDIA/apex/issues/14

PR #1028, opened 2021-01-11
Fix example dockerfile

liamwazherealso:
The example docker image does not build per my comments on this [issue](https://github.com/NVIDIA/apex/issues/101).

The suggested image also doesn't work unless an update/install of git is added. I changed the docker file to start with the `pytorch/pytorch:1.6.0-cuda10.1-cudnn7-devel` image and added the two needed commands. 

Output of successful build can be seen in this [gist](https://gist.github.com/liamwazherealso/ef1ff40c507c8f69baf15d879b0f1ec2).


liamwazherealso:
Hi @mcarilli,

I saw you have worked on this file before. Didn't know if you could take a look, thanks.

PR #1032, opened 2021-01-13
[WIP] make DistributedFusedLAMB sync-free

kexinyu:
Changes include:

1. Remove sync points (e.g. D2H memcpy caused by dtype mismatch between kernel arguments and kernel parameters, `.item()` from overflow check, `masked_select()` and `masked_scatter()` from update norm calculation).
2. Enable `step_supports_amp_scaling` to that optimizer handles gradient un-scaling.
2. Add `state_dict()` and `load_state_dict()` so that params/grads/moments buffers & step number can be properly saved and restored from checkpoint.


thorjohnsen:
Looks good to me. What kind of testing has been done? Is this the final version before the merge?

slayton58:
@kexinyu there's a _lot_ of commits here, is a rebase needed?

kexinyu:
use another one

PR #1037, opened 2021-01-27
fast layer norm

yjk21:
Provides faster fwd and bwd kernels for LayerNorm, intended for BERT training.
It should serve as dropin replacement (see `apex/contrib/test/layer_norm/test_fast_layer_norm.py` for details):
```
from apex.contrib.layer_norm import FastLayerNorm

fln = FastLayerNorm(hidden_size).cuda()
y = fln(x)
```
Performance for B=32, S=512 on GA100 - current APEX affine LN kernels vs NEW:
[APEX FWD] Time: 0.0751ms Throughput: 895.7822 GB/sec
[APEX BWD] Time: 0.1762ms Throughput: 572.1678 GB/sec

[ NEW FWD] Time: 0.0535ms Throughput: 1258.0318 GB/sec
[ NEW BWD] Time: 0.0866ms Throughput: 1163.4357 GB/sec

### DIFF
--- a/apex/contrib/csrc/layer_norm/ln_fwd_cuda_kernel.cu
+++ b/apex/contrib/csrc/layer_norm/ln_fwd_cuda_kernel.cu
@@ -1,0 +1,186 @@
+#include "utils.cuh"
+#include "ln_kernel_traits.h"
+#include "ATen/cuda/CUDAContext.h"
+
+template <typename Ktraits>
+__global__ __launch_bounds__(Ktraits::THREADS_PER_CTA) void ln_fwd_kernel(
+    void *__restrict__ y_, void *__restrict__ mu_, void *__restrict__ rsigma_,
+    const void *__restrict__ x_, const void *__restrict__ gamma_,
+    const void *__restrict__ beta_, const float epsilon, int rows) {
+
+  using Vec = typename Ktraits::Vec;
+
+  using base_t = typename Ktraits::base_t;
+  using compute_t = typename Ktraits::compute_t;
+  enum { NUM_ELTS = Vec::NUM_ELTS };
+  enum { WARPS_N = Ktraits::WARPS_N };
+  enum { WARPS_M = Ktraits::WARPS_M };
+  enum { ROWS_PER_CTA = Ktraits::ROWS_PER_CTA };
+
+  enum { THREADS_PER_ROW = Ktraits::THREADS_PER_ROW };
+  enum { BYTES_PER_LDG = Ktraits::BYTES_PER_LDG };
+  static_assert(BYTES_PER_LDG == 16, "");
+
+  enum { BYTES_PER_ROW = Ktraits::BYTES_PER_ROW };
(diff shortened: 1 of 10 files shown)


yjk21:
FYI @seryilmaz

yjk21:
Hi @ptrblck could you pls take a look at this?

kevinstephano:
@yjk21 What improvements did you add to Layer Norm?  It looks like the backward pass reduces the kernels down to 2 but I wasn't so sure what was improved in the forward pass beyond perhaps more vectorization?

yjk21:
Hi @kevinstephano. Thanks for taking a look.
Yes, the main backward kernel crucially reads the large input tensors `dout` and `x` only once, which should constitute the main part of the speedup. It produces `dgrad` and partially `dgamma` and `dbeta`, so still needs a second launch to finalize `wgrad`.
For `fprop`, yes, loads/stores are vectorized to 16B, and the input is kept live in registers. It achieves an ~1.4x speedup on GA100. 
I think the old kernel might be reading the input again before normalizing.

apex/contrib/csrc/layer_norm/ln_fwd_cuda_kernel.cu
```diff
@@ -0,0 +1,186 @@
...
+    const at::Tensor & gamma,
+    const at::Tensor & beta,
+    const float epsilon,
+    const int rows,
+    const int cols,
+    const int max_gridx,
+    cudaStream_t stream
+){
+
+  if (cols == 1024) {
```
tbmihailov, line 142: Is there any particular reason to support only tensors with hidden_size 1024?

yjk21: As it's written, this particular kernel supports only dimensions which are multiples of 512B. So BERT base _should_ also work.

tbmihailov: It seems to me that it only works with dimension of 1024. I did not look deeper than this file and I do not know how kernels are implemented. I tested with different sizes and it raises an exception for any dimension different than 1024:
`
from apex.contrib.layer_norm import FastLayerNorm
dim1=2048
dim2=1024. # 512, 2048, 4096 fail
x = torch.rand(dim1, dim2, dtype=torch.float32).cuda()
fln = FastLayerNorm(dim2).cuda()
y = fln(x)
`

yjk21: Oh yes, I meant that the kernel template can be instantiated with other sizes under the constraint mentioned above.

ngoyal2707: @yjk21 any suggestions on kernel traits for other dimensions (like 2048, 4096, etc)? I tried same as `1024`, but got no perf improvements on `A100`.  I do get better perf with `1024` on A100 as yours

PR #1055, opened 2021-02-16
sync-free Distributed LAMB + parameter reordering

kexinyu:
Changes include:

- Remove sync points (e.g. D2H memcpy caused by dtype mismatch between kernel arguments and kernel parameters, .item() from overflow check, masked_select() and masked_scatter() from update norm calculation).

- Enable step_supports_amp_scaling to that optimizer handles gradient un-scaling.

- Add state_dict() and load_state_dict() so that params/grads/moments buffers & step number can be properly saved and restored from checkpoint.
- Parameter reordering to allow more overlap between backward pass and gradient reduction.


seryilmaz:
@slayton58  did you take a look at this?

PR #1065, opened 2021-03-15
Add unit tests for Fused NovoGrad

sudhakarsingh27:
1. Added unit tests for fused-novograd
2. Fixed issue that noop flag tensor was not on the same device as other tensors.
3. Fixed issue that even with all tensors on the same device, cuda stream is on a different device than tensors.

### DIFF
--- a/tests/L0/run_optimizers/test_fused_novograd.py
+++ b/tests/L0/run_optimizers/test_fused_novograd.py
@@ -1,0 +1,168 @@
+import torch
+from torch.optim import Optimizer
+import math
+import apex
+import unittest
+
+from .test_fused_optimizer import TestFusedOptimizer
+from itertools import product
+
+class Novograd(Optimizer):
+    """
+    Implements Novograd algorithm.
+
+    Args:
+        params (iterable): iterable of parameters to optimize or dicts defining
+            parameter groups
+        lr (float, optional): learning rate (default: 1e-3)
+        betas (Tuple[float, float], optional): coefficients used for computing
+            running averages of gradient and its square (default: (0.95, 0))
+        eps (float, optional): term added to the denominator to improve
+            numerical stability (default: 1e-8)
+        weight_decay (float, optional): weight decay (L2 penalty) (default: 0)
+        grad_averaging: gradient averaging
+        amsgrad (boolean, optional): whether to use the AMSGrad variant of this
+            algorithm from the paper `On the Convergence of Adam and Beyond`_
+            (default: False)
+    """
+
+    def __init__(self, params, lr=1e-3, betas=(0.95, 0), eps=1e-8,
+                 weight_decay=0, grad_averaging=False, amsgrad=False):
+        if not 0.0 <= lr:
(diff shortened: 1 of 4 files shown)


slayton58:
\cc @ptrblck

tests/L0/run_optimizers/test_fused_novograd.py
```diff
@@ -0,0 +1,168 @@
+import torch
+from torch.optim import Optimizer
+import math
+import apex
+import unittest
+
+from .test_fused_optimizer import TestFusedOptimizer
```
ptrblck, line 7: This import will fail with:
```python
ImportError: attempted relative import with no known parent package
```
Could you remove the dot at the beginning?

tests/L0/run_optimizers/test_fused_novograd.py
```diff
@@ -0,0 +1,168 @@
...
+            tensors, self.options, self.tst_options
+        )
+
+        for _ in range(self.iters):
+            self.gen_grad(ref_param, tst_param)
+            ref_optim.step()
+            tst_optim.step()
+            max_abs_diff, max_rel_diff = self.get_max_diff(ref_param, tst_param)
+            self.assertLessEqual(max_abs_diff, self.max_abs_diff)
+            self.assertLessEqual(max_rel_diff, self.max_rel_diff)
```
ptrblck, line 168: The unittest execution seems to be missing via:
```python
if __name__ == '__main__':
    unittest.main()
```
Currently no tests will be run py `python test_fused_novograd.py -v`.

tests/L0/run_optimizers/test_fused_novograd.py
```diff
@@ -0,0 +1,168 @@
...
+
+    def test_half(self):
+        self.gen_single_type_test(param_type=torch.float16)
+
+    @unittest.skipIf(torch.cuda.device_count()<2, "more than 1 GPU required")
+    def test_multi_device(self):
+        devices = ("cuda:1", "cuda:0")
+        for current_dev, tensor_dev in product(devices, devices):
+            with torch.cuda.device(current_dev):
+                print("Some info from the test", current_dev, tensor_dev)
```
ptrblck, line 147: Could you remove this `print` statement or is this needed for the test?

ptrblck (changes requested):
Thanks for the PR! Some changes would be needed before we can merge it. :)

tests/L0/run_optimizers/test_fused_novograd.py
sudhakarsingh27: Done.

tests/L0/run_optimizers/test_fused_novograd.py
sudhakarsingh27: Right, added it now.

tests/L0/run_optimizers/test_fused_novograd.py
sudhakarsingh27: I've removed it.

PR #1074, opened 2021-04-13
fmhalib

yjk21:
Optimized mha kernels for BERT

### DIFF
--- a/apex/contrib/csrc/fmha/fmha_api.cpp
+++ b/apex/contrib/csrc/fmha/fmha_api.cpp
@@ -1,0 +1,305 @@
+/******************************************************************************
+ * Copyright (c) 2011-2021, NVIDIA CORPORATION.  All rights reserved.
+ * 
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *     * Redistributions of source code must retain the above copyright
+ *       notice, this list of conditions and the following disclaimer.
+ *     * Redistributions in binary form must reproduce the above copyright
+ *       notice, this list of conditions and the following disclaimer in the
+ *       documentation and/or other materials provided with the distribution.
+ *     * Neither the name of the NVIDIA CORPORATION nor the
+ *       names of its contributors may be used to endorse or promote products
+ *       derived from this software without specific prior written permission.
+ * 
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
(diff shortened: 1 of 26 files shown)


yjk21:
FYI @slayton58 @seryilmaz

slayton58:
\cc  @ptrblck @csarofeen

apex/contrib/csrc/fmha/src/fmha/philox.h
```diff
@@ -0,0 +1,96 @@
+/***************************************************************************************************
+ * This code was copied from APEX (https://github.com/NVIDIA/apex). 
+ *
+ * APEX is released under BSD-3 license.
+ *
+ **************************************************************************************************/
+#pragma once
+// Philox CUDA.
+
+class Philox {
```
csarofeen, line 10: do we need a copy of this? CC @mcarilli

mcarilli: Looks like it's modified to do a "7-round" philox. Pytorch and curand do a "10-round" philox. No idea what the mathematical implications are, but this doesn't do the same thing as pytorch and curand philox. If they want it to do the 7 round thing they need their own copy.

seryilmaz: This also returns float4, whereas the original code returns float.

yjk21: I forgot to remove it. It's actually coming from here: https://github.com/NVIDIA/apex/blob/master/apex/contrib/csrc/multihead_attn/philox.h

seryilmaz: Right, and that one is coming from pytorch philox that @mcarilli  was referring to above.

apex/contrib/csrc/fmha/fmha_api.cpp
```diff
@@ -0,0 +1,281 @@
+/***************************************************************************************************
+ * Copyright (c) 2011-2021, NVIDIA CORPORATION.  All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without modification, are not permit-
+ * ted.
```
cliffwoolley, line 5: This is not an appropriate license for an open-source project; please fix to match APEX -- but also please ensure that doing so is permissible based on the original source of the code being contributed.

(Same comment for all files.)

cliffwoolley (changes requested):

ngoyal2707:
@yjk21
Curious what speed up this gets compared to `fast_multihead_attention`? and curious if there are plans to add `1024` sequence length?

PR #1075, opened 2021-04-13
DistributedFusedLAMB: enable no_copy and add barrier for SHARP

kexinyu:
Changes include:

- enable no_copy

- add barrier for SHARP


apex/contrib/optimizers/distributed_fused_lamb.py
```diff
@@ -87,7 +87,7 @@ def __init__(self, params,
                  step_supports_amp_scaling=True, overlap_reductions=True,
                  dwu_group_size=0, dwu_num_blocks=4, dwu_num_chunks=4,
                  dwu_num_rs_pg=1, dwu_num_ar_pg=4, dwu_num_ag_pg=0, 
-                 e5m2_allgather=False):
+                 e5m2_allgather=False, verbose=True):
```
slayton58, line 90: Should `verbose = True` by default?

PR #1079, opened 2021-04-16
Adding fast bottleneck implementation into contrib

FDecaYed:


thorjohnsen:
We need this MR in 21.05 for mlperf 1.0. I have verified that it compiles for master-py3-devel image and it is protected by a flag in setup.py, so the risk is low.

apex/contrib/csrc/bottleneck/bottleneck.cpp
```diff
@@ -0,0 +1,1634 @@
...
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a
+ * copy of this software and associated documentation files (the "Software"),
+ * to deal in the Software without restriction, including without limitation
+ * the rights to use, copy, modify, merge, publish, distribute, sublicense,
+ * and/or sell copies of the Software, and to permit persons to whom the
+ * Software is furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
```
cliffwoolley, line 12: This is MIT license.  APEX should be under BSD 3-clause license https://github.com/NVIDIA/apex/blob/master/LICENSE .  (And a header like this should be in all the files, not just this one.)

cliffwoolley (changes requested):
License headers need to be correct.

cliffwoolley:
@seryilmaz Please review.

matthew-frank (approved):
Cliff asked for an additional review.  On a quick perusal this looks good to me, but I didn't do any very careful checking.

seryilmaz:
This looks good to me. Deyu may not be able to add license headers soon since he is offline due to his timezone.

mcarilli:
Will this break import of apex in general if the extension is not built?

ptrblck:
@mcarilli No, importing apex is working without the `--fast_bottleneck` option (checked it locally).
PR is also building with the last commit.

ptrblck (approved):

PR #1080, opened 2021-04-16
initial cublaslt support for MLP

seryilmaz:
Adds cublasLt support for MLP

### DIFF
--- a/csrc/mlp.cpp
+++ b/csrc/mlp.cpp
@@ -1,32 +1,3 @@
-/*
- * Copyright (c) 2020, NVIDIA CORPORATION. All rights reserved.
- * 
- * Redistribution and use in source and binary forms, with or without
- * modification, are permitted provided that the following conditions are met:
- *
- * 1. Redistributions of source code must retain the above copyright notice, this
- *    list of conditions and the following disclaimer.
- *
- * 2. Redistributions in binary form must reproduce the above copyright notice,
- *    this list of conditions and the following disclaimer in the documentation
- *    and/or other materials provided with the distribution.
- *
- * 3. Neither the name of the copyright holder nor the names of its
- *    contributors may be used to endorse or promote products derived from
- *    this software without specific prior written permission.
- *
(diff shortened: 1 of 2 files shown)


cliffwoolley (changes requested):
Please add the correct license headers to these files.

cliffwoolley:
@matthew-frank please review.

csrc/mlp.cpp
```diff
@@ -60,9 +90,11 @@ std::vector<at::Tensor> mlp_forward(int use_bias, int activation, std::vector<at
   auto reserved_size = get_mlp_reserved_space(batch_size, num_layers, output_features.data());
 
   // create output/workspace tensor
-  // TODO(deyuf): just get buffer?
+  // TODO(deyuf): just get buffer? why empty with .type() gets GPU tensor?
```
matthew-frank, line 93: is there an open TODO here?  If so, what?  If not, remove the TODO?

seryilmaz: removed now.

csrc/mlp.cpp
```diff
@@ -1,10 +1,39 @@
+/*
+ * Copyright (c) 2020, NVIDIA CORPORATION. All rights reserved.
```
ptrblck, line 2: Should this be changed to 2021?

matthew-frank (approved):

csrc/mlp.cpp
cliffwoolley: At least add the -2021.  We can come back later and delete the 2020- part if it's inaccurate.

cliffwoolley (approved):

LSinev:
@seryilmaz Please consider restoring support for cuda 10.0 which do not have cublaslt.
Installation instruction
```
pip install -v --disable-pip-version-check --no-cache-dir --global-option="--cpp_ext" --global-option="--cuda_ext" ./
```
is not working now with cuda 10.0, and removing `--cuda_ext` turns off ability to use FusedNovograd which was usable before this improvement
```
RuntimeError: apex.optimizers.FusedNovoGrad requires cuda extensions
```

PR #1083, opened 2021-04-19
Fix cublasLt context create/destroy overhead in MLP extension

seryilmaz:
@FDecaYed  please review.

### DIFF
--- a/csrc/mlp_cuda.cu
+++ b/csrc/mlp_cuda.cu
@@ -718,7 +718,7 @@ void get_biasAddRelu_bprop_grid_size(
   // Get number of SMs for efficient reduction.
   int num_SMs = at::cuda::getCurrentDeviceProperties()->multiProcessorCount;
   // can switch to occupancy calculation. use 4 below now for sm_70
-  int max_blocks_y = num_SMs * 4 / (*grid_x);
+  int max_blocks_y = (num_SMs * 4+(*grid_x)-1) / (*grid_x);
   // block_y should be from minimal work per thread
   int nRedSplits = (batch_size + block_y - 1) / block_y;
   // increase number of elem per thread redcution to not launch more than enough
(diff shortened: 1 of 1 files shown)


csrc/mlp_cuda.cu
```diff
@@ -1274,28 +1271,29 @@ int mlp_fp(
...
+          &one,
+          weight,
+          ifeat,
+          input,
+          ifeat,
+          &zero,
+          output,
+          ofeat,
+          lt_workspace,
+          1 << 22,
```
cliffwoolley, line 1292: what's with the magic number here?  Is it the same magic number as is at https://github.com/seryilmaz/apex/blob/seryilmaz/mlp_fixes2/csrc/mlp.cpp#L67 ?  Can those please be combined and given a name if so?

ptrblck (approved):

csrc/mlp_cuda.cu
FDecaYed: Yeah they are the same number. The idea at the time is to use something large enough for small-ish gemm, since those are the sizes mlp extension targeting to help, while also not too big so we don't care enough to try shrink and save.
agree with given it a name

csrc/mlp_cuda.cu
```diff
@@ -1274,28 +1271,29 @@ int mlp_fp(
...
-        &zero,
-        output,
-        ofeat,
-        lt_workspace,
-        1 << 22,
-        stream,
-        use_bias == 1,
-        activation == 1,
-        bias);
+    if(activation < 1){
```
ngoyal2707, line 1274: Curious was this intentional to change it to `activation < 1` instead of `activation < 2` cause now cublaslt wont get hit for relu, and this seems to kill performance for me quite a lot

FDecaYed: sorry to get back to you so late @ngoyal2707.
This is intentional due to some perf issue. We're working on it and we'll re-enable relu fusion once the issue get resolved.

PR #1108, opened 2021-05-16
compile cublasLt code only for cublas >= 11.0

seryilmaz:
Fixes compilation issue with cublas version < 11 by selectively compiling for cublasLt for >=11 version only.


ptrblck:
Thanks for the fix! :)

Issue #913, opened 2020-07-16
local variable 'count_all' referenced before assignment error


kyle5794:
I am using apex installed from the latest commit the master branch.

I am getting an error: `local variable 'count_all' referenced before assignment` from https://github.com/NVIDIA/apex/blob/master/apex/parallel/optimized_sync_batchnorm_kernel.py#L63 when running my model in evaluation mode

I think during evaluation mode, `track_running_stats` will be set to False which causes `count_all` not to be initialized since `count_all` is initialized under `if track_running_stats` block

SmithLiu95:
I am encountering the same issue

58733511:
the same

taey16:
the same as well

GuodongQi:
the same as well

yi-ming-lin:
same

Xiangyu-CAS:
the same

Noir97:
same issue here

YinAoXiong:
same issue here

YinAoXiong:
@kyle5794 Have you found a solution to the problem? 😂

kyle5794:
> @kyle5794 Have you found a solution to the problem? joy

I think that the simplest solution is to checkout the previous commit.

YinAoXiong:
> > @kyle5794 Have you found a solution to the problem? joy
> 
> I think that the simplest solution is to checkout the previous commit.

Thank you, I will try later. I found that not using syncbn the model also works well

Alec-Lin:
If you use BatchNorm1d? I meet this problem when I use BatchNorm1d. But when I replace it to BatchNorm2d, the code runs correctly.
I guess SyncBatchNorm in apex do not support BatchNorm1d.

Xiangyu-CAS:
@Alec-Lin I think you are right, I use BN 1d in my network too. I solved the problem by reseting it back to previous version, before commit 1ff54b8

richardwth:
I had the same issue when running `bash script_A.sh` that runs `python script_B.py`, but the issue magically disappears when running the `python script_C.py` that calls a shell script via `os.system("bash scriptA.sh")` which eventually runs `python script_B.py`.

richardwth:
I had the same issue, but it magically disappeared when I add `cd path_to_working_dir` at the top of the shell script that sets up the environment and calls the python scripts to train the model. I guess it is not really an issue of the local variable `count_all` but an issue of how relative paths are handled.

[...]

PR #1150, opened 2021-08-30
Fix Issue #913

JinwooWang:
Fix the error: "local variable 'count_all' referenced before assignment" during evaluation mode.
Tested on hardware configured with 3 GTX1080Ti: training an EfficientnetB5 with performing evaluation every 100 iterations using all three GPUs.


minostauros:
Thanks a lot!
CC @crcrpar @ptrblck

Cucaracha18:
I have a question about track_runnung_stats == false

jimmy19991222:
Thanks！

PR #1152, opened 2021-08-31
option to set param views to flat buffer

seryilmaz:
Separates flat param buffer creation and autograd hook creation to allow setting parameter views to the flat buffer in between. Parameter reordering is disabled for this option to make sure QKV GEMM parameters are in order.  
This MR also adds the option to skip allgather to handle it on user side.


apex/contrib/optimizers/distributed_fused_lamb.py
```diff
@@ -266,13 +256,60 @@ def allreduce_hook(*unused):
...
+    def _lazy_init_stage1(self):
+        if self._lazy_init_stage1_done: return
+
+        p_offset = 0
+        p_i = 0
+        #self._model_params = []
+        #self._grad_accs = []
+        #self._group_properties = []
+        for group in self.param_groups:
+            prev = None
```
kexinyu, line 272: prev, beta, bias_correction, eps, weight_decay, p_offset not used and needed in this for loop

seryilmaz: thanks! Just removed those.

Issue #1155, opened 2021-09-01
Apex installation failed


Teng-xu:
I was trying to install apex through dockerfile via the following commands
```
RUN git clone https://github.com/NVIDIA/apex && \
        cd apex && \
        pip install -v --no-cache-dir --global-option="--cpp_ext" --global-option="--cuda_ext" ./
```

and I got the following errors, it was able to build 2 days ago, but it fails now and the failure seems to be related to `fused_dense_cuda.cu`
```
[0m[91m    csrc/fused_dense_cuda.cu(415): error: identifier "CUBLASLT_EPILOGUE_GELU_AUX" is undefined

    csrc/fused_dense_cuda.cu(427): error: identifier "CUBLASLT_MATMUL_DESC_EPILOGUE_AUX_POINTER" is undefined

[0m[91m    csrc/fused_dense_cuda.cu(428): error: identifier "CUBLASLT_MATMUL_DESC_EPILOGUE_AUX_LD" is undefined

    csrc/fused_dense_cuda.cu(435): error: identifier "CUBLASLT_EPILOGUE_GELU_AUX_BIAS" is undefined

[0m[91m    csrc/fused_dense_cuda.cu(555): error: identifier "CUBLASLT_EPILOGUE_GELU_AUX" is undefined

    csrc/fused_dense_cuda.cu(567): error: identifier "CUBLASLT_MATMUL_DESC_EPILOGUE_AUX_POINTER" is undefined
[0m[91m
    csrc/fused_dense_cuda.cu(568): error: identifier "CUBLASLT_MATMUL_DESC_EPILOGUE_AUX_LD" is undefined

[0m[91m    csrc/fused_dense_cuda.cu(575): error: identifier "CUBLASLT_EPILOGUE_GELU_AUX_BIAS" is undefined
[0m[91m
[0m[91m    csrc/fused_dense_cuda.cu(687): error: identifier "CUBLASLT_EPILOGUE_BGRADB" is undefined

[0m[91m    csrc/fused_dense_cuda.cu(826): error: identifier "CUBLASLT_EPILOGUE_BGRADB" is undefined

[0m[91m    csrc/fused_dense_cuda.cu(920): error: identifier "CUBLASLT_EPILOGUE_DGELU_BGRAD" is undefined
[0m[91m
[0m[91m    csrc/fused_dense_cuda.cu(936): error: identifier "CUBLASLT_MATMUL_DESC_EPILOGUE_AUX_POINTER" is undefined
[...]

PR #1162, opened 2021-09-03
fix CUBLAS guards

seryilmaz:


crcrpar:
rel: https://github.com/NVIDIA/apex/issues/1155

csrc/fused_dense.cpp
```diff
@@ -62,7 +62,7 @@ std::vector<at::Tensor> linear_bias_backward(at::Tensor input, at::Tensor weight
 
   // create output/workspace tensor
   auto d_weight = at::empty({out_features, in_features}, input.type());
-#if defined(CUDA_VERSION) && CUDA_VERSION < 11000
+#if defined(CUDA_VERSION) && CUDA_VERSION < 11600
```
ptrblck, line 65: I assume you would like to guard against CUDA11.4 (if so, then you would have to use `11040`) or against the `CUBLAS_VERSION`?

seryilmaz: Updated it now. It is tricky to guard against CUDA version since different epilogue constants in CUBLAS are added at different versions while still using CUDA 11.4.

cliffwoolley: Note: Compile-time guards checking for new features may still require runtime guards and fallbacks, since it's legal to build against a newer version than you run with.

PR #1172, opened 2021-09-22
Fix typo in contrib FusedLamb.

romerojosh:
There was a typo (`dytpe` instead of `dtype`) in the contrib FusedLamb that would only cause error in cases with FP16 gradients (which are uncommon).


kevinstephano (approved):
Trivial typo fix likely missed because Native AMP only exercises the FP32 path and APEX AMP O2 also only uses the FP32 path.

PR #1179, opened 2021-09-29
Fix compilation for cuda 9 by setting correct compute capabilities

iseessel:
I was trying to compile w/ pytorch 1.6.0 and CUDA 9.2 and was getting error:

```
nvcc fatal : Unsupported gpu architecture 'compute_75'
```

Cuda 9 does not support compute_75, hence specify only versions below this.


iseessel:
@ptrblck @crcrpar Can you review?

PR #1185, opened 2021-10-07
minor fix for dist lamb

nanz-nv:
This PR fix two problems.
1. remove the weight broadcast in the constructor
Currently there are two weight broadcasts in the code: one in the constructor, and the other in `_lazy_init_stage1`. One of this broadcast is redundant. The fix is to remove the one in the constructor.

2. disable unnecessary allreduces for clip-after-ar
For clip-after-ar, the L2 norm is already obtained through all reduce, therefore there is no need to do a allreduce on `is_finite` and `self._overflow_buf` in `_pipeline_step`


nanz-nv:
@ptrblck , could you please review and merge this PR?

PR #1205, opened 2021-11-09
Update parallel_state with Megatron pipeline parallel size and embedding rank

eqy:
Two misc updates to `parallel_state.py`:
+  `is_rank_in_embedding_group` which appears to be necessary for models like BERT.
+ check that the pipeline-model-parallel world size is greater than 2 when interleaving is enabled (to mirror Megatron constraint):

https://github.com/NVIDIA/Megatron-LM/blob/b31e1296354e979722627a6c4dedafe19b51fa97/megatron/arguments.py#L132

CC @crcrpar @ptrblck


crcrpar (approved):
thank you for doing this

PR #1216, opened 2021-11-12
minimal bert pipeline parallel test

eqy:
CC @crcrpar @ptrblck

### DIFF
--- a/apex/transformer/testing/standalone_bert.py
+++ b/apex/transformer/testing/standalone_bert.py
@@ -1,0 +1,221 @@
+import torch
+
+from apex.normalization import FusedLayerNorm as LayerNorm
+from apex.transformer import tensor_parallel
+from apex.transformer.testing.global_vars import get_args
+from .standalone_gpt import get_language_model, get_linear_layer, init_method_normal, parallel_lm_logits, scaled_init_method_normal
+from .standalone_gpt import AttnMaskType, MegatronModule
+
+def bert_extended_attention_mask(attention_mask):
+    # We create a 3D attention mask from a 2D tensor mask.
+    # [b, 1, s]
+    attention_mask_b1s = attention_mask.unsqueeze(1)
+    # [b, s, 1]
+    attention_mask_bs1 = attention_mask.unsqueeze(2)
+    # [b, s, s]
+    attention_mask_bss = attention_mask_b1s * attention_mask_bs1
+    # [b, 1, s, s]
+    extended_attention_mask = attention_mask_bss.unsqueeze(1)
+
+    # Convert attention mask to binary:
+    extended_attention_mask = (extended_attention_mask < 0.5)
+
+    return extended_attention_mask
+
+def bert_position_ids(token_ids):
+    # Create position ids
+    seq_length = token_ids.size(1)
+    position_ids = torch.arange(seq_length, dtype=torch.long,
+                                device=token_ids.device)
+    position_ids = position_ids.unsqueeze(0).expand_as(token_ids)
+
+    return position_ids
+
+class BertLMHead(MegatronModule):
+    """Masked LM head for Bert
+
+    Arguments:
+        mpu_vocab_size: model parallel size of vocabulary.
+        hidden_size: hidden size
+        init_method: init method for weight initialization
+        layernorm_epsilon: tolerance for layer norm divisions
+        parallel_output: whether output logits being distributed or not.
+    """
+
+    def __init__(self, mpu_vocab_size, hidden_size, init_method,
+                 layernorm_epsilon, parallel_output):
+
+        super(BertLMHead, self).__init__()
+
+        args = get_args()
+
+        self.bias = torch.nn.Parameter(torch.zeros(mpu_vocab_size))
+        # TODO: do we need this?
+        # mpu.set_tensor_model_parallel_attributes(self.bias, True, 0, 1)
+        self.parallel_output = parallel_output
+
+        self.dense = get_linear_layer(hidden_size, hidden_size, init_method)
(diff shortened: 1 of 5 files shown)


apex/transformer/testing/standalone_bert.py
```diff
@@ -0,0 +1,221 @@
+import torch
+
+from apex.normalization import FusedLayerNorm as LayerNorm
+from apex.transformer import tensor_parallel
+from apex.transformer.testing.global_vars import get_args
+from .standalone_gpt import get_language_model, get_linear_layer, init_method_normal, parallel_lm_logits, scaled_init_method_normal
+from .standalone_gpt import AttnMaskType, MegatronModule
```
crcrpar, line 7: apex has enums in https://github.com/NVIDIA/apex/blob/master/apex/transformer/enums.py.
Could you check if the enums.py is au courant with Megatron's def.s and use apex's if possible?

apex/transformer/testing/standalone_bert.py
```diff
@@ -0,0 +1,221 @@
...
+        # Load word_embeddings.
+        if self.post_process and not self.pre_process:
+            self.word_embeddings.load_state_dict(
+                state_dict[self._word_embeddings_for_head_key], strict=strict)
+
+
+def gpt_model_provider(pre_process=True, post_process=True):
+    model = GPTModel(num_tokentypes=0, parallel_output=True, pre_process=pre_process, post_process=post_process)
+    return model
+
```
crcrpar, line 218: [ultra nit-picking] Seems like we can move this function back to standalone_gpt.py.

tests/L0/run_transformer/run_bert_minimal_test.py
```diff
@@ -0,0 +1,168 @@
...
+        output_tensor, _ = output_tensor
+        lm_loss_ = output_tensor.float()
+        lm_loss = torch.sum(
+                        lm_loss_.view(-1) * loss_mask.reshape(-1)) / loss_mask.sum()
+        averaged_loss = average_losses_across_data_parallel_group([lm_loss])
+        if data_idx >= 1536:
+            assert lm_loss < 4.8
+            if not ONCE:
+                print("LOSS OK")
+                ONCE = True
```
crcrpar, line 95: I like this. How many iterations (= mini batches) does one epoch have, out of curiosity?

tests/L0/run_transformer/run_bert_minimal_test.py
```diff
@@ -0,0 +1,168 @@
...
+    effective_length = fancy_data.size(0) // global_vars.get_args().seq_length
+    effective_length = fancy_data.size(0) - global_vars.get_args().seq_length
+
+
+    initialize_distributed()
+    world_size = torch.distributed.get_world_size()
+    failure = None
+    try:
+        args = global_vars.get_args()
+        args.padded_vocab_size = 128 # needed in standalone gpt
```
crcrpar, line 128: [nit-picking] is this `padded_vocab_size` needed in Bert? If so, could you set this value via `global_vars.set_global_variables()` like https://github.com/NVIDIA/apex/blob/9b7d2ac080d7aaa3d44622b8b0cd7a0299e43ea5/tests/L0/run_transformer/run_dynamic_batchsize_test.py#L33-L36?

crcrpar (commented):
Looks great overall

apex/transformer/testing/standalone_bert.py
eqy: Good catch, need to cleanup

tests/L0/run_transformer/run_bert_minimal_test.py
eqy: It looks like this requires the argument to already be defined; here the hack is because we're skipping the tokenizer which usually sets it: https://github.com/NVIDIA/Megatron-LM/blob/42c1cf4279acea5a554500dcb552211f44cbec45/megatron/tokenizer/tokenizer.py#L25

tests/L0/run_transformer/run_bert_minimal_test.py
eqy: I believe it would be around `710797//4` (total positions divided by microbatch size).

tests/L0/run_transformer/run_bert_minimal_test.py
crcrpar: okay.

tests/L0/run_transformer/run_bert_minimal_test.py
```diff
@@ -0,0 +1,168 @@
...
+                ONCE = True
+        return lm_loss, {'avg': averaged_loss}
+    return y, loss_func
+
+
+def train(model, optim):
+    sequence_len = global_vars.get_args().seq_length
+    micro_batch_size = global_vars.get_args().micro_batch_size
+    hidden_size = global_vars.get_args().hidden_size
+    forward_backward_func = _forward_backward_pipelining_with_interleaving
```
crcrpar, line 104: not a big deal, but how about calling `apex.transformer.pipeline_parallel.get_forward_backward_func` to get this interleaving schedule function?

- it's exposed here https://github.com/NVIDIA/apex/blob/25bfcb914796c7955956fc1e39839e7efed997d5/apex/transformer/pipeline_parallel/__init__.py#L6
- signature is https://github.com/NVIDIA/apex/blob/25bfcb914796c7955956fc1e39839e7efed997d5/apex/transformer/pipeline_parallel/schedules/__init__.py#L16-L18

eqy:
One thing I discovered is that the BERT minimal tests doesn't seem to like `--use-cpu-initialization`, not sure if this is a known issue or limitation...

PR #1218, opened 2021-11-14
Use `torch.float32` for p2p communications

crcrpar:
Per title.

As in the docstring, we don't have means to tell the dtype of an activation torch.Tensor (meaning the output of `nn.Module.forward` on different devices without
a strong assumption if we are in the context of `torch.cuda.amp.autocast`.


crcrpar:
oh I completely forgot that #1212 made this change

PR #1222, opened 2021-11-18
Minimal gpt pipeline parallel (builds off of minimal_bert_pipeline_parallel) including cpu-offloading

puririshi98:
Based on #1216

### DIFF
--- a/apex/transformer/pipeline_parallel/schedules/fwd_bwd_pipelining_without_interleaving.py
+++ b/apex/transformer/pipeline_parallel/schedules/fwd_bwd_pipelining_without_interleaving.py
@@ -73,7 +73,6 @@ def forward_backward_pipelining_without_interleaving(
         f"num_microbatches_remaining: {num_microbatches_remaining}"
     )
 
-
     # Input, output tensors only need to be saved when doing backward passes
     input_tensors = None
     output_tensors = None
--- a/apex/transformer/pipeline_parallel/schedules/fwd_bwd_pipelining_without_interleaving.py
+++ b/apex/transformer/pipeline_parallel/schedules/fwd_bwd_pipelining_without_interleaving.py
@@ -67,13 +67,11 @@ def forward_backward_pipelining_without_interleaving(
     num_warmup_microbatches = min(num_warmup_microbatches, num_microbatches)
     num_microbatches_remaining = num_microbatches - num_warmup_microbatches
 
-    # TODO (mkozuki): Remove once debug gets done
-    # print(
-    #     f">>> rank: {torch.distributed.get_rank()}, "
-    #     f"num_microbatches: {num_microbatches}, "
-    #     f"num_warmup_microbatches: {num_warmup_microbatches}, "
-    #     f"num_microbatches_remaining: {num_microbatches_remaining} -- "
-    # )
+    _logger.info(
+        f"num_microbatches: {num_microbatches}, "
+        f"num_warmup_microbatches: {num_warmup_microbatches}, "
+        f"num_microbatches_remaining: {num_microbatches_remaining}"
+    )
 
 
     # Input, output tensors only need to be saved when doing backward passes
--- a/tests/L0/run_transformer/gpt_scaling_test.py
+++ b/tests/L0/run_transformer/gpt_scaling_test.py
@@ -48,7 +48,7 @@ def plot(runtimes):
 
 def main():
 	runtimes = {}
-	for data_parr, tens_parr, pipe_parr in [(8,1,1), (4,2,1), (2,1,4), (1,2,4), (1,1,8)]:
+	for data_parr, tens_parr, pipe_parr in [(8,1,1), (4,2,1), (2,1,4), (1,2,4)]:
 		dist_setting = 'ddp=' + str(data_parr) + ', tensor_parr=' + str(tens_parr) + ', pipe_parr=' + str(pipe_parr)
 		runtimes[dist_setting] = {} 
 		print("Beginning Testing for", dist_setting)
--- a/tests/L0/run_transformer/gpt_scaling_test.py
+++ b/tests/L0/run_transformer/gpt_scaling_test.py
@@ -71,7 +71,7 @@ def main():
 			else:
 				print("GPT-2 w/", n, "layers failed using", dist_setting)
 				print("Moving on to the next distributed setting...")
-				print("#*25")
+				print("#"*(25))
 				print()
 				break
 	print(runtimes)
--- a/tests/L0/run_transformer/gpt_scaling_test.py
+++ b/tests/L0/run_transformer/gpt_scaling_test.py
@@ -44,7 +44,7 @@ def run_gpt(cmd):
 				quit()
 		if str(out) == str(TEST_SUCCESS_MESSAGE):
 			success=True
-	return runtime, float(int(init_dict['num_params']))/10.0**9, success, errs
+	return runtime, round(float(int(init_dict['num_params']))/10.0**9,2), success, errs
 
 
 def plot(runtimes):
@@ -74,7 +74,7 @@ def main():
 			runtime, bill_params, success, errs = run_gpt(cmd)
 			if success:
 				runtimes[dist_setting][bill_params] = runtime
-				print(runtime, 'ms')
+				print(runtime, 's per training iter for ', str(bill_params) + 'B parameter GPT-2')
 			else:
 				print("GPT-2 w/", n, "layers failed using", dist_setting)
 				if sys.argv[1] == '-v' or sys.argv[1] == '--verbose':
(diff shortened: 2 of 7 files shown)


tests/L0/run_transformer/run_gpt_minimal_test.py
```diff
@@ -0,0 +1,170 @@
...
+    sequence_len = global_vars.get_args().seq_length
+    micro_batch_size = global_vars.get_args().micro_batch_size
+    hidden_size = global_vars.get_args().hidden_size
+    fwd_bwd_func = forward_backward_pipelining_without_interleaving
+
+    tensor_shape = (args.seq_length, args.micro_batch_size, args.hidden_size)
+    for i in range(8):
+      if torch.distributed.get_rank() == 0:
+        print('begin iter', i)
+      batch = [generate_fancy_data_labels(args.seq_length, args.global_batch_size) for _ in range(pipeline_model_parallel_size)]
```
eqy, line 114: Is this intentionally changed fro the BERT-script? We switched to using separate indices for each virtual pipeline stage and each "real" pipeline stage should get its own batch implicitly as they are separate processes.

tests/L0/run_transformer/run_gpt_minimal_test.py
```diff
@@ -0,0 +1,170 @@
...
+    return output_tensor, partial(loss_func, loss_mask)
+
+
+def train(model, optim, pipeline_model_parallel_size):
+    sequence_len = global_vars.get_args().seq_length
+    micro_batch_size = global_vars.get_args().micro_batch_size
+    hidden_size = global_vars.get_args().hidden_size
+    fwd_bwd_func = forward_backward_pipelining_without_interleaving
+
+    tensor_shape = (args.seq_length, args.micro_batch_size, args.hidden_size)
```
eqy, line 110: For debugging, it is very clunky, but I would recommend comparing the `tensor_shape` here and the outputs of the intermediate pipeline stages to make sure they match. If data loading is a problem (e.g., the forward backward step runs past the end of the current batch), the shapes would reflect this by having zeros in some dimensions.

tests/L0/run_transformer/run_gpt_minimal_test.py
puririshi98: i changed it to fit the shape requirements of the fwd_bwd_func being used in masakis gpt example:
```
def get_batch(int_tensors: List[torch.Tensor]):
    data = int_tensors[0]
    # Unpack.
    tokens_ = data.long()
    labels = tokens_[:, 1:].contiguous()
    tokens = tokens_[:, :-1].contiguous()
    # Get the masks and position ids.
    attention_mask, loss_mask, position_ids = get_ltor_masks_and_position_ids(
        tokens,
        N_VOCAB,  # tokenizer.eod,
        False,  # args.reset_position_ids,
        False,  # args.reset_attention_mask,
        False,  # args.eod_mask_loss,
    )
    return tokens, labels, loss_mask, attention_mask, position_ids


# Ref: https://github.com/NVIDIA/Megatron-LM/blob/b31e1296354e979722627a6c4dedafe19b51fa97/pretrain_gpt.py#L75
def loss_func(loss_mask, output_tensor):
    losses = output_tensor.float()
    loss_mask = loss_mask.view(-1).float()
    loss = torch.sum(losses.view(-1) * loss_mask) / loss_mask.sum()

    # Reduce loss for logging.
    averaged_loss = average_losses_across_data_parallel_group([loss])

    return loss, {'lm loss': averaged_loss[0]}


# Ref: https://github.com/NVIDIA/Megatron-LM/blob/b31e1296354e979722627a6c4dedafe19b51fa97/pretrain_gpt.py#L86
# TODO (mkozuki): Currently I'm seeing no attribute `word_embeddings` which looks weird.
def fwd_step_func(batch, model):
    """Forward step."""
    tokens, labels, loss_mask, attention_mask, position_ids = get_batch(batch)
    output_tensor = model(tokens, position_ids, attention_mask, labels=labels)
    return output_tensor, partial(loss_func, loss_mask)
```

tests/L0/run_transformer/run_gpt_minimal_test.py
```diff
@@ -0,0 +1,172 @@
...
+    if inds is None or data_idx >= len(inds):
+       # hack as use of RNG will fall out of sync due to pipelines being different
+      model_parallel_cuda_manual_seed(MANUAL_SEED)
+      inds = torch.randperm(effective_length, device='cuda')
+      MANUAL_SEED += 1
+      data_idx = 0
+    data_idx_ = data_idx
+    offset = inds[data_idx_] #* SEQUENCE_LEN
+    data_idx += 1
+    curr = fancy_data[offset:offset+sequence_len].clone().detach()
```
crcrpar, line 59: I think the following suggestion is needed because `get_batch` function curtails the sequence by one element in [L.70](https://github.com/NVIDIA/apex/pull/1222/files#diff-c300b35d6e57bca85106f7aac069ff5856e5ee68966b305fbb9bb55aef5d3f20R70) & [L.71](https://github.com/NVIDIA/apex/pull/1222/files#diff-c300b35d6e57bca85106f7aac069ff5856e5ee68966b305fbb9bb55aef5d3f20R71), i.e. the "sequence length" dim of `tokens` and `labels` is unexpectedly `args.seq_length - 1`
```suggestion
    curr = fancy_data[offset:offset+sequence_len + 1].clone().detach()
```

tests/L0/run_transformer/run_gpt_minimal_test.py
eqy: If it works it works, but I would be curious if using a single item would work here...

tests/L0/run_transformer/gpt_scaling_test.py
```diff
@@ -0,0 +1,85 @@
...
+import numpy as np
+import subprocess
+from apex.transformer.testing.commons import TEST_SUCCESS_MESSAGE
+import json
+
+def set_seed():
+	random.seed(0)
+	np.random.seed(0)
+	torch.manual_seed(0)
+	torch.cuda.manual_seed_all(0)
```
crcrpar, line 18: [Question]
Doesn't this function need to call `apex.transformer.tensor_parallel.model_parallel_cuda_manual_seed(seed)`?
If this doesn't need to do so, could you tell me why?

apex/transformer/pipeline_parallel/schedules/fwd_bwd_pipelining_without_interleaving.py
```diff
@@ -67,11 +67,14 @@ def forward_backward_pipelining_without_interleaving(
...
-        f"num_microbatches_remaining: {num_microbatches_remaining}"
-    )
+    # TODO (mkozuki): Remove once debug gets done
+    # print(
+    #     f">>> rank: {torch.distributed.get_rank()}, "
+    #     f"num_microbatches: {num_microbatches}, "
+    #     f"num_warmup_microbatches: {num_warmup_microbatches}, "
+    #     f"num_microbatches_remaining: {num_microbatches_remaining} -- "
+    # )
+
```
crcrpar, line 77: Could you revert these changes?

tests/L0/run_transformer/gpt_scaling_test.py
```diff
@@ -0,0 +1,82 @@
...
+			if success:
+				runtimes[dist_setting][bill_params] = runtime
+				print(str(runtime) + 's per training iter for', str(bill_params) + 'B parameter GPT-2')
+			else:
+				print("GPT-2 w/", n, "layers failed using", dist_setting)
+				print("Moving on to the next distributed setting...")
+				break
+	print(runtimes)
+	plot(runtimes)
+main()
```
crcrpar, line 82: I don't think this file is imported somewhere else but to make sure that this test is executed only when we run this script directly.

```suggestion

if __name__ == "__main__":
    main()
```

tests/L0/run_transformer/gpt_scaling_test.py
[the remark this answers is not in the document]
eqy: On a single node, for data loading/generation, it may be sufficient for smoke testing though `model_parallel_cuda_manual_seed` is more comprehensive.

tests/L0/run_transformer/run_gpt_minimal_test.py
puririshi98:
```
      # batch = [generate_fancy_data_labels(args.seq_length, args.global_batch_size) for _ in range(pipeline_model_parallel_size)]
      batch = generate_fancy_data_labels(args.seq_length, args.global_batch_size)
```
i can confirm that making the above switch breaks things

crcrpar (commented):
Feels like it better to run `run_gpt_minimal_test.py` with multiple GPUs so could you 
add the substring of this filename e.g. `"gpt_minimal_test"` to https://github.com/NVIDIA/apex/blob/2ff06d0e73be710929333abd14e36e54044ed3ad/tests/L0/run_transformer/test_transformer_module.py#L15-L17?

crcrpar (approved):

PR #1229, opened 2021-11-25
[FMHA] add support for later CUDA (8.x)

jqueguiner:
add support for later minor dprops 8.0+ for CUDA DEVICES such as A40


crcrpar:
@jqueguiner Thanks for the PR but I'm afraid we need to relax the condition in backward path as well in https://github.com/NVIDIA/apex/blob/f7421555c3d2ff01eed0e7c0c4321f3e4dd58fc6/apex/contrib/csrc/fmha/fmha_api.cpp#L182-L191.

Thank you for your patience in my delayed response.

jqueguiner:
@crcrpar : thanks for the feedback; should be fixed now.

thanks

jqueguiner:
Hi @crcrpar  any update regarding this matter ?

crcrpar:
I found that this FMHA failed when I ran https://github.com/NVIDIA/apex/blob/master/apex/contrib/test/fmha/test_fmha.py with this PR on A40 because the kernel might violate some limitation unfortunately...

PR #1235, opened 2021-11-30
Bind the new_fwd method to the model during amp.initialize

baijumeswani:
During `amp.initialize()`, the patched forward is directly assigned to the model without binding it using `MethodType`.
This resulted in an error while trying to access the `__func__` attribute of the forward method of the model.

`AttributeError: 'function' object has no attribute '__func__'`

Minor change in this pull request just binds the patched forward method to the model.

### DIFF
--- a/apex/amp/_initialize.py
+++ b/apex/amp/_initialize.py
@@ -192,13 +192,13 @@ def _initialize(models, optimizers, properties, num_losses=1, cast_model_outputs
             # outgoing data to float32, so "the user never needs to call .half()."
             # I like writing things explicitly more than decorators.
             def patch_forward(old_fwd):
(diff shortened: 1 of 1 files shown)


baijumeswani:
Any update on this? I'm hoping to have this reviewed so we can avoid using workarounds in user code.

baijumeswani:
Hello, wanted to follow up to see if this can be reviewed or looked at?

ptrblck:
Hi @baijumeswani, 
thanks for creating the PR! `apex.amp` is deprecated and not supported anymore (see https://github.com/NVIDIA/apex/issues/818).
Please use the native amp implementation via `torch.cuda.amp` as described [here](https://pytorch.org/docs/stable/amp.html).

Closing.

PR #1237, opened 2021-12-01
Add fused mixed precision lamb optimizer.

kevinstephano:


apex/optimizers/fused_mixed_precision_lamb.py
```diff
@@ -0,0 +1,251 @@
...
+            bias_correction = 1 if group['bias_correction'] else 0
+            beta1, beta2 = group['betas']
+            grad_averaging = 1 if group['grad_averaging'] else 0
+
+            # assume same step across group now to simplify things
+            # per parameter step can be easily support by making it tensor, or pass list into kernel
+            if 'step' in group:
+                group['step'] += (self._dummy_overflow_buf != 1).int()
+            else:
+                group['step'] = (self._dummy_overflow_buf != 1).int()
```
crcrpar, line 193: Since `step` is included in `defaults`, we can safely assume each group has `"step"`?
Also, following the `test_multi_device` failure, might better to cast using this `.to()`.
 
```suggestion
            # assume same step across group now to simplify things
            # per parameter step can be easily support by making it tensor, or pass list into kernel
            group['step'] += (self._dummy_overflow_buf != 1).to(dtype=torch.int, device=device)
```

apex/optimizers/fused_mixed_precision_lamb.py
```diff
@@ -0,0 +1,251 @@
...
+                 amsgrad=False, adam_w_mode=True,
+                 grad_averaging=True, max_grad_norm=1.0, use_nvlamb=False,
+                 reduced_precision_dtype=None):
+        if amsgrad:
+            raise RuntimeError('FusedLAMB does not support the AMSGrad variant.')
+
+        # The learning rate (lr) and optimizer step (step) should be located on device
+        # in order to faciliated device sync free execution
+        defaults = dict(lr=torch.tensor(lr, dtype=torch.float32, device=torch.cuda.current_device()),
+                        step=torch.tensor([step], dtype=torch.int, device=torch.cuda.current_device()),
```
crcrpar, line 21: Using `torch.cuda.current_device()` can lead to the failure of [`test_multi_device`](https://github.com/NVIDIA/apex/pull/1237/files#diff-f16fafddd5bfd009eaf6d3fec48df46f50326c2d9a25c829d6e544fee49447e1R289).

A fix might be to send these two tensors to the device [here](https://github.com/NVIDIA/apex/pull/1237/files#diff-fe6f34523dea6871f255b052775b32df8f6af5bae5dae0d2497f190b60c24beaR156) lazily in `step` method.

apex/optimizers/fused_mixed_precision_lamb.py
kevinstephano: Good point, I removed the conditional and changed the casting to use `to()`

apex/optimizers/fused_mixed_precision_lamb.py
kevinstephano: I used the device of the parameters that are passed in and that seems to work.

kevinstephano: The mixed device test now passes.

apex/optimizers/fused_mixed_precision_lamb.py
```diff
@@ -0,0 +1,250 @@
...
+                 reduced_precision_dtype=None):
+        if amsgrad:
+            raise RuntimeError('FusedLAMB does not support the AMSGrad variant.')
+        
+        device = params[0].device
+
+        # The learning rate (lr) and optimizer step (step) should be located on device
+        # in order to faciliated device sync free execution
+        defaults = dict(lr=torch.tensor(lr, dtype=torch.float32, device=device),
+                        step=torch.tensor([step], dtype=torch.int, device=device),
```
crcrpar, line 23: I'm sorry I completely missed that `params` can be a `dict`.
For example, https://github.com/NVIDIA/DeepLearningExamples/blob/fd9fecd2b22e6b9e25e75de8b0a90a711cf91477/PyTorch/LanguageModeling/BERT/run_pretraining.py#L376-L38484

So moving `lr` and `step` tensors after `super().__init__(params, defaults)` because we'll have `param_groups` attribute for sure.
```suggestion

        # The learning rate (lr) and optimizer step (step) should be located on device
        # in order to faciliated device sync free execution.
        # That being said, `params` is either `dict` or `list`
        # so we're sending these two tensors to device after `super().__init__()`.
        defaults = dict(lr=torch.tensor(lr, dtype=torch.float32),
                        step=torch.tensor([step], dtype=torch.int),
```

apex/optimizers/fused_mixed_precision_lamb.py
```diff
@@ -0,0 +1,250 @@
...
+        else:
+            raise RuntimeError('apex.optimizers.FusedLAMB requires cuda extensions')
+
+        # Mixed Precision support
+        self.reduced_precision_dtype = reduced_precision_dtype
+        self.param_groups_full_precision = []
+        
+        self._step_supports_amp_scaling = True
+        self.adam_w_mode = 1 if adam_w_mode else 0
+        self.use_nvlamb = use_nvlamb
```
crcrpar, line 44: So now that we have `param_groups`, we can move `lr` and `step` to GPU with an assumption that all param_groups are hosted on the same GPU device (the same assumption as the existing FusedLAMB I believe according to https://github.com/NVIDIA/apex/blob/aa756cec4359aff3df1d9abb68dc6e6e92920e0c/apex/optimizers/fused_lamb.py#L120).
```suggestion
        self.use_nvlamb = use_nvlamb
        
        # moving `lr` and `step` tensors to device as
        device = self.param_groups[0]["params"][0].device
        for group in self.param_groups:
            group["lr"] = group["lr"].to(device)
            group["step"] = group["step"].to(device)
```

crcrpar (commented):
I forgot that the argument of `argument` can be a `dict` so I think it's safer to move `lr` and `step` tensors to device after dunder init.

apex/optimizers/fused_mixed_precision_lamb.py
kevinstephano: I just realized you put up this suggestion.  I added a similar one.

crcrpar (approved):

PR #1247, opened 2021-12-11
Check size of minibatch in `get_kth_microbatch`

eqy:
We've seen many bugs due to `get_kth_microbatch` stepping past the number of elements in a minibatch silently. This PR changes the behavior to fire an assert when this happens and fixes a few bugs in tests uncovered by this change.

The batch size increase in the dynamic batch size is to support machines with 16 processes (GPUs), as there seems to be a limitation when the number of pipeline stages (16 GPU * 2 virtual) reaches the number of items in a microbatch (32).

CC @crcrpar @ptrblck

### DIFF
--- a/apex/transformer/pipeline_parallel/utils.py
+++ b/apex/transformer/pipeline_parallel/utils.py
@@ -126,7 +126,15 @@ def get_kth_microbatch(batch: List[torch.Tensor], k: int) -> List[torch.Tensor]:
     `a local minibatch` consists of `global_batch_size / data_parallel_size` samples.
     """
     micro_batch_size = get_micro_batch_size()
-    return [x[k * micro_batch_size:(k + 1) * micro_batch_size] for x in batch]
+    start = k * micro_batch_size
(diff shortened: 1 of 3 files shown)


apex/transformer/pipeline_parallel/utils.py
```diff
@@ -126,7 +126,15 @@ def get_kth_microbatch(batch: List[torch.Tensor], k: int) -> List[torch.Tensor]:
     `a local minibatch` consists of `global_batch_size / data_parallel_size` samples.
     """
     micro_batch_size = get_micro_batch_size()
-    return [x[k * micro_batch_size:(k + 1) * micro_batch_size] for x in batch]
+    start = k * micro_batch_size
+    end = start + micro_batch_size
+    microbatch = list()
```
crcrpar, line 131: nit-picking: better to use literal to initialize a list, instead of calling built-in function

crcrpar (approved):

PR #1249, opened 2021-12-14
Add an argument of `dtype` to forward_backward functions to specify the dtype used in p2p comm

crcrpar:
let users specify dtype for p2p comm taking the possibility of O2 style AMP into account

cc @eqy @ptrblck

### DIFF
--- a/tests/L0/run_transformer/run_pipeline_parallel_test.py
+++ b/tests/L0/run_transformer/run_pipeline_parallel_test.py
@@ -1,5 +1,3 @@
-from typing import Optional
-
 import torch
 
 from apex.transformer import parallel_state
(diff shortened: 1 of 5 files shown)


tests/L0/run_transformer/run_pipeline_parallel_test.py
```diff
@@ -91,8 +92,10 @@ def forward_backward_func_template(
     tensor_shape[0] = micro_batch_size
 
     update_num_microbatches(0)
-    forward_backward_func(
-        fwd_step_func, batch, model, forward_only=forward_only, tensor_shape=tensor_shape)
+    dtype = torch.half if enable_autocast else None
+    with torch.cuda.amp.autocast():
+        forward_backward_func(
+            fwd_step_func, batch, model, forward_only=forward_only, tensor_shape=tensor_shape, dtype=dtype)
```
crcrpar, line 98: We can use torch.bfloat16 here as well but autocast of bf16 requires ampere arch.
Notice that `forward_backward_func` takes the keyword only argument of "dtype".

crcrpar:
Thank you for double check, @eqy

PR #1250, opened 2021-12-14
Reduce OOM potential and report it if it happens in BERT test

eqy:
It appears running all L0 tests via `python tests/L0/run_tests.py`... can result in different amounts of GPU memory being available when the minimal BERT test launches. On certain configurations, such as 16GiB V100, this can cause the minimal BERT test to OOM and hang as one or more pipeline stages becomes unresponsive. This generally seems to occur within the first few forward-backward passes (as memory usage is still growing).

This PR reduces the memory footprint of the minimal BERT via smaller global batch sizes and reduced model size, and changes the failure reporting for the BERT test to show when a pipeline stage OOMs. Also includes a minor change to the minimal GPT test to use placeholder data.

CC @crcrpar @ptrblck

### DIFF
--- a/tests/L0/run_transformer/run_bert_minimal_test.py
+++ b/tests/L0/run_transformer/run_bert_minimal_test.py
@@ -16,6 +16,10 @@ from apex.transformer.testing.commons import TEST_SUCCESS_MESSAGE
 from apex.transformer.testing.commons import initialize_distributed
 from apex.transformer.testing.commons import print_separator
 
+import warnings
+class DebugWarning(Warning):
+    pass
+
 mode = None
 MANUAL_SEED = 42
 inds = None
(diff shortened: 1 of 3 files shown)


tests/L0/run_transformer/run_bert_minimal_test.py
```diff
@@ -135,13 +138,13 @@ def train(model, optim, virtual_pipeline_model_parallel_size, pipeline_model_par
             args.rampup_batch_size,
             args.global_batch_size,
             args.micro_batch_size,
-            1,  # args.data_parallel_size,
+            args.data_parallel_size,
         )
+        # virtual_pipeline_model_parallel_size = None
```
crcrpar, line 143:
```suggestion
```

tests/L0/run_transformer/run_bert_minimal_test.py
```diff
@@ -105,7 +108,7 @@ def train(model, optim, virtual_pipeline_model_parallel_size, pipeline_model_par
     hidden_size = global_vars.get_args().hidden_size
     forward_backward_func = get_forward_backward_func(virtual_pipeline_model_parallel_size, pipeline_model_parallel_size)
     tensor_shape = (args.seq_length, args.micro_batch_size, args.hidden_size)
-    for _ in range(8):
+    for i in range(16):
```
crcrpar, line 111: Is `i` used anywhere?

```suggestion
    for _ in range(16):
```

crcrpar (approved):

PR #1264, opened 2022-01-09
fix graph capture failure, fix norm computation with full_ar 

seryilmaz:
Removes an unnecessary wait_stream() call which causes graph capture failure due to uncaptured stream.
Fixes norm computation when  full_ar and clip_after_ar are enabled.


crcrpar:
let me close this in favor of 1266

PR #1265, opened 2022-01-09
add inline asm 128-bit counter

seryilmaz:
Adds inline PTX code for faster dropout fusion: the speed of fused FMHA kernel improves from 75 us to 69 us for num_seq=3, seq_length=512 (BERT). Changes are:
-- Inline PTX for 128 bit counter
-- Inline PTX for 32-bit integer multiply with 64-bit output


crcrpar (approved):
`pytest apex/contrib/test/multihead_attn/` was green.

PR #1266, opened 2022-01-10
fix group range to compute l2_norm

chochowski:
Use only subset of gpus to reduce the l2 norm instead of all


apex/contrib/optimizers/distributed_fused_lamb.py
```diff
@@ -683,6 +687,7 @@ def _pipeline_block_reductions(self, block_id):
                     # Since the packed format is contiguous after reductions, only one norm is needed
                     l2_grad_norm_sq = torch.empty([1], device='cuda')
                     if 0:#self._full_ar:
+                        #TODO
```
crcrpar, line 690: Is this PR ready for merge?

chochowski: I knew Burc was working on this case (line 690), so the #1264 addresses the missing if branch

crcrpar (approved):

PR #1267, opened 2022-01-11
fix: support new path to CUDAGeneratorImpl.h

shintaro-iwasaki:
OSS PyTorch moved `ATen/CUDAGeneratorImpl.h` to `ATen/cuda/CUDAGeneratorImpl.h` in https://github.com/pytorch/pytorch/pull/70650. This patch makes apex use the new path if a newer PyTorch package is used.

(This is my first time to create a PR for this repo. Please teach me any rule to send a patch)

### DIFF
--- a/apex/contrib/csrc/fmha/src/fmha.h
+++ b/apex/contrib/csrc/fmha/src/fmha.h
@@ -30,7 +30,12 @@
 #include <cuda.h>
 #include <vector>
 
+#ifdef OLD_GENERATOR_PATH
 #include <ATen/CUDAGeneratorImpl.h>
+#else
+#include <ATen/cuda/CUDAGeneratorImpl.h>
+#endif
+
 #include <ATen/cuda/CUDAGraphsUtils.cuh>
 
 #include <fmha_utils.h>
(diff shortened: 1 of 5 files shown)


crcrpar:
thanks for the PR but could you apply the same change to some more files which include CUDAGeneratorImpl.h (see: https://github.com/NVIDIA/apex/search?q=CUDAGeneratorImpl)?

shintaro-iwasaki:
Thanks, @crcrpar. I overlooked those.

Can I just rewrite the header path to `ATen/cuda/`? Or should I use a macro like `OLD_GENERATOR` to support both old and new PyTorch?

crcrpar:
I like the current approach, could you please keep the current approach for apex to be somewhat backward compatible?

shintaro-iwasaki:
@crcrpar: Thanks! I updated the commit by following the current approach.

Since `ATen/CUDAGenerator.h` is no longer used (they were changed 2 years ago and, for example, `apex/contrib/csrc/multihead_attn/softmax.cuh` directly uses `ATen/CUDAGeneratorImpl.h`), I removed the corresponding code in `apex/contrib/csrc/multihead_attn/dropout.cuh`.

crcrpar (approved):
Much appreciated.

PR #1279, opened 2022-01-24
T5 pipeline parallel changes

crcrpar:
- new: dealloc output.data and call C++ backward engine directly to skip output.shape == grad.shape check
- new: position embedding group


crcrpar (commented):
Tests with 8 gpus is green.

apex/transformer/pipeline_parallel/schedules/fwd_bwd_pipelining_without_interleaving.py
```diff
@@ -287,12 +290,20 @@ def forward_backward_pipelining_without_interleaving(
...
             output_tensor = output_tensors.pop(0)
 
-            input_tensor_grad = backward_step(input_tensor, output_tensor, output_tensor_grad, model_type=model_type, grad_scaler=grad_scaler)
+            input_tensor_grad = backward_step(
+                input_tensor,
+                output_tensor,
+                output_tensor_grad,
+                model_type=model_type,
+                grad_scaler=grad_scaler,
+                deallocate_pipeline_outputs=deallocate_pipeline_outputs,
```
eqy, line 305: Is the deallocation only compatible with the non-interleaved schedule?

eqy (approved):

apex/transformer/pipeline_parallel/schedules/fwd_bwd_pipelining_without_interleaving.py
crcrpar: yes, it is.

PR #1281, opened 2022-01-25
transformer: Allows for custom sync context in no pipelining forward backward function

crcrpar:
cc: @erhoo82 @ericharper @eqy @ptrblck


crcrpar (commented):
tests/L0/run_test.py looks okay

PR #1295, opened 2022-02-11
[FusedRMSNorm doc] document where epsilon is added

stas00:
This PR adds the the missing epsilon to formula to match the code

------------

It's actually very unlikely that the denominator will ever be zero in this class - unless each input is zero, since this is the root of sum of squares of numbers.

I don't know if it's faster to add epsilon than to check if all inputs are zero, since the output will be zero then.

@eqy

### DIFF
--- a/apex/normalization/fused_layer_norm.py
+++ b/apex/normalization/fused_layer_norm.py
@@ -310,7 +310,7 @@ class FusedRMSNorm(torch.nn.Module):
     :attr:`normalized_shape`.
     :math:`\gamma` is a learnable affine transform parameter of
     :attr:`normalized_shape` if :attr:`elementwise_affine` is ``True``.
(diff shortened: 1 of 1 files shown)


eqy:
I'm not sure this formula is accurate, as the epsilon value is used during the computation of the root-mean-square: https://github.com/NVIDIA/apex/blob/e1aa1fc1316a84e66869666270941265ec9cde77/csrc/layer_norm_cuda_kernel.cu#L378 rather than after.

stas00:
Thank you for catching that, @eqy. You're correct - perhaps we just document this nuance as in my last edit?

eqy (approved):

crcrpar:
thank you @stas00

crcrpar (approved):

PR #1309, opened 2022-02-25
[WIP] [DO NOT MERGE] NVFuser JIT eager mode InstanceNorm3d

eqy:
Prototype version with "cuDNN conv style" caching.


csrc/instance_norm_nvfuser_kernel.cu
```diff
@@ -0,0 +1,269 @@
...
+  // Params must be a POD because we read out its memory
+  // contenst as char* when comparing
+  static_assert(std::is_pod<Params>::value, "Params is not POD");
+
+  bool operator()(const Params& a, const Params& b) const {
+    auto ptr1 = reinterpret_cast<const uint8_t*>(&a);
+    auto ptr2 = reinterpret_cast<const uint8_t*>(&b);
+    return memcmp(ptr1, ptr2, sizeof(Params)) == 0;
+  }
+};
```
crcrpar, line 41: Isn't [ATen/native/utils/ParamsHash.h](https://github.com/pytorch/pytorch/blob/29d759948ee7cd2275ea60ebd62c311408e3798b/aten/src/ATen/native/utils/ParamsHash.h) available?

csrc/instance_norm_nvfuser_kernel.cu
```diff
@@ -0,0 +1,269 @@
...
+
+using namespace torch::jit::fuser::cuda;
+using namespace at::indexing;
+
+// Make a tensor that is known to be fully contiguous of dimensionality=ndims,
+// but unknown sizes
+TensorView* makeContigTensor(size_t ndims, DataType dtype = DataType::Float) {
```
crcrpar, line 48: [memo for me] comes from https://github.com/pytorch/pytorch/blob/127c9402d0ae7e3f72aa8ec18dbb9b8f12089bae/benchmarks/cpp/nvfuser/utils.h#L37

setup.py
```diff
@@ -7,6 +7,8 @@
 import warnings
 import os
 
+PYTORCH_HOME = os.path.abspath(os.environ['PYTORCH_HOME']) if 'PYTORCH_HOME' in os.environ else None
```
crcrpar, line 10: IMHO, making sure this `PYTORCH_HOME` has the two headers below that are included in the .cu file:
torch/csrc/jit/codegen/cuda/kernel_cache.h, and torch/csrc/jit/codegen/cuda/ops/all_ops.h?

what do you think?

tests/L0/run_instance_norm_nvfuser/test_instance_norm_nvfuser.py
```diff
@@ -0,0 +1,69 @@
...
+class TestInstanceNormNVFuser(unittest.TestCase):
+    dtype = torch.float
+    track_running_stats = False
+    channels_last = False
+    affine = False
+    batch_size = 5
+    channel_size = 7
+    spatial_size = 3
+
+    def setUp(self):
```
crcrpar, line 18: [nit] I think it's better to rename this method given how this is called.
`setUp`/`tearDown` is automatically called before/after each method starting with `test`, see [docs](https://docs.python.org/3/library/unittest.html#unittest.TestCase.setUp) for the details.

tests/L0/run_instance_norm_nvfuser/test_instance_norm_nvfuser.py
```diff
@@ -0,0 +1,69 @@
...
+                else:
+                    torch.testing.assert_close(self.m.weight.grad, self.reference_m.weight.grad)
+            if self.m.bias is not None:
+                if self.dtype == torch.float16:
+                    torch.testing.assert_close(self.m.bias.grad, self.reference_m.bias.grad, atol=5e-3, rtol=5e-3)
+                else:
+                    torch.testing.assert_close(self.m.bias.grad, self.reference_m.bias.grad)
+
+    def test_sweep(self):
+        for dtype, track_running_stats, channels_last, affine in itertools.product((torch.float, torch.half), (False, True), (False, True), (False, True)):
```
crcrpar, line 63: could you tell me why this test skips `torch.bfloat16` while the kernel seems to support it https://github.com/NVIDIA/apex/blob/5fa9b1e59b6fbaeabd5e4c5b592df1b6b52ae215/csrc/instance_norm_nvfuser_kernel.cu#L128?

csrc/instance_norm_nvfuser_kernel.cu
```diff
@@ -0,0 +1,269 @@
...
+  at::Tensor bias,
+  at::Tensor run_mean,
+  at::Tensor run_var,
+  const bool use_input_stats,
+  const float momentum,
+  const float eps,
+  const bool channels_last) {
+    InstanceNormKey forward_key;
+    memset(&forward_key, 0, sizeof(InstanceNormKey));
+    getKey(input, weight, run_mean, channels_last, eps, forward_key);
```
ngimel, line 108: why is `eps` part of the key, but `momentum` isn't? Also, can they be passed as args and not constants, so that the kernel doesn't have to be recompiled?

csrc/instance_norm_nvfuser_kernel.cu
```diff
@@ -0,0 +1,269 @@
...
+  at::Tensor weight,
+  at::Tensor bias,
+  at::Tensor run_mean,
+  at::Tensor run_var,
+  const bool use_input_stats,
+  const float momentum,
+  const float eps,
+  const bool channels_last) {
+    InstanceNormKey forward_key;
+    memset(&forward_key, 0, sizeof(InstanceNormKey));
```
ngimel, line 107: why are there 2 memsets?

csrc/instance_norm_nvfuser_kernel.cu
ngimel, line 108: as we've discussed, this is more like `setKey` than `getKey`

ngimel (commented):
This looks reasonable (but of course there'll need to be a few changes when moving upstream) 
What is the latency of compiling the kernel for the first time?
What is the latency if the kernel is found in cache?

ngimel (commented):
Also, can there be multiple kernels per fusion key or, since all the inputs are forced to contiguous, there's a single kernel only, and tuning is runtime at the level of thread/block sizes?

setup.py
eqy: +1 Sounds like a reasonable check before attempting to build

csrc/instance_norm_nvfuser_kernel.cu
eqy: Thanks, originally was having some issues including this header when prototyping, but seems to build fine now...

tests/L0/run_instance_norm_nvfuser/test_instance_norm_nvfuser.py
eqy: Not sure if we need to test/evaluate bf16 by default as it would remove Volta/`sm_70` support? Which I believe was the original use-case/target for instance-norm.

eqy: I'll look into testing this via https://pytorch.org/docs/stable/generated/torch.cuda.get_device_capability.html

apex/normalization/instance_norm.py
```diff
@@ -0,0 +1,149 @@
...
+        if instance_norm_nvfuser_cuda is None:
+            instance_norm_nvfuser_cuda = importlib.import_module("instance_norm_nvfuser_cuda")
+
+        channels_last = input.is_contiguous(memory_format=torch.channels_last) or input.is_contiguous(memory_format=torch.channels_last_3d)
+        if channels_last:
+            order = [0] + [i for i in range(2, len(input.shape))] + [1]
+            _input = input.permute(order)
+        else:
+            _input = input
+        assert _input.is_contiguous()
```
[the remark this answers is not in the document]
eqy: Right, some more checks need to be added, for now contiguous is assumed to simplify the support matrix, though non-contiguous can be supported by tweaking the caching for kernels and potentially more recompilation.

csrc/instance_norm_nvfuser_kernel.cu
eqy: good catch, thanks!

eqy:
> Also, can there be multiple kernels per fusion key or, since all the inputs are forced to contiguous, there's a single kernel only, and tuning is runtime at the level of thread/block sizes?

My understanding is that there would only be one compilation per fusion key; the restriction on contiguous is for initial testing purposes. The kernel is supposed to be general for multiple shapes, but the fuser folks would know more on launch bounds tuning.

tests/L0/run_instance_norm_nvfuser/test_instance_norm_nvfuser.py
ngimel: Hm, currently eager mode instance_norm supports bfloat16 even on pre-ampere GPUs, so with nvfuser we'd lose that support? We could fall back to old implementation then, I guess.

eqy: Could just be an artifact of how I'm prototyping things: naively generating bf16 casts and expecting it to work, haven't confirmed that it would be caused by nvfuser.

ngimel:
> My understanding is that there would only be one compilation per fusion key;

I don't think that's true, at the very least there could be generated kernels with int32 indexing for smaller tensors, and int64 indexing for larger ones, there could be other situations where different parallelization strategy will be chosen depending on the sizes?

eqy:
> 

Sorry, I mean "compilation step," or rather that if something is in the cache, it would be unexpected for running the fusion to trigger another compilation. But in general I also agree that more data needs to be collected for the latency of first execution and subsequent executions.

ngimel:
>Sorry, I mean "compilation step," or rather that if something is in the cache, it would be unexpected for running the fusion to trigger another compilation. But in general I also agree that more data needs to be collected for the latency of first execution and subsequent executions.

I've checked the code, and it is definitely possible to trigger another compilation even if something is in the cache, for example, for differently aligned input, or for some very different input sizes (I don't know which exact attributes of the tensors, other than alignment, will trigger recompile), or for large inputs (where int64 indexing has to be used). Which is fine, and the existing cache still saves some computation, but it should be documented that even hitting function level cache doesn't guarantee fast turnaround. 
It would be good to understand which parameter ranges will result in kernelRuntime match  in runFusionWithInputs.
And we still need the measurements for cache miss and full cache hit (function level cache is hit, and no recompilation is triggered in runFusionWithInputs)

csrc/instance_norm_nvfuser_kernel.cu
csarofeen: Neither should need to be part of the key.

csarofeen:
You're correct @ngimel. We can end up generating many kernels, if we want to limit the number of kernels we generate we would need implement coarser grained heuristics (definitely possible to do but we haven't done it yet). Cache misses will be dominated by nvrtc compile time. Cache hits are also not really easy to quantify as our finest grained cache is on input sizes, then there'd be a cache for heuristics, then this top level cache. The lowest level miss will not force recompilation so will be cheaper, but the higher two will require recompilation and will be dominated by nvrtc compile time. I'll work with Eddie to measure the hits at the lower two levels.

ngimel:
Awesome, thanks! So as far as I understand, lowest level will adjust launch parameters, maybe dynamic shared memory size and things like that? For my education, can you roughly describe what'll trigger recompilation? E.g. I compiled for a tensor of size (B,C,H,W), then how different (B1,C1,H1,W1) should be to trigger it? Or if you have documentation somewhere that I can look up?

csarofeen:
The highest level that Eddie was working on, or the heuristics? Heuristic recompilations are all dependent on heuristic changes and subject to change from one release to the next.

Practically for InstanceNorm 3D the big switches are:
Persistent vs non-persistent (we're working on grid persistence now)
Alignment on vectorization size (1, 2, 4, 8)
If not vectorized same factors for unrolling (depends on what fits well)
Grid reduction vs non-grid reduction
For Persistent cases these decisions are returned by: https://github.com/csarofeen/pytorch/blob/devel/torch/csrc/jit/codegen/cuda/scheduler/normalization.h#L21
which returns this reduction heuristic structure, if that structure is returned unchanged by this fun mega "==" check: https://github.com/csarofeen/pytorch/blob/devel/torch/csrc/jit/codegen/cuda/scheduler/reduction_heuristic.h#L111-L140
then no recompilation has to happen.

The function to determine these parameters is quite complex and "predictability" (obviously we know what will happen in an instance because code) may be relatively low as things can easily change as we continue to tune the heuristics.

If this is unacceptable for eager mode (I can imagine it may be as we don't want to recompile too much), we could generate simpler/flatter heuristics in place of the current ones which is more oriented to maximum performance at compilation cost. The thing we don't understand is how quickly these heuristics would converge, in practice on something like large transformers the answer is really quick even with dynamic sequence sizes. If we had something instead like a 3D segmentation model where inputs could change in all directions frequently, that's a lot more degrees of freedom and something like this approach may not be applicable. I don't know how we could easily solve the latter problem without impacting the former case (lower performance but less recompilation).

The one good news about this, is the heuristic structure is what determines recompilation, i.e. if we reimplemented https://github.com/csarofeen/pytorch/blob/devel/torch/csrc/jit/codegen/cuda/scheduler/normalization.cpp#L35-L492 to be coarser, recompilation will be less frequent, so we do have have control over this as this was required to not have to constantly change lots of code as we update heuristics.

ngimel:
Thanks, that's very helpful!

eqy:
Some ballpark numbers on the basic compile-cache-reuse workflow on the small workload in the tests; first execution is around 0.5-0.9s on V100, with about 30-60us of that being the actual kernel execution time and the bulk of it being the compilation. After the first execution the cache lookup takes around 0.9-1.5us.

ngimel:
0.9-1.5us is great

crcrpar:
I want to have pytorch ship `ATen/native/utils/*.h`, `torch/csrc/jit/codegen/cuda/*.h`, and `torch/csrc/jit/codegen/cuda/ops/*.h` before merging this.  
For this, what I guess we need to do is to update
- https://github.com/pytorch/pytorch/blob/543eaac415bfba59e648a3da8be5c4f964f9bc6e/setup.py#L938
- https://github.com/pytorch/pytorch/blob/543eaac415bfba59e648a3da8be5c4f964f9bc6e/aten/src/ATen/CMakeLists.txt#L149
- some file e.g. https://github.com/pytorch/pytorch/blob/master/torch/CMakeLists.txt and/or https://github.com/pytorch/pytorch/blob/master/torch/csrc/jit/codegen/cuda/nvfuser.cmake

crcrpar:
Getting closer to remove `PYTORCH_HOME` env var dependency. https://github.com/pytorch/pytorch/pull/78281

setup.py
```diff
@@ -7,6 +7,8 @@
...
+
```
crcrpar, line 11:
```suggestion
```

setup.py
```diff
@@ -335,6 +337,16 @@ def append_nvcc_threads(nvcc_extra_args):
             )
         )
 
+if PYTORCH_HOME is not None and os.path.exists(PYTORCH_HOME):
+    print(PYTORCH_HOME)
```
crcrpar, line 341: If we're to treat this as a `--cuda_ext`...
```suggestion
```

tests/L0/run_instance_norm_nvfuser/test_instance_norm_nvfuser.py
```diff
@@ -0,0 +1,77 @@
...
+        dtypes = [torch.float, torch.half]
+        if torch.cuda.get_device_capability() >= (8, 0):
+            dtypes.append(torch.bfloat16)
+        for dtype, track_running_stats, channels_last, affine in itertools.product(dtypes, (False, True), (False, True), (False, True)):
+            self.dtype = dtype
+            self.track_running_stats = track_running_stats
+            self.channels_last = channels_last
+            self.affine = affine
+            self.init_modules()
+            self.check_same_output()
```
crcrpar, line 77: To make failure messages more informative...
```suggestion
        for dtype, track_running_stats, channels_last, affine in itertools.product(dtypes, (False, True), (False, True), (False, True)):
            with self.subTest(dtype=dtype, track_running_stats=track_running_stats, channels_last=channels_last, affine=affine):
                self.dtype = dtype
                self.track_running_stats = track_running_stats
                self.channels_last = channels_last
                self.affine = affine
                self.init_modules()
                self.check_same_output() 
```

see: https://docs.python.org/3/library/unittest.html#unittest.TestCase.subTest

csrc/instance_norm_nvfuser_kernel.cu
```diff
@@ -0,0 +1,279 @@
...
+                                           _grad_output,
+                                           _weight,
+	  				                       _running_mean,
+                                           _running_var,
+	  				                       _save_mean,
+                                           _save_invstd,
+                                           use_input_stats,
+                                           _eps,
+                                           {true, true, true}, // TODO: is output mask useful?
+                                           channels_last);
```
crcrpar, line 252: [nit-picking] could you lint these lines as it seems like you've interchangeably used spaces and tabs...

csrc/instance_norm_nvfuser_kernel.cu
```diff
@@ -0,0 +1,279 @@
...
+#include <map>
+#include <vector>
+#include <chrono>
+
+#include <torch/extension.h>
+
+#include <torch/csrc/jit/codegen/cuda/kernel_cache.h>
+#include <torch/csrc/jit/codegen/cuda/ops/all_ops.h>
+
+#include <aten/src/ATen/native/utils/ParamsHash.h>
```
crcrpar, line 11:
```suggestion
#include <ATen/native/utils/ParamsHash.h>
```

csrc/instance_norm_nvfuser_kernel.cu
```diff
@@ -0,0 +1,279 @@
...
+  at::Tensor input,
+  at::Tensor grad_output,
+  at::Tensor weight,
+  at::Tensor run_mean,
+  at::Tensor run_var,
+  at::Tensor save_mean,
+  at::Tensor save_invstd,
+  const bool use_input_stats,
+  const float eps,
+  // const std::vector<bool>& output_mask,
```
crcrpar, line 184: Is this lack of `output_mask` worth mentioning in `instance_norm.py`?

crcrpar (commented):
final nit-pickings expect to be accessed once dependent headers get shipped...

setup.py
```diff
@@ -335,6 +337,16 @@ def append_nvcc_threads(nvcc_extra_args):
...
+    ext_modules.append(
+       CUDAExtension('instance_norm_nvfuser_cuda',
+                     ['csrc/instance_norm_nvfuser.cpp', 'csrc/instance_norm_nvfuser_kernel.cu'],
+                     extra_compile_args={"cxx": ["-O3"] + version_dependent_macros,
+                                         "nvcc": append_nvcc_threads(["-O3"] + version_dependent_macros + [f"-I {PYTORCH_HOME}"])},
```
crcrpar, line 346:
```suggestion
                                         "nvcc": append_nvcc_threads(["-O3"] + version_dependent_macros)},
```

tests/L0/run_instance_norm_nvfuser/test_instance_norm_nvfuser.py
```diff
@@ -76,3 +79,24 @@ def test_sweep(self):
...
+                return x
+
+        device = torch.device(f"cuda:1")
+        model = Model().to(device)
+
+        x = torch.randn(2, 4, 128, 128, 128, device=device, requires_grad=True)
+        y = torch.randn(2, device=device)
+        pred = model(x)
+        loss = nn.functional.mse_loss(pred, y.float())
+        loss.backward()
```
crcrpar, line 102: could you add 

```python
if __name__ == '__main__':
    unittest.main()
```

here so that we can run this file via command line using unittest? (as is, pytest can run this test though)

tests/L0/run_instance_norm_nvfuser/test_instance_norm_nvfuser.py
```diff
@@ -76,3 +79,24 @@ def test_sweep(self):
                 self.affine = affine
                 self.init_modules()
                 self.check_same_output() 
+
+    @unittest.skipIf(torch.cuda.device_count() < 2, "more than 1 GPU required")
+    def test_multigpu(self):
+        class Model(nn.Module):
+            def __init__(self):
+                super(Model, self).__init__()
```
crcrpar, line 87: super nit-picking to super() to use python3 syntax
```suggestion
                super().__init__()
```

crcrpar (commented):
sgtm

PR #1322, opened 2022-03-10
Mchochowski/zero fmha tensors

chochowski:
This is to change the initialization of tensors in fmha lib. As they were initialized with torch::empty there was no guarantee the memory will be zeroed. As long as we fill the tensor with computed values we don't care, but in case of fmha and ignoring padded elements, some spurios values may stay and result in NaNs later on. hypothesis: We didn't observe this before as for large scale BERT runs the number of iterations/allocations was small enough that the allocator did not reuse the memory used before. For small scale runs I think it already happened.


crcrpar (commented):
There's no race/memory errors but I needed changes like https://github.com/chochowski/apex/pull/1 to run the test.

crcrpar (approved):

PR #1340, opened 2022-03-23
Peer memory halo exchange

thorjohnsen:
This PR implements a peer memory pool that ranks can allocate tensors from.
Tensors allocated from the peer memory pool can be accessed from all ranks within a single node.
This PR also implements a push-pull 1d halo exchange kernel that relies on busy waiting on flags in peer memory to synchronize sender and receiver. This way of exchanging halos has very low latency.

### DIFF
--- a/apex/contrib/bottleneck/bottleneck_module_test.py
+++ b/apex/contrib/bottleneck/bottleneck_module_test.py
@@ -49,16 +49,29 @@ def rel_diff(x1, x2):
         return rel_diff_t(x1,x2)
 
 
-def fprop_and_bprop(x, bottleneck, dy=None):
+def graph_it(bottleneck, x):
+    print("Graphing")
     with torch.no_grad():
         x = x.clone()
         x.grad = None
         x.requires_grad = True
-    y = bottleneck(x)
-    if dy is None:
-        with torch.no_grad():
+    return torch.cuda.make_graphed_callables(bottleneck, (x,))
+
+
+def clone_inputs(bottleneck, x, dy=None):
+    with torch.no_grad():
+        x = x.clone()
+        x.grad = None
+        x.requires_grad = True
+        if dy is None:
+            y = bottleneck(x)
             dy = torch.randn_like(y) / 1e2
             torch.distributed.broadcast(dy, 0)
+    return x, dy
+
+
+def fprop_and_bprop(bottleneck, x, dy):
+    y = bottleneck(x)
     y.backward(dy)
     dgrad = x.grad.detach()
     wgrad = {}
@@ -74,7 +87,8 @@ def ground_truth(N, C, H, W, dtype, memory_format, bottleneck):
         with torch.no_grad():
             x = torch.randn([N,H,W,C], dtype=dtype, device='cuda')
             torch.distributed.broadcast(x, 0)
-        return fprop_and_bprop(x, bottleneck)
+            x, dy = clone_inputs(bottleneck, x)
+        return fprop_and_bprop(bottleneck, x, dy)
     else:
         # 2 -> native nhwc
         # 3 -> nchw
@@ -92,11 +106,9 @@ def print_ground_truth(gt):
 
 def apply_to_different_bottleneck(gt, bottleneck):
     with torch.no_grad():
-        x, y, dy, dgrad, wgrad = gt
-        x = x.clone()
-        x.requires_grad = True
-        dy = dy.clone()
-    return fprop_and_bprop(x, bottleneck, dy)
+        x, _, dy, _, _ = gt
+        x, dy = clone_inputs(bottleneck, x, dy)
+    return fprop_and_bprop(bottleneck, x, dy)
 
 
 def compare_single_field(results, f1, f2, l0, l1, l2):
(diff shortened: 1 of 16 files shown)


cliffwoolley:
@thorjohnsen Did you consider NVSHMEM for this?  Just curious.  /cc @nvcastet

apex/contrib/csrc/peer_memory/peer_memory_cuda.cu
```diff
@@ -0,0 +1,481 @@
...
+	    c = i % NC;
+	    w = i / NC;
+	    h = w / NW;
+	    w = w % NW;
+	}
+	else {
+	    w = i % NW;
+	    h = i / NW;
+	    c = h / NH;
+            h = h % NH;
```
cliffwoolley, line 143: Maybe consider using the int fastdiv trick here to avoid repeated integer divide/modulus by the same value?

thorjohnsen:
@cliffwoolley We had a prolonged discussion about whether to use NVSHMEM or not. It was initially thought that CUDNN libraries would use NVSHMEM internally. Code that uses NVSHMEM must be compiled with -rdc=True, which is something CUDNN team wasn't ready to do at the time, so we came up with this as an alternative. It's a moot point now, since we decided to postpone the CUDNN features that would have needed NVSHMEM until next round. Most of the code in this PR would have been the same with NVSHEM, for instance the push-pull halo exchange kernel would not change, only the few lines that exchange pointers via IPC handles would be different with NVSHEM.

apex/contrib/csrc/peer_memory/peer_memory_cuda.cu
thorjohnsen: I hadn't heard of the fastdiv trick, but it makes perfect sense. How do I use this in my code? Maxim Milakov has an implementation on github https://github.com/milakov/int_fastdiv. Can I just use his int_fastdiv.h header file?

thorjohnsen:
There are multiple ways to do halo exchanges in Pytorch. The easiest is NCCL all-gather, but all-gather is an expensive and redundant operation. An alternative is NCCL point-2-point communication, which is cheaper than all-gather algorithmically, but not necessarily cheaper in terms of runtime (all-gather is better optimized than more esoteric point-2-point comms). The third option is peer memory copy, which can be implemented either with NVSHMEM or via IPC handles. I provide all 3 implementations in this PR so we can test runtime which gives us best performance.

thorjohnsen:
Regarding NCCL point-2-point, torch.distributed has a send method and a receive method, but no send-and-receive method. This is a subtle but important distinction. NCCL can send and receive simultaneously, but AFAIK the send and receive calls have to run within a single NCCL group, which is what I implemented in nccl_p2p. The same exchange can be done with torch.distributed in the following manner:

if rank % 2 == 0:
  # even ranks send then receive
  ireq1 = torch.distributed.isend(t_out, rec_rank)
  ireq2 = torch.distributed.irecv(t_inp, src_rank)
else:
  # odd ranks receive then send
  ireq1 = torch.distributed.irecv(t_inp, src_rank)
  ireq2 = torch.distributed.isend(t_out, rec_rank)
ireq1.wait()
ireq2.wait()

If torch.distributed is smart enough to place the send and receive calls in a single NCCL group, nccl_p2p is superfluous, but I don't think it is.

apex/contrib/csrc/peer_memory/peer_memory_cuda.cu
```diff
@@ -0,0 +1,481 @@
...
+	}
+	size_t dst_off = c*dst_stride_C + h*dst_stride_H + w*dst_stride_W;
+	size_t src_off = c*src_stride_C + h*src_stride_H + w*src_stride_W;
+	dst[dst_off] = src[src_off];
```
nvcastet, line 147: Would 16byte store be  helpful here by casting to `float4*`?

apex/contrib/csrc/peer_memory/peer_memory_cuda.cu
```diff
@@ -0,0 +1,481 @@
...
+    // push top output halo to transfer buffer
+    strided_copy_kernel<T,is_HWC>(tox, tox_stride_C, tox_stride_H, tox_stride_W, toh, toh_stride_C, toh_stride_H, toh_stride_W, NC, NH, NW);
+    // push btm output halo to transfer buffer
+    strided_copy_kernel<T,is_HWC>(box, box_stride_C, box_stride_H, box_stride_W, boh, boh_stride_C, boh_stride_H, boh_stride_W, NC, NH, NW);
+    // signal to top and btm neigbhbors that output halos are ready to be read
+    // the choice of values for v1-v4 is arbitrary and does not matter, as long as all ranks use the same values
+    dual_signal_wait_clear<true,true>(signal1_flag, wait1_flag, signal2_flag, wait2_flag, -987751720, 840868300, -225529332, 281513358);
+    // pull top halo from transfer buffer in peer memory to input
+    strided_copy_kernel<T,is_HWC>(tih, tih_stride_C, tih_stride_H, tih_stride_W, tix, tix_stride_C, tix_stride_H, tix_stride_W, NC, NH, NW);
+    // pull btm halo from transfer buffer in peer memory to input
```
nvcastet, line 214: Have you tried pushing the halos instead of pulling them?

apex/contrib/peer_memory/peer_memory.py
```diff
@@ -0,0 +1,70 @@
...
+
+    def __init__(self, rank, world_size, peer_group_size, static_size, dynamic_size):
+        self.peer_group = rank // peer_group_size
+        self.peer_rank = rank % peer_group_size
+        self.peer_group_size = peer_group_size
+        self.alignment = 256
+        self.static_size = ((static_size + self.alignment - 1) // self.alignment) * self.alignment
+        self.dynamic_size = ((dynamic_size + self.alignment - 1) // self.alignment) * self.alignment
+        # allocate giant pool of device memory
+        self.raw = pm.allocate_raw(self.static_size+self.dynamic_size)
```
ptrblck, line 15: `allocate_raw` doesn't seem to be defined

apex/contrib/peer_memory/peer_memory.py
```diff
@@ -0,0 +1,70 @@
...
+        raw_ipc = pm.get_raw_ipc_address(self.raw).cuda()
+        peer_raw_ipcs = [torch.empty_like(raw_ipc) for _ in range(world_size)]
+        torch.distributed.all_gather(peer_raw_ipcs, raw_ipc)
+        peer_raw_ipcs = torch.stack(peer_raw_ipcs).cpu()
+        self.peer_raw = pm.get_raw_peers(peer_raw_ipcs, self.peer_rank, self.raw)
+        self.static_offset = 0
+        self.dynamic_offset = 0
+
+    def __del__(self):
+        pm.free_raw(self.raw)
```
ptrblck, line 26: `free_raw` also does't seem to be defined

thorjohnsen: The changes in this PR means that --fast_bottleneck now depends on two new extensions: --peer_memory and --nccl_p2p. I did not find a way to specify this dependency in setup.py. To compile --fast_bottleneck, the following switches are necessary: 

python setup.py install --cpp_ext --cuda_ext --fast_bottleneck --peer_memory --nccl_p2p

thorjohnsen: To be clear, --fast_bottleneck will compile without --peer_memory and --nccl_p2p, but it won't run correctly.

thorjohnsen:
Hi @ptrblck I have renamed the new extensions peer_memory and nccl_p2p to peer_memory_cuda and nccl_p2p_cuda. This fixes the issue you saw, the module tests now run fine from their original directories. Everything still compiles the same, the switches are still --peer_memory and --nccl_p2p, this just affects how the extensions are imported in python source.

apex/contrib/bottleneck/bottleneck_module_test.py
```diff
@@ -1,271 +1,232 @@
...
+    N, H, W, C = list(x.shape) # Tensor is already shaped properly for n-way parallel
+    dtype = x.dtype
+
+    spatial_group_size = world_size
+    spatial_group_rank = rank
+    spatial_communicator = None
+    spatial_halo_exchanger = halex
+    spatial_method = 3 # 1 -> overlap halo and main conv, 2 -> wait for halo, conv on padded x
+    spatial_parallel_args = (spatial_group_size, spatial_group_rank, spatial_communicator, spatial_halo_exchanger, spatial_method)
+    spatial_bottleneck = spatial_parallel_bottleneck(C, dtype, explicit_nhwc, gt_bottleneck, spatial_parallel_args)
```
ptrblck, line 166: `spatial_parallel_bottleneck` init seems to be missing an argument

thorjohnsen: Fixed.

crcrpar: `python -m torch.distributed.run --nproc_per_node 8 apex/contrib/bottleneck/bottleneck_module_test.py` ran successfully. thanks.

PR #1343, opened 2022-03-24
[transformer] `parallel_state`: Position Embedding

crcrpar:
### Additions
- Position Embedding global ranks
- Position Embedding groups
- :func:`get_position_embedding_group`
- :func:`is_rank_in_position_embedding_group`
- :func:`get_data_parallel_src_rank`

### Clean-up
- :func:`destroy_model_parallel`

### DIFF
--- a/apex/transformer/parallel_state.py
+++ b/apex/transformer/parallel_state.py
@@ -12,12 +12,22 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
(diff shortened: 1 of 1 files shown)


apex/transformer/parallel_state.py
```diff
@@ -451,15 +562,3 @@ def destroy_model_parallel():
     _EMBEDDING_GROUP = None
     global _POSITION_EMBEDDING_GROUP
     _POSITION_EMBEDDING_GROUP = None
-    global _VIRTUAL_PIPELINE_MODEL_PARALLEL_RANK
```
eqy, line 454: is there a change in a previous PR that enables these to be removed?

PR #1362, opened 2022-04-23
Add LANS Optimizer

szhengac:
As per the discussion offline, I am opening this PR for upstreaming [LANS optimizer](https://arxiv.org/abs/2006.13484) into Apex master. LANS improves over LAMB on large-batch training for Transformer based models. The implementation is built based on Apex LAMB code. We clocked 54 mins training of BERT-large model on 1536 V100 GPUs dated back in 2020. In addition to LANS, this PR also adds support for stochastic rounding for optimization step with bfloat16. A [recent study ](https://arxiv.org/abs/2010.06192) shows that we can use bfloat16 optimizer state if apply stochastic rounding to the optimization step. In our experiment, it still does not match with the accuracy with the float32 training, but it does improve over the vanilla training with bfloat16 optimizer state.

### DIFF
--- a/apex/optimizers/__init__.py
+++ b/apex/optimizers/__init__.py
@@ -4,3 +4,4 @@ from .fused_novograd import FusedNovoGrad
 from .fused_lamb import FusedLAMB
 from .fused_adagrad import FusedAdagrad
 from .fused_mixed_precision_lamb import FusedMixedPrecisionLamb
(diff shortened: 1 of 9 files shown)


ptrblck:
Thanks for the MR! Currently, we are not seeing enough use cases for this optimizer and will close the MR for now.
If this changes we will revisit and reopen this MR.

PR #1374, opened 2022-05-05
[WIP] [transformer] add loss comparison to test_pipeline_parallel_fwd_bwd

Aidyn-A:
Added a loss comparison to `test_pipeline_parallel_fwd_bwd.py`.
In order to effectively predict the target loss value, `apex/transformer/testing/commons.py` was also modified to prevent randomization of weights. 

EDIT: 
On second commit I applied some of the suggested changes:
- Reverted the changes in `commons.py`. Instead, the test calls `torch.nn.Module.apply`.
- Minor change in the assertion line.
- Added `get_target_loss` that calculates the loss assuming that `forward_backward_no_pipelining` is correct...

I kept `rank = torch.distributed.get_rank()` because ~~`parallel_state.get_rank_info()` returns something weird...~~ it is easier to debug.

EDIT 2:
- Target loss calculation function is now correct for any size of `pipeline_model_parallel_world_size`.
- Test changed for running on `float32` by default.

### DIFF
--- a/tests/L0/run_transformer/test_pipeline_parallel_fwd_bwd.py
+++ b/tests/L0/run_transformer/test_pipeline_parallel_fwd_bwd.py
@@ -34,7 +34,7 @@ weight_coeff = 1024
 @torch.no_grad()
 def init_weights(m):
     rank = torch.distributed.get_rank()
-    if type(m) == torch.nn.Linear:
+    if isinstance(m, torch.nn.Linear):
         m.weight.fill_((rank + 1.0) / weight_coeff)
         m.bias.fill_(1.0)
 
@@ -42,15 +42,15 @@ def init_weights(m):
 def get_target_loss(hidden_size: int, microbatch_size: int, parallel_model_world_size: int, world_size: int) -> float:  
     layers_per_rank = world_size // parallel_model_world_size
     data = torch.arange(start = 0, end = layers_per_rank, dtype = torch.int) + 1
-    
-    w = (torch.arange(world_size, dtype = torch.int) + 1) / weight_coeff
+
+    w = (torch.arange(world_size, dtype = torch.float) + 1.0) / weight_coeff
     b = torch.ones(world_size, dtype = torch.int)
     w = hidden_size * w
 
-    for pd in range(0, world_size, layers_per_rank):
-        eid = pd+layers_per_rank
-        data = w[pd:eid] * data + b[pd:eid]
-    
+    for s_id in range(0, world_size, layers_per_rank):
+        e_id = s_id+layers_per_rank
+        data = w[s_id:e_id] * data + b[s_id:e_id]
+
     return hidden_size * hidden_size * torch.sum(data).item() * microbatch_size / layers_per_rank
 
 
@@ -70,9 +70,10 @@ class PipelineParallelForwardBackwardTest(DistributedTestBase):
         fwd_bwd_func: FwdStepFunc,
         pipeline_model_parallel_world_size: Optional[int],
         virtual_pipeline_model_parallel_size: Optional[int],
-        dtype: Optional[torch.dtype] = torch.float32,
     ) -> None:
-        for deallocate_pipeline_outputs in (True, False):
+        for dtype, deallocate_pipeline_outputs in itertools.product(
+            [torch.float32] + _get_autocast_dtypes(), (True, False),
+        ):
             grad_scaler = (
                 torch.cuda.amp.GradScaler(init_scale=4.0)
                 if dtype == torch.half
@@ -139,13 +140,14 @@ class PipelineParallelForwardBackwardTest(DistributedTestBase):
                 deallocate_pipeline_output=deallocate_pipeline_outputs,
             )
 
-            hidden_size = PipelineParallelForwardBackwardTest.HIDDEN_SIZE
-            microbatch_size = PipelineParallelForwardBackwardTest.MICRO_BATCH_SIZE
-            target_loss = get_target_loss(hidden_size, microbatch_size, pipeline_model_parallel_world_size, self.world_size)
+            if dtype == torch.float32:
+                hidden_size = PipelineParallelForwardBackwardTest.HIDDEN_SIZE
+                microbatch_size = PipelineParallelForwardBackwardTest.MICRO_BATCH_SIZE
+                target_loss = get_target_loss(hidden_size, microbatch_size, pipeline_model_parallel_world_size, self.world_size)
 
-            for loss_item in loss:
-                x = loss_item['avg']
-                torch.testing.assert_close(x, target_loss*torch.ones_like(x))
+                for loss_item in loss:
+                    x = loss_item['avg']
+                    torch.testing.assert_close(x, target_loss*torch.ones_like(x))
 
             if not forward_only:
                 for m in model:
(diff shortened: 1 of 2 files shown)


tests/L0/run_transformer/test_pipeline_parallel_fwd_bwd.py
```diff
@@ -112,6 +113,17 @@ def _forward_backward_test_impl(
...
+            h_size = PipelineParallelForwardBackwardTest.HIDDEN_SIZE
+            m_size = PipelineParallelForwardBackwardTest.MICRO_BATCH_SIZE
+            target_loss = 0
+            for r in range(self.world_size):
+                target_loss = target_loss + h_size*(r + 1) + 1
+            target_loss = target_loss * (h_size ** 2) * m_size / self.world_size
+
+            for loss_item in loss:
+                x = loss_item['avg']
+                self.assertTrue(torch.allclose(x, target_loss*torch.ones_like(x)))
```
eqy, line 125: Currently I think the test runs for multiple iterations, would there need to be some logic to revert the parameters to their previous state due to the call to `optimizer.step()`?

tests/L0/run_transformer/test_pipeline_parallel_fwd_bwd.py
```diff
@@ -112,6 +113,17 @@ def _forward_backward_test_impl(
                 deallocate_pipeline_output=deallocate_pipeline_outputs,
             )
 
+            h_size = PipelineParallelForwardBackwardTest.HIDDEN_SIZE
+            m_size = PipelineParallelForwardBackwardTest.MICRO_BATCH_SIZE
+            target_loss = 0
+            for r in range(self.world_size):
+                target_loss = target_loss + h_size*(r + 1) + 1
```
eqy, line 120: I'm not sure about the arithmetic here, should there be a multiply in the loop if output of the previous stage is being multiplied with the current layer's weights? Alternatively the hidden layer weights could be set to zero.

tests/L0/run_transformer/test_pipeline_parallel_fwd_bwd.py
crcrpar, line 125: nit-picking

```suggestion
                torch.testing.assert_close(x, target_loss*torch.ones_like(x))
```

ref: https://pytorch.org/docs/stable/testing.html#torch.testing.assert_close

crcrpar: The reason of multiple iterations is that we once experienced a hang in the second or later iteration.

For this, an alternative is to store the initial `torch.nn.Module.state_dict` and load it before each forward/backward function call.

crcrpar (commented):
appreciate your swift work. 
As Eddie's covered things, I wrote some minor comments.

tests/L0/run_transformer/test_pipeline_parallel_fwd_bwd.py
```diff
@@ -129,22 +164,22 @@ def test_no_pipelining_inference(self):
 
     def test_pipelining(self):
         self._forward_backward_test_impl(
-            False, forward_backward_pipelining_without_interleaving, None, None
+            False, forward_backward_pipelining_without_interleaving, 8, None
```
eqy, line 167: Do we still need this specification of `8`? Might break in some test environments (e.g., 4 GPU CI machines).

tests/L0/run_transformer/test_pipeline_parallel_fwd_bwd.py
```diff
@@ -46,23 +70,21 @@ def _forward_backward_test_impl(
         fwd_bwd_func: FwdStepFunc,
         pipeline_model_parallel_world_size: Optional[int],
         virtual_pipeline_model_parallel_size: Optional[int],
+        dtype: Optional[torch.dtype] = torch.float32,
```
crcrpar, line 73: I'm afraid I don't think this argument will be used so I'd like to either have this test class stick to torch.float32 or leave the `product` of dtypes and "dealloc"
```suggestion
```

tests/L0/run_transformer/test_pipeline_parallel_fwd_bwd.py
```diff
@@ -46,23 +70,21 @@ def _forward_backward_test_impl(
...
     ) -> None:
-        for dtype, deallocate_pipeline_outputs in itertools.product(
-            [torch.float32] + _get_autocast_dtypes(), (True, False),
-        ):
+        for deallocate_pipeline_outputs in (True, False):
```
crcrpar, line 75: and IMHO leaving the second option is preferable because it shows me that the fwd-bwd function works with torch.float16 & torch.bfloat16. My major concerns with these two dtypes are (a) p2p communication using them and (b) `grad_scaler` for torch.half.

```suggestion
        for dtype, deallocate_pipeline_outputs in itertools.product(
            [torch.float32] + _get_autocast_dtypes(), (True, False),
        ):
```

tests/L0/run_transformer/test_pipeline_parallel_fwd_bwd.py
```diff
@@ -112,6 +139,14 @@ def _forward_backward_test_impl(
                 deallocate_pipeline_output=deallocate_pipeline_outputs,
             )
 
+            hidden_size = PipelineParallelForwardBackwardTest.HIDDEN_SIZE
+            microbatch_size = PipelineParallelForwardBackwardTest.MICRO_BATCH_SIZE
+            target_loss = get_target_loss(hidden_size, microbatch_size, pipeline_model_parallel_world_size, self.world_size)
+
+            for loss_item in loss:
+                x = loss_item['avg']
+                torch.testing.assert_close(x, target_loss*torch.ones_like(x))
```
crcrpar, line 148: and as dtype can be torch.bfloat16/torch.float16, I want to guard this nice numerical check
```suggestion
            if dtype == torch.float32:
                for loss_item in loss:
                    x = loss_item['avg']
                    torch.testing.assert_close(x, target_loss*torch.ones_like(x))
```

tests/L0/run_transformer/test_pipeline_parallel_fwd_bwd.py
```diff
@@ -29,11 +29,35 @@
 
 logging.getLogger("apex").setLevel(logging.WARNING)
 
+weight_coeff = 1024
```
crcrpar, line 32: [question] Is this a coeff to let the numerical check with torch.float16 as well as torch.float32?

tests/L0/run_transformer/test_pipeline_parallel_fwd_bwd.py
```diff
@@ -29,11 +29,35 @@
...
+
+@torch.no_grad()
+def init_weights(m):
+    rank = torch.distributed.get_rank()
+    if type(m) == torch.nn.Linear:
```
crcrpar, line 37: nit-picking
```suggestion
    if isinstance(m, torch.nn.Linear):
```

tests/L0/run_transformer/test_pipeline_parallel_fwd_bwd.py
Aidyn-A: Yes, it is for both float16 and float32. Although, in my tests the loss based on float16 only went out of numerical bounds and became `inf` or `nan`.

tests/L0/run_transformer/test_pipeline_parallel_fwd_bwd.py
Aidyn-A: With `None` it will run fine on any machine. However, on machines with 8 GPUs and more, there is a condition that I do not quite understand the purpose:

```
data_parallel_size = 1 + (self.world_size >= 8 and self.world_size % 2 == 0)
```

crcrpar: The purpose was to have multiple parallelism enabled in a test, which might not be that critical.

eqy (approved):

crcrpar (approved):

PR #1378, opened 2022-05-09
Add get_world_info()

Aidyn-A:
A method `get_world_info()` will be very useful for variety of purposes especially for testing and debugging.
Additionally, returning all zeros in `get_rank_info()` when parallel is not initialized can be misleading. 
I think it would be even better to have `assert model_parallel_is_initialized()` in future.

### DIFF
--- a/apex/transformer/parallel_state.py
+++ b/apex/transformer/parallel_state.py
@@ -268,7 +268,7 @@ def get_rank_info() -> Tuple[int, int, int]:
             get_pipeline_model_parallel_rank(),
             get_virtual_pipeline_model_parallel_rank(),
         )
-    return (0, 0, 0, 0)
+    return (-1, -1, -1, -1)
 
 
 def model_parallel_is_initialized():
--- a/apex/transformer/parallel_state.py
+++ b/apex/transformer/parallel_state.py
@@ -247,6 +247,18 @@ def initialize_model_parallel(
             _POSITION_EMBEDDING_GLOBAL_RANKS = position_embedding_ranks
 
 
+def get_world_info() -> Tuple[int, int, int]:
+    """Returns a tuple of (data, tensor, pipeline, virtual pipeline)-parallel-world_size for logger."""
+    if model_parallel_is_initialized():
+        return (
+            get_data_parallel_world_size(),
+            get_tensor_model_parallel_world_size(),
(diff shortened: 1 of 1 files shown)


crcrpar (commented):
IMHO a warning might be helpful when model parallel is NOT initialized.
What do you think?

cc @eqy

eqy:
I agree on the warning part, additionally we could perhaps format the output as a log and add it to debug logging in apex.

Aidyn-A:
> I agree on the warning part, additionally we could perhaps format the output as a log and add it to debug logging in apex.

I think the log is already present [here](https://github.com/NVIDIA/apex/blob/c3018b132c3dc180e5fda64e8ef432a1a22c7734/apex/transformer/parallel_state.py#L116-L129). I do not why it is not printing it though...

apex/transformer/parallel_state.py
```diff
@@ -247,7 +247,21 @@ def initialize_model_parallel(
...
+    """Returns a tuple of (data, tensor, pipeline, virtual pipeline)-parallel-world_size for logger."""
+    if model_parallel_is_initialized():
+        return (
+            get_data_parallel_world_size(),
+            get_tensor_model_parallel_world_size(),
+            get_pipeline_model_parallel_world_size(),
+            get_virtual_pipeline_model_parallel_world_size(),
+        )
+    else:
+        print("Warning: model and data parallel groups are not initialized")
```
crcrpar, line 260: sorry for having confused you but by a warning I meant something like

```suggestion
        warnings.warn("model and data parallel groups are not initialized. Have you called `parallel_state.initialize_model_parallel`?")
```
which needs `import warnings`.

With this we can test this warning with `unittest` using e.g. https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertWarnsRegex (needn't to access in this PR)

apex/transformer/parallel_state.py
```diff
@@ -256,7 +270,9 @@ def get_rank_info() -> Tuple[int, int, int]:
             get_pipeline_model_parallel_rank(),
             get_virtual_pipeline_model_parallel_rank(),
         )
-    return (0, 0, 0, 0)
+    else:
+        print("Warning: model and data parallel groups are not initialized")
```
crcrpar, line 274: ditto

Issue #1381, opened 2022-05-12
Disable TF32 for numerical check of test_pipeline_parallel_fwd_bwd.py


crcrpar:
Numerical check is added in https://github.com/NVIDIA/apex/commit/68440264f3fc564fa53b5a2f286d313774fcfe49.
It turned out that this check requires float32.

cc @Aidyn-A @eqy 

https://github.com/NVIDIA/apex/blob/68440264f3fc564fa53b5a2f286d313774fcfe49/tests/L0/run_transformer/test_pipeline_parallel_fwd_bwd.py#L143-L150

eqy:
I will check to see if this can be done with the `allow_tf32` flag that already exists.

PR #1382, opened 2022-05-13
[Pipeline-Parallelism][TF32] Disable TF32 for Pipeline-Parallel numerical checks

eqy:
The numerical checks in #1374 assume typical fp32 exact integer range which isn't supported with TF32. This isn't an issue on Volta and earlier but Ampere devices will dispatch to TF32 by default for the linear layers leading to unexpected mismatches.

See also: #1381

CC @crcrpar @Aidyn-A

### DIFF
--- a/tests/L0/run_transformer/test_pipeline_parallel_fwd_bwd.py
+++ b/tests/L0/run_transformer/test_pipeline_parallel_fwd_bwd.py
@@ -72,6 +72,8 @@ class PipelineParallelForwardBackwardTest(DistributedTestBase):
         virtual_pipeline_model_parallel_size: Optional[int],
         async_comm: bool = False,
     ) -> None:
+        orig = torch.backends.cuda.matmul.allow_tf32
+        torch.backends.cuda.matmul.allow_tf32 = False
         for dtype, deallocate_pipeline_outputs in itertools.product(
             [torch.float32] + _get_autocast_dtypes(), (True, False),
         ):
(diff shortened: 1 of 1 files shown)


crcrpar (approved):
Thanks, could it be simpler to use `tf32_off` decorator instead?
https://github.com/pytorch/pytorch/blob/c08b8f0967efc2eec078da4541c5fdd003fbdd75/torch/testing/_internal/common_cuda.py#L66

eqy:
Are we importing/able to use testing/_internal in `apex`? Otherwise it seems the contextmanager style is only available for cuDNN...
https://github.com/pytorch/pytorch/blob/8bb720304973f9a22562d9b314edfee4558076fc/torch/testing/_internal/common_cuda.py#L69

Aidyn-A:
> Are we importing/able to use testing/_internal in `apex`? Otherwise it seems the contextmanager style is only available for cuDNN... https://github.com/pytorch/pytorch/blob/8bb720304973f9a22562d9b314edfee4558076fc/torch/testing/_internal/common_cuda.py#L69

I just checked and confirmed that we can use `tf32_off` with and without cuDNN being enabled or installed.

crcrpar:
> Are we importing/able to use testing/_internal in `apex`?

I've already used some `torch/testing/_internal` in transformer module tests e.g.
- https://github.com/NVIDIA/apex/blob/3fe352118d2fe53dab734e0995a730ca5ffc9cc4/apex/transformer/testing/distributed_test_base.py#L5-L9
- https://github.com/NVIDIA/apex/blob/3fe352118d2fe53dab734e0995a730ca5ffc9cc4/tests/L0/run_transformer/test_pipeline_parallel_fwd_bwd.py#L6

PR #1384, opened 2022-05-16
Test `len(model) > 1` in `test_pipelining_with_interleaving`

eqy:
As noticed by @crcrpar, after the refactor in #1325 the interleaved test was not setting the virtual pipeline pipeline model parallel size which results in interleaving not being fully exercised.

This PR sets the virtual pipeline model parallel size and revises the numerical check to simplify the logic for various configurations as the number of layers per pipeline stage changes with interleaving.

CC @Aidyn-A

### DIFF
--- a/tests/L0/run_transformer/test_pipeline_parallel_fwd_bwd.py
+++ b/tests/L0/run_transformer/test_pipeline_parallel_fwd_bwd.py
@@ -2,6 +2,8 @@ import logging
 import itertools
 from typing import Optional
 
+import numpy as np
+
 import torch
 from torch.testing._internal import common_utils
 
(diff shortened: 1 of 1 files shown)


tests/L0/run_transformer/test_pipeline_parallel_fwd_bwd.py
```diff
@@ -108,7 +109,9 @@ def _forward_backward_test_impl(
                 PipelineParallelForwardBackwardTest.HIDDEN_SIZE,
             )
 
-            batch =(((self.rank + 1) * torch.ones(global_batch_shape)).cuda(), )
+            batch = None
+            if parallel_state.is_pipeline_first_stage:
+                batch = (torch.ones(global_batch_shape, dtype=torch.float).cuda(), )
```
Aidyn-A, line 114: This is all right if we are 100% certain that this will not cause any trouble if `recv` fails.

eqy: We would want that to raise an error in any case, it would not be expected behavior. In the current implementation, `batch` should not be used by the downstream stages anyway so this check just enforces this property.

crcrpar (approved):
lgtm

PR #1391, opened 2022-05-27
Do pipeline parallelism tests in double because TF32 environment variables can be painful to manage across test suites

eqy:
CC @Aidyn-A @crcrpar

### DIFF
--- a/tests/L0/run_transformer/test_pipeline_parallel_fwd_bwd.py
+++ b/tests/L0/run_transformer/test_pipeline_parallel_fwd_bwd.py
@@ -48,7 +48,7 @@ def get_init_weights_func(offset: int = 0):
     return init_weights
 
 
-def get_target_loss_and_model(global_batch_shape: tuple, hidden_size: int, total_layers: int) -> Tuple[float, List[torch.nn.Module]]: 
+def get_target_loss_and_model(global_batch_shape: tuple, hidden_size: int, total_layers: int) -> Tuple[torch.Tensor, List[torch.Tensor]]: 
     model = []
     data = torch.ones(global_batch_shape, dtype=torch.double)
     for i in range(total_layers):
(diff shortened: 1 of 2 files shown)


Aidyn-A:
Awesome! Looks good to me :+1:

tests/L0/run_transformer/test_pipeline_parallel_fwd_bwd.py
```diff
@@ -50,10 +50,10 @@ def init_weights(m):
 
 def get_target_loss_and_model(global_batch_shape: tuple, hidden_size: int, total_layers: int) -> float:
```
Aidyn-A, line 51: Sorry, this was my mistake, but could you please change the returning type

tests/L0/run_transformer/test_pipeline_parallel_fwd_bwd.py
```diff
@@ -48,12 +48,12 @@ def init_weights(m):
     return init_weights
 
 
-def get_target_loss_and_model(global_batch_shape: tuple, hidden_size: int, total_layers: int) -> float:  
+def get_target_loss_and_model(global_batch_shape: tuple, hidden_size: int, total_layers: int) -> Tuple[float, List[torch.nn.Module]]:
```
crcrpar, line 51: I guess `[w, b]` should be `List[torch.Tensor]`?
```suggestion
def get_target_loss_and_model(global_batch_shape: tuple, hidden_size: int, total_layers: int) -> Tuple[float, List[torch.Tensor]]: 
```

crcrpar (approved):

tests/L0/run_transformer/test_pipeline_parallel_fwd_bwd.py
Aidyn-A: I shall add that `loss` is a `torch.Tensor` of single element.

PR #1397, opened 2022-06-01
[transformer] Specify process group of `batch_isend_irecv`

crcrpar:
## Summary
- Specify process group in p2p_communication.py
- Add `torch.cuda.synchronize()` as per @Aidyn-A's [comment](https://github.com/NVIDIA/apex/pull/1397#discussion_r899466940)
- as of https://github.com/openucx/ucc/commit/8afd34a11ae961bffc29c44be3a08fb994f53c00, we cannot make true overlapping happen with torch_ucc backend even if we set `async_comm` to `True`.

## TODO
In a follow-up
- [ ] let `_default_pg` argument be a real argument and avoid a redundant `torch.cuda.synchronize()` if `_different_pg` is `False`.

## concerns
- `_different_pg` is not that good name in my humble opinion. The situation is rather Pipeline Model Parallel ProcessGroups depends on a different backend from the other ProcessGroups...
- Whether or not to have users specify `_different_pg`
    - As Aidyn's pointed out, we can tell by comparing Pipeline Model Parallel ProcessGroup and another Model Parallel ProcessGroup

`UCP_TLS=ucp`

---

## Motivation
https://github.com/NVIDIA/apex/blob/5ffb22d0d4ff2b899638b768af13bb1b97cbc923/tests/L0/run_transformer/test_pipeline_parallel_fwd_bwd.py#L262-L289 doesn't show appropriately UCX/UCC warnings as P2P ops implicitly use the default process group, i.e. NCCL Process Group

Tests seem fine with UCC of [a43fad17fa4f1e57a712b6bfb5abcf087d0bb95f](https://github.com/openucx/ucc/commit/a43fad17fa4f1e57a712b6bfb5abcf087d0bb95f) and torch_ucc of [a9451896f2592313322cca04ee0d4e0c5d5f922f](https://github.com/facebookresearch/torch_ucc/commit/a9451896f2592313322cca04ee0d4e0c5d5f922f).

### DIFF
--- a/apex/transformer/pipeline_parallel/p2p_communication.py
+++ b/apex/transformer/pipeline_parallel/p2p_communication.py
@@ -53,32 +53,37 @@ def _run_p2pops(
     async_comm: bool = False
 ):
     ops = []
+    group = parallel_state.get_pipeline_model_parallel_group()
     if tensor_send_prev is not None:
         send_prev_op = torch.distributed.P2POp(
-            torch.distributed.isend,
-            tensor_send_prev,
-            parallel_state.get_pipeline_model_parallel_prev_rank(),
+            op=torch.distributed.isend,
+            tensor=tensor_send_prev,
+            peer=parallel_state.get_pipeline_model_parallel_prev_rank(),
+            group=group,
         )
         ops.append(send_prev_op)
     if tensor_recv_prev is not None:
         recv_prev_op = torch.distributed.P2POp(
-            torch.distributed.irecv,
-            tensor_recv_prev,
-            parallel_state.get_pipeline_model_parallel_prev_rank(),
+            op=torch.distributed.irecv,
+            tensor=tensor_recv_prev,
+            peer=parallel_state.get_pipeline_model_parallel_prev_rank(),
+            group=group,
         )
         ops.append(recv_prev_op)
     if tensor_send_next is not None:
         send_next_op = torch.distributed.P2POp(
-            torch.distributed.isend,
-            tensor_send_next,
-            parallel_state.get_pipeline_model_parallel_next_rank(),
+            op=torch.distributed.isend,
+            tensor=tensor_send_next,
+            peer=parallel_state.get_pipeline_model_parallel_next_rank(),
+            group=group,
         )
         ops.append(send_next_op)
     if tensor_recv_next is not None:
         recv_next_op = torch.distributed.P2POp(
-            torch.distributed.irecv,
-            tensor_recv_next,
-            parallel_state.get_pipeline_model_parallel_next_rank(),
+            op=torch.distributed.irecv,
+            tensor=tensor_recv_next,
+            peer=parallel_state.get_pipeline_model_parallel_next_rank(),
+            group=group,
         )
         ops.append(recv_next_op)
     if len(ops) > 0:
--- a/apex/transformer/pipeline_parallel/p2p_communication.py
+++ b/apex/transformer/pipeline_parallel/p2p_communication.py
@@ -45,13 +45,44 @@ class FutureTensor:
         return self.tensor
 
 
+# NOTE(mkozuki): [Necessity of `torch.cuda.synchronize()`]
+# When multiple torch.distributed backends are used, e.g.
+# `torch_ucc` for P2P (= Pipeline Parallel) and `nccl` for the rest,
+# it's possible that a preceding NCCL `torch.distributed.all_reduce` call
+# does not finish before P2P collective using UCC, which would resolved by
+# `torch.cuda.synchronize()`.
+# TODO(mkozuki): Think about a way to tell P2P functions whether or not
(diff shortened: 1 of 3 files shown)


crcrpar:
@eqy @Aidyn-A could you share your findings regarding openucx/ucc & facebookresearch/torch_ucc repos?

apex/transformer/pipeline_parallel/p2p_communication.py
```diff
@@ -49,32 +49,37 @@ def _run_p2pops(
...
-            torch.distributed.irecv,
-            tensor_recv_next,
-            parallel_state.get_pipeline_model_parallel_next_rank(),
+            op=torch.distributed.irecv,
+            tensor=tensor_recv_next,
+            peer=parallel_state.get_pipeline_model_parallel_next_rank(),
+            group=group,
         )
         ops.append(recv_next_op)
     if len(ops) > 0:
```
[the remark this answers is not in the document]
crcrpar: Is that a problem emerging from the specific combination of NCCL as default and UCC for P2P?
In other words, if all the PGs are either NCCL or UCC, isn't that harmful?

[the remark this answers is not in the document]
Aidyn-A: Yes, this is the case solely for hybrid setup.
NCCL-only and UCC-only PGs work absolutely fine.

crcrpar:
https://github.com/NVIDIA/apex/blob/1403c21acf87b0f2245278309071aef17d80c13b/apex/transformer/testing/distributed_test_base.py#L117 should be "tcp,cuda". I remember I tried this right after I got started but why this is still here...

Aidyn-A:
>https://github.com/NVIDIA/apex/blob/1403c21acf87b0f2245278309071aef17d80c13b/apex/transformer/testing/distributed_test_base.py#L117
>should be "tcp,cuda". I remember I tried this right after I got started but why this is still here...

@eqy, could this be the reason behind that deadlock?

eqy:
> > https://github.com/NVIDIA/apex/blob/1403c21acf87b0f2245278309071aef17d80c13b/apex/transformer/testing/distributed_test_base.py#L117
> > 
> > should be "tcp,cuda". I remember I tried this right after I got started but why this is still here...
> 
> @eqy, could this be the reason behind that deadlock?

Thanks, I'll recheck...

crcrpar:
@Aidyn-A @eqy could you review this, please?

apex/transformer/pipeline_parallel/p2p_communication.py
```diff
@@ -41,43 +41,74 @@ def get(self):
...
+        tensor_send_next:
+        tensor_recv_prev:
+        tensor_recv_next:
+        async_comm:
+
+    Returns:
+        Tuple of 4 elements either :obj:`None` or distributed request object.
+    """
     ops = []
+    group = parallel_state.get_pipeline_model_parallel_group()
```
Aidyn-A, line 75: I just realized that we can do something like:

```
group = parallel_state.get_pipeline_model_parallel_group()
group2 = parallel_state.get_model_parallel_group()
_different_pg = group.name() == group2.name()
```

crcrpar: It's a matter of preference/taste, but I prefer to expose this kind of things as possible to do on behalf of them, behind their back...
That said, it could also be used to warn users in some occasions e.g. different backends are used but the arg isn't set.

crcrpar:
probably related in the near future: https://github.com/pytorch/pytorch/pull/79918

apex/transformer/testing/distributed_test_base.py
```diff
@@ -114,7 +114,7 @@ def _setup_pre_spawn(self) -> None:
 
         self._has_ucx_tls = "UCX_TLS" in os.environ
         if not self._has_ucx_tls:
-            os.environ["UCX_TLS"] = "tcp,cuda_copy"
+            os.environ["UCX_TLS"] = "tcp,cuda"
         print('os.environ[\"UCX_TLS\"] = {}'.format(os.environ["UCX_TLS"]))
```
Aidyn-A, line 118: For correct usage with https://github.com/Sergei-Lebedev/torch_ucc/commits/topic/active_set
I suggest to add `UCC_TLS=ucp`
```suggestion
        print('os.environ[\"UCX_TLS\"] = {}'.format(os.environ["UCX_TLS"]))
        
        self._has_ucc_tls = "UCC_TLS" in os.environ
        if not self._has_ucc_tls:
            os.environ["UCC_TLS"] = "ucp"
        print('os.environ[\"UCC_TLS\"] = {}'.format(os.environ["UCC_TLS"]))
```

crcrpar: nice catch

crcrpar:
close in favor of #1417

PR #1396, opened 2022-05-31
[transformer] Port Sequence Parallel

crcrpar:
- `tests/L0/run_transformer/test_transformer_module.py` is failing therefore is being skipped due to the lack of appropriate updates in apex/transformer/testing.
- apex.transformer expects tensors to be 3D and have the shape of (sequence, batch, feature).
- some operations in sequence parallel region takes care of certain chunk of tensors split in the dimension of sequence.
- gradients of params in sequence parallel region should be all_reduce'ed.

Ref: https://github.com/NVIDIA/Megatron-LM/releases/tag/v3.0.2

PR #1400, opened 2022-06-07
[transformer] Port Sequence Parallelism (takeover of #1396)

crcrpar:
To run GPT/Bert script with sequence parallel enabled, run the script with `--sequence-parallel` command line option.

## Changes
- apex.transformer expects tensors to be 3D and have the shape of (sequence, batch, feature).
- some operations in sequence parallel region takes care of certain chunk of tensors split in the dimension of sequence.
- gradients of params in sequence parallel region should be all_reduce'ed.

My reference was 
Ref: https://github.com/NVIDIA/Megatron-LM/releases/tag/v3.0.2.

## TODO
in follow-up PRs...
- [ ] update the `test_transformer_module.py` so that both with and without sequence parallel enabled are tested
- [ ] [optional] run GPT/Bert in autocast context
- [ ] take benchmark/profile to see the effects on performance
- [ ] refactor p2p_communication.py crcrpar/apex#4 can be referenced.


tests/L0/run_transformer/run_gpt_minimal_test.py
```diff
@@ -132,10 +134,25 @@ def train(model, optim, pipeline_model_parallel_size, async_comm):
...
         optim.zero_grad()
         fwd_bwd_func(
-            fwd_step_func, batch, model, forward_only=False, tensor_shape=tensor_shape, async_comm=async_comm
+            fwd_step_func,
+            batch,
+            model,
+            forward_only=False,
+            tensor_shape=tensor_shape,
+            async_comm=async_comm,
+            sequence_parallel_enabled=args.sequence_parallel,
```
crcrpar, line 143:
```suggestion
            sequence_parallel_enabled=args.sequence_parallel,
            dtype=torch.float32,
```

PR #1401, opened 2022-06-13
ZeRO-2 support in DistributedFusedAdam

timmoon10:
This builds on the work in PR #1242 so that we can fully distribute the gradient buffers in `DistributedFusedAdam`:
- Lazily initialize parameters to allow good compute/communication overlap in backward pass.
- Store optimizer state and gradients in fixed-sized buckets rather than a single large buffer.
- Dynamically initialize/allocate/deallocate buckets so that we do not need to store all gradients in first backward pass.
- Implement `no_sync` context to support micro-batching (similar to `DistributedDataParallel.no_sync`).
- Miscellaneous code cleanup and documentation.

Running GPT-2 on a Selene node:
| Algorithm | Step time | Required memory | Used memory |
|---|---|---|---|
| Data-parallel | 441 ms | 18.4 GB | 18.8 GB |
| ZeRO | 433 ms | 14.2 GB | 14.6 GB |

Pinging @thorjohnsen. Closes #1242.


thorjohnsen:
I am not very familiar with this code anymore, I haven't worked on it for a couple of years. I believe this code may be in use by mlperf benchmarks like BERT. Before I approve it, I will ask somebody from that team to have a look.

crcrpar:
@timmoon10 Hi Tim, could you tell me how you've run tests/L0/run_optimizers/test_dist_adam.py to appropriately verify the behavior in distributed setup?

crcrpar:
also personally removing two files looks a bit too drastic to me.

timmoon10:
Currently, `test_dist_adam.py` isn't integrated into the main L0 test script and needs to be run manually:
```bash
torchrun --nproc_per_node=${NUM_GPUS} tests/L0/run_optimizers/test_dist_adam.py
```
Would it be better to move this test to the `L1` or `distributed` test dirs? That would avoid the need to add distributed setup to the L0 tests.

I think the plan was for the two other implementations to be temporary (https://github.com/NVIDIA/apex/pull/753#issuecomment-627590615), but we can restore them if there are any downstream users.

crcrpar:
An ideal changes in my humble opinion include

- to make the test executable by python unittest. i.e. make that discoverable by python unittest runner
- to clarify how the two custom extensions are used

Given where the definition is and its dependencies, the test might be better hosted under https://github.com/NVIDIA/apex/tree/master/apex/contrib/test.

PR #1405, opened 2022-06-17
Gradient clipping with fused kernels

timmoon10:
I've noticed that gradient clipping is taking a significant amount of runtime in one of my applications, mostly in kernel launch overheads. Fortunately, we already have fused kernels to compute L2 norms for our optimizers. I've just copy-pasted [PyTorch's implementation](https://pytorch.org/docs/stable/generated/torch.nn.utils.clip_grad_norm_.html) and added a special case when the gradients are FP32 or FP16, they are on GPU, and we are computing a 2-norm.

I'm not sure if this is the right place for this, or if we should upstreaming this to PyTorch. In any case, the test passes for me.

### DIFF
--- a/apex/contrib/test/clip_grad/test_clip_grad.py
+++ b/apex/contrib/test/clip_grad/test_clip_grad.py
@@ -40,7 +40,6 @@ class ClipGradNormTest(unittest.TestCase):
     def setUp(self, seed=1234):
         random.seed(seed)
         torch.manual_seed(seed)
-        torch.cuda.manual_seed_all(seed)
 
     def test_matches_pytorch(
             self,
@@ -50,6 +49,7 @@ class ClipGradNormTest(unittest.TestCase):
             max_norm=0.54321,
             norm_type=2.0,
             rtol=1e-3,
+            atol=1e-15
     ):
         """Make sure PyTorch and Apex gradient clipping produce same results"""
         torch_params, apex_params = make_params(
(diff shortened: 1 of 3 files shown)


apex/contrib/test/clip_grad/test_clip_grad.py
```diff
@@ -0,0 +1,121 @@
...
+        return params, params_copy
+    else:
+        return params
+
+class ClipGradNormTest(unittest.TestCase):
+
+    def setUp(self, seed=1234):
+        random.seed(seed)
+        torch.manual_seed(seed)
+        torch.cuda.manual_seed_all(seed)
```
crcrpar, line 43: [nit-picking] `torch.manual_seed` should be sufficient.
```suggestion
        random.seed(seed)
        torch.manual_seed(seed)
```

Ref: https://pytorch.org/docs/stable/notes/randomness.html#controlling-sources-of-randomness

apex/contrib/test/clip_grad/test_clip_grad.py
```diff
@@ -0,0 +1,121 @@
+import random
+import unittest
+
+import torch
+from apex.contrib.clip_grad import clip_grad_norm_
+
+def make_params(
```
crcrpar, line 7: I like this function. nice.

apex/contrib/test/clip_grad/test_clip_grad.py
```diff
@@ -0,0 +1,121 @@
...
+            ),
+            msg=('Discrepancy in gradient norms: '
+                 f'torch = {torch_norm.item()}, apex = {apex_norm.item()})'))
+        for torch_p, apex_p in zip(torch_params, apex_params):
+            torch_g = torch_p.grad.to(dtype=torch.float64, device='cpu')
+            apex_g = apex_p.grad.to(dtype=torch.float64, device='cpu')
+            g_rel_err = torch.max(torch.abs((torch_g-apex_g)/(torch_g+1e-32)))
+            self.assertTrue(
+                torch.equal(torch_p, apex_p),
+                msg=f'Torch and Apex parameters are not identical')
```
crcrpar, line 85: [question] This is here to guarantee that the new clip_grad_norm does __not__ affect the input parameters. Is my interpretation correct?

apex/contrib/test/clip_grad/test_clip_grad.py
```diff
@@ -0,0 +1,121 @@
...
+                torch_norm.to(dtype=torch.float64),
+                apex_norm.to(dtype=torch.float64),
+                rtol=rtol,
+            ),
+            msg=('Discrepancy in gradient norms: '
+                 f'torch = {torch_norm.item()}, apex = {apex_norm.item()})'))
+        for torch_p, apex_p in zip(torch_params, apex_params):
+            torch_g = torch_p.grad.to(dtype=torch.float64, device='cpu')
+            apex_g = apex_p.grad.to(dtype=torch.float64, device='cpu')
+            g_rel_err = torch.max(torch.abs((torch_g-apex_g)/(torch_g+1e-32)))
```
crcrpar, line 82: see the comment about torch.testing.assert_close
```suggestion
```

crcrpar (commented):
I don't have any objection to add this to apex/contrib

apex/contrib/test/clip_grad/test_clip_grad.py
```diff
@@ -0,0 +1,121 @@
...
+            self.assertTrue(
+                torch.equal(torch_p, apex_p),
+                msg=f'Torch and Apex parameters are not identical')
+            self.assertTrue(
+                torch.allclose(torch_g, apex_g, rtol=rtol),
+                msg=f'Discrepancy in gradients: relative error = {g_rel_err}')
```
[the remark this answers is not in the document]
timmoon10: That's much cleaner. Done.

apex/contrib/test/clip_grad/test_clip_grad.py
```diff
@@ -0,0 +1,124 @@
...
+    def test_matches_pytorch_infnorm(self):
+        self.test_matches_pytorch(norm_type=float('inf'))
+
+    def test_matches_pytorch_1norm(self):
+        self.test_matches_pytorch(norm_type=1.0)
+
+    def test_raises_on_mismatch(self):
+        self.assertRaises(
+            AssertionError, self.test_matches_pytorch, rtol=1e-20)
+
```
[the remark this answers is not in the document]
timmoon10: I've been using the PyTorch NGC 22.04 container and I'm running on a workstation with a 3080. It's possible, although unlikely, that the two implementations got the same result, so I should change it to force a mismatch.

crcrpar (approved):

PR #1406, opened 2022-06-17
Move distributed Adam unit test to contrib dir

timmoon10:
This is based on suggestions from @crcrpar in #1401. I've added `unittest` support, although it assumes that the script has been called with `torchrun` or `torch.distributed.launch` with `--use_env=True`.


crcrpar (commented):
`torchrun --nproc_per_node 4 python test_dist_adam.py` locally hanged. What was your command to run the test?

timmoon10:
I ran `torchrun --nproc_per_node 2 test_dist_adam.py` on a workstation with a 3080 and a P400. I should try running on more GPUs to see how that affects the numerical error.

timmoon10:
- I think we were running into numerical instability because we were dividing by a small number in the Adam step. I've adjusted the hyperparameters and random number generation to keep the denominator manageable.
- I reproduced @crcrpar's hang on 4 GPUs, but it's gone away now. I think the deadlock happened because a test failed on some GPUs and passed on others.

crcrpar (commented):
okay, the hang is gone but the hang has some chance to revive as it's difficult to guarantee that the test is going to keep sane.
so could you try use https://github.com/NVIDIA/apex/blob/1403c21acf87b0f2245278309071aef17d80c13b/apex/transformer/testing/distributed_test_base.py#L83 to let us run this test via pytest or python while using multiple devices?
I think https://gist.github.com/crcrpar/604c16a971f926f31104d592f2176662 would let you run the test without torchrun while using multiple devices

timmoon10:
Good call, that'll be more robust. With your code, the test passes on 1, 2, 3, and 4 V100s.

crcrpar (approved):
thanks for your nice refactoring.

PR #1407, opened 2022-06-21
Temporary Solution to Let `FusedAdam` support BFloat16

crcrpar:
- [ ] Refactor tests. The existing float16 case seems to have failed due to the hardcoded threshold values, which is why I added an argument of `skip_assert` in order for the test to pass by skipping numerical checks


crcrpar:
Thank you @Aidyn-A for sanity check.

PR #1411, opened 2022-06-22
[UCC][TORCH_UCC]Do integer driver version comparison for UCC

eqy:
Avoid false-negative comparisions e.g., "470.100.0" < "470.42.0" that happen with string comparison.

CC @crcrpar

### DIFF
--- a/apex/transformer/testing/distributed_test_base.py
+++ b/apex/transformer/testing/distributed_test_base.py
@@ -1,6 +1,7 @@
 import os
 import sys
 import unittest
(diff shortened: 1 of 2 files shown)


Aidyn-A (approved):
Looks good to me :+1:

crcrpar (commented):
ultimately nit-picking though could you try [`packaging`](https://github.com/pypa/packaging) to compare versions?

I think we need to add it to setup.py requirements though...

PR #1413, opened 2022-06-27
fix function call (param instead of direct)

chochowski:
I think there's a typo, instead of passed argument there's a closure


crcrpar (commented):
Where is this `_flat_split_no_shards` used?

chochowski:
looks like nowhere

PR #1414, opened 2022-07-01
Add features to distributed Adam for Megatron support

timmoon10:
This PR adds the following features to `DistributedFusedAdam`:
- Gradient clipping with a sub-set of parameters (for model-parallel parameters)
- Support for [`torch.cuda.amp.GradScaler`](https://pytorch.org/docs/stable/amp.html#torch.cuda.amp.GradScaler)
- Support for models with multiple dtypes and devices
- Option to accumulate gradients in buckets (for accumulating FP16 grads in FP32 buffers)

### DIFF
--- a/apex/contrib/optimizers/distributed_fused_adam.py
+++ b/apex/contrib/optimizers/distributed_fused_adam.py
@@ -761,14 +761,21 @@ class DistributedFusedAdam(torch.optim.Optimizer):
         self._inv_grad_scale = torch.clamp(inv_clip_coef, min=1.0).view(1)
         return total_norm
 
-    def step(self, grad_scaler=None):
+    def step(self, closure=None, *, grad_scaler=None):
         """Apply Adam optimizer step
 
         Arguments:
+            closure (callable, optional): closure to recompute loss
+                (default: None)
             grad_scaler (torch.cuda.amp.GradScaler, optional):
                 gradient scaler (default: None)
 
         """
+
+        # Apply closure
+        loss = None
+        if closure is not None:
+            loss = closure()
 
         # Make sure that gradients have been reduced
         self.grad_sync()
@@ -929,3 +936,5 @@ class DistributedFusedAdam(torch.optim.Optimizer):
         # Synchronize pipeline streams
         for stream in self._pipeline_streams:
             current_stream.wait_stream(stream)
+
+        return loss
--- a/apex/contrib/optimizers/distributed_fused_adam.py
+++ b/apex/contrib/optimizers/distributed_fused_adam.py
@@ -9,7 +9,7 @@ import threading
 import torch
 import amp_C
 from apex.multi_tensor_apply import multi_tensor_applier
-from torch.distributed.distributed_c10d import _get_default_group
+from torch.distributed.distributed_c10d import _get_default_group, _get_global_rank
 
 class DistributedFusedAdam(torch.optim.Optimizer):
     """AdamW optimizer with ZeRO algorithm.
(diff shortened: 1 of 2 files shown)


apex/contrib/optimizers/distributed_fused_adam.py
```diff
@@ -648,95 +647,152 @@ def grad_sync(self):
...
+                p-norm (default: 2)
+
+        """
+        assert max_norm > 0
+        total_norm = self.grad_norm(parameters=parameters, norm_type=norm_type)
+        inv_clip_coef = (total_norm + 1e-6) / max_norm
+        self._inv_grad_scale = torch.clamp(inv_clip_coef, min=1.0).view(1)
+        return total_norm
+
+    def step(self, grad_scaler=None):
```
crcrpar, line 764: My gut feeling says it's not safe to remove `closure`

apex/contrib/optimizers/distributed_fused_adam.py
```diff
@@ -648,95 +647,152 @@ def grad_sync(self):
...
+        # Note: We compute gradient norm to check for non-finite
+        # values. This is more conservative and compute intensive than
+        # directly checking, but it avoids extra communication if we
+        # have already computed gradient norm e.g. for gradient
+        # clipping.
+        if grad_scaler is not None:
+            grad_norm = self.grad_norm()
+            found_inf = torch.logical_not(torch.isfinite(grad_norm))
+            scaler_state = grad_scaler._per_optimizer_states[id(self)]
+            scaler_state['found_inf_per_device'] = {found_inf.device: found_inf.float()}
```
crcrpar, line 786: This cast here is redundant if you assume that `grad_scaler` is an instance of `torch.cuda.amp.GradScaler`
```suggestion
            scaler_state['found_inf_per_device'] = {found_inf.device: found_inf}
```

apex/contrib/optimizers/distributed_fused_adam.py
```diff
@@ -648,95 +647,152 @@ def grad_sync(self):
...
+            if found_inf.item():
+                return
+            else:
+                assert grad_scaler._scale is not None
+                self._inv_grad_scale *= grad_scaler._scale
+        inv_grad_scale = self._inv_grad_scale.item()
```
crcrpar, line 792: This could incur a device-to-host copy.

apex/contrib/optimizers/distributed_fused_adam.py
crcrpar, line 764: also, you can define `grad_scaler` as a keyword only argument given https://github.com/pytorch/pytorch/blob/fe73528a904aa0627b5a6f7376a2331cb5985bfa/torch/cuda/amp/grad_scaler.py#L329.

timmoon10: Good call, I ran into a problem in PyTorch Lightning without it.

apex/contrib/optimizers/distributed_fused_adam.py
timmoon10: Unfortunately it is needed as a CPU arg in the Adam kernel. I think the next step will be to tweak some of the kernels, but I'd like to keep this PR Python-only for now.

apex/contrib/optimizers/distributed_fused_adam.py
timmoon10: I don't think `GradScaler` does any casting. I got runtime errors when I passed in a `bool`.

crcrpar: uh this logic overloads `torch.cuda.amp.GradScaler`'s `found_inf` logic. I see

apex/contrib/optimizers/distributed_fused_adam.py
crcrpar: makes sense.

crcrpar (approved):
Verified the unittest with nightly, `nvcr.io/nvidia/pytorch:22.06-py3`, `pytorch/pytorch:1.12.0-cuda11.3-cudnn8-devel`

PR #1417, opened 2022-07-07
[transformer] UCC async test

Aidyn-A:
This is the test for async UCC. It is almost the same as `test_pipeline_parallel_fwd_bwd.py` except for deleted sequence parallel and NCCL tests and couple of more modifications. 
The batch and microbatch sizes are increased because I wanted to see the performance under the load. 
At this stage it is set up such that it squeezes maximum memory of V100 with 16GB.
Please let me know your opinion. 
Perhaps we can rewrite `test_pipeline_parallel_fwd_bwd.py` to do the thing we want.

I ran this test on UCX-master,  UCC-master and torch_ucc-topic/active_set.

cc @eqy @crcrpar @ptrblck


eqy:
Since the body of the test is essentially same, would it be difficult to write this as an extension of the existing tests? The larger batch sizes/shapes could also be added as arguments with defaults set to existing values and overriden with the ucc-async versions of the tests.

apex/transformer/pipeline_parallel/p2p_communication.py
```diff
@@ -53,35 +55,50 @@ def _run_p2pops(
     async_comm: bool = False
 ):
     ops = []
+    p2p_group = parallel_state.get_pipeline_model_parallel_group()
+    default_group = parallel_state.get_model_parallel_group()
+
+    ucc_tls_ucp = False
+    if "UCC_TLS" in os.environ:
```
Aidyn-A, line 62: We do not need `torch.cuda.synchronize()` for `ucp` transport layer.

tests/L0/run_transformer/test_pipeline_parallel_fwd_bwd_ucc_async.py
```diff
@@ -0,0 +1,64 @@
...
+class UccPipelineParallelForwardBackwardUnderLoadTest(UccDistributedTestBase, PipelineParallelForwardBackwardTestBase):
+
+    def __init__(self, *args, **kwargs) -> None:
+        super().__init__(*args, **kwargs)
+        self.GLOBAL_BATCH_SIZE = 8192
+        self.MICRO_BATCH_SIZE =  self.GLOBAL_BATCH_SIZE // self.world_size
+        self.HIDDEN_SIZE = 256
+        self.NUM_EPOCHS = 1
+        self.deallocate_options = (False,)
+        self.dtypes = (torch.float32,)
```
Aidyn-A, line 53: double is much slower than float on V100.

eqy: Is it slow enough that the test becomes annoying to run? Otherwise we might consider gating this on compute capability e.g., as https://pytorch.org/docs/stable/generated/torch.cuda.get_device_capability.html `>= (8, 0)` (Ampere) starts to introduce potentially fishy behavior due to TF32 that becomes more involved to account for.

tests/L0/run_transformer/test_pipeline_parallel_fwd_bwd.py
```diff
@@ -170,95 +171,108 @@ def _forward_backward_test_impl(
...
+                        self.HIDDEN_SIZE,
+                        self.HIDDEN_SIZE,
+                    ),
+                    dtype=dtype,
+                    async_comm=async_comm,
+                    grad_scaler=grad_scaler,
+                    deallocate_pipeline_output=deallocate_pipeline_outputs,
+                )
+
+                if dtype == torch.double and epoch == 0:
```
eqy, line 192: Might need to adjust this condition if one test variant has `float` specified as the dtype.

tests/L0/run_transformer/test_pipeline_parallel_fwd_bwd_ucc_async.py
```diff
@@ -0,0 +1,37 @@
...
+from apex.transformer.testing.distributed_test_base import UccDistributedTestBase
+
+from test_pipeline_parallel_fwd_bwd import PipelineParallelForwardBackwardTestBase
+
+logging.getLogger("torch").setLevel(logging.WARNING)
+logging.getLogger("apex").setLevel(logging.WARNING)
+
+weight_coeff = 1024
+
+class UccPipelineParallelForwardBackwardUnderLoadTest(UccDistributedTestBase, PipelineParallelForwardBackwardTestBase):
```
crcrpar, line 17: What does class test?

tests/L0/run_transformer/test_pipeline_parallel_fwd_bwd.py
```diff
@@ -170,95 +171,108 @@ def _forward_backward_test_impl(
...
+                            params = list(model_module.parameters())
+                            rank = params[0].get_device()
+                            offset = pipeline_model_parallel_world_size
+                            param_id = rank // data_parallel_size + vm_id * offset
+                            target_params = target_model[param_id]
+
+                            torch.testing.assert_close(params[0].cpu(), target_params[0])
+                            torch.testing.assert_close(params[1].cpu(), target_params[1])
+                            torch.testing.assert_close(params[0].grad.cpu() / microbatch_size, target_params[0].grad)
+                            torch.testing.assert_close(params[1].grad.cpu() / microbatch_size, target_params[1].grad)
```
crcrpar, line 215: Would this numerical check be accurate even with `NUM_EPOCHS > 1`?

crcrpar (commented):
leaving some minor comments but it in general looks good

tests/L0/run_transformer/test_pipeline_parallel_fwd_bwd_ucc_async.py
Aidyn-A: There are two main reasons for existence of this class. Compared to the original `test_pipeline_parallel_fwd_bwd.py` the load here is much higher, so one would be able to see the operations in profiler (nsight systems). Second, when running `test_pipeline_parallel_fwd_bwd.py -k test_learning_async_pipelining_without_interleaving` it will run the tests for NCCL and NCCL/UCC as well, which is often unwanted.

tests/L0/run_transformer/test_pipeline_parallel_fwd_bwd_ucc_async.py
Aidyn-A: Sounds like a good idea, I added the compute capability and applied it to another your comment as well.

tests/L0/run_transformer/test_pipeline_parallel_fwd_bwd_ucc_async.py
crcrpar: > Second, when running `test_pipeline_parallel_fwd_bwd.py -k test_learning_async_pipelining_without_interleaving` it will run the tests for NCCL and NCCL/UCC as well, which is often unwanted.

It's because the expression is a bit too ambiguous. For example, `UccPipelineParallelForwardBackwardTest.test_learning_async_pipelining_without_interleaving` wold only run one test for unittest. For pytest "test_learning_async_pipelinign_without_interleaving and Ucc" would only run one case.

tests/L0/run_transformer/test_pipeline_parallel_fwd_bwd.py
```diff
@@ -90,6 +97,7 @@ class PipelineParallelForwardBackwardTestBase:
     GLOBAL_BATCH_SIZE = 16
     MICRO_BATCH_SIZE = 2
     HIDDEN_SIZE = 32
+    NUM_EPOCHS = 1
```
eqy, line 100: Nit: Is this an Epoch or a fwd-bwd iteration?

apex/transformer/pipeline_parallel/p2p_communication.py
```diff
@@ -53,35 +55,48 @@ def _run_p2pops(
     async_comm: bool = False
 ):
     ops = []
+    p2p_group = parallel_state.get_pipeline_model_parallel_group()
+    default_group = parallel_state.get_model_parallel_group()
+
+    ucc_tls_ucp = "ucp" == os.getenv("UCC_TLS", "")
+    need_to_sync = p2p_group.name() != default_group.name() and not ucc_tls_ucp
```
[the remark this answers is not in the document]
Aidyn-A: Yes, when I finalize the PR, I will add more details about that. Recently I found out that `ucp` transport layer is based on CUDA kernels, so it is stream aware.

tests/L0/run_transformer/test_pipeline_parallel_fwd_bwd.py
Aidyn-A: I believe it is both. We can name it `NUM_FWD_BWD_ITERATIONS`. What do you think?

tests/L0/run_transformer/test_pipeline_parallel_fwd_bwd_ucc_async.py
eqy: Do we expect this test to uncover any additional failures that would not be covered by the previous use-cases, or would it be for profiling purposes only? (Would it fit under the umbrella of a slow test that is usually skipped?)

tests/L0/run_transformer/test_pipeline_parallel_fwd_bwd_ucc_async.py
```diff
@@ -0,0 +1,32 @@
...
+logging.getLogger("apex").setLevel(logging.WARNING)
+
+class UccPipelineParallelForwardBackwardUnderLoadTest(UccDistributedTestBase, PipelineParallelForwardBackwardTestBase):
+
+    def __init__(self, *args, **kwargs) -> None:
+        super().__init__(*args, **kwargs)
+        self.GLOBAL_BATCH_SIZE = 1024
+        self.MICRO_BATCH_SIZE = 64
+        self.HIDDEN_SIZE = 32
+        self.NUM_EPOCHS = int(os.getenv("NUM_EPOCHS", "1"))
```
crcrpar, line 22: [optional] can we avoid this dependence on an env var? e.g. adding a new argument of `num_epochs` to the impl.

tests/L0/run_transformer/test_pipeline_parallel_fwd_bwd_ucc_async.py
Aidyn-A: The main purpose of this file was to do the profiling. And do I think this is not the best folder for hosting it. I am thinking of moving it somewhere to a new location like `apex/profiling/transformer` or something similar. 
@crcrpar @eqy what do you think?

crcrpar: I'd put it in tests/L1/transformer and I'm not sure if it's needed to make this file testable.

tests/L0/run_transformer/test_pipeline_parallel_fwd_bwd.py
```diff
@@ -187,7 +194,8 @@ def _forward_backward_test_impl(
                 deallocate_pipeline_output=deallocate_pipeline_outputs,
             )
 
-            if dtype == torch.double:
+            if dtype == get_dtype_for_comparison():
+                torch.cuda.synchronize()
```
crcrpar, line 198: QQ: I think this would let us run numerical checks in single precision if device prior to Ampere.
Is this expected?

eqy (approved):

tests/L0/run_transformer/run_bert_minimal_test.py
```diff
@@ -186,7 +186,12 @@ def train(
     failure = None
     init = True
     try:
-        for virtual_pipeline_model_parallel_size in (2, None):
+        virtual_pipeline_model_parallel_sizes = (None, 2,)
+        if HAS_TORCH_UCC:
+            # Deliberately skipping test with interleaved schedule for BERT model.
+            # It deadlocks on hybrid UCC/NCCL backend.
+            virtual_pipeline_model_parallel_sizes = (None,)
+        for virtual_pipeline_model_parallel_size in virtual_pipeline_model_parallel_sizes:
```
Aidyn-A, line 194: This test is deadlocking if test is running with interleaved schedule (`virtual_pipeline_model_parallel_sizes>=2`) on hybrid backend (NCCL for collectives and UCC for p2p). We will have to revisit interleaved schedule to fix the hang.

tests/L1/transformer/pipeline_parallel_fwd_bwd_ucc_async.py
```diff
@@ -0,0 +1,229 @@
...
+
+    if pipeline_model_parallel_world_size is None:
+        pipeline_model_parallel_world_size =  world_size // (tensor_model_parallel_world_size * data_parallel_size)
+    else:
+        data_parallel_size = world_size // (tensor_model_parallel_world_size * pipeline_model_parallel_world_size)
+
+    return tensor_model_parallel_world_size, data_parallel_size, pipeline_model_parallel_world_size
+
+
+class UccPipelineParallelForwardBackwardProf(UccDistributedTestBase):
```
eqy, line 66: Perhaps add a comment that indicates that this "test" is maintained for Prof[iling] purposes to check overlapping---which explains why it is safe to usually skip and that is the reason the usual numerical checks also removed.

eqy (approved):

crcrpar (approved):

PR #1421, opened 2022-07-13
Time dimension shape check for fused scale mask softmax kernel

MaximumEntropy:
Signed-off-by: MaximumEntropy <sandeep.subramanian.1@umontreal.ca>


crcrpar (commented):
could you add a test case which had failed without this patch to https://github.com/NVIDIA/apex/blob/master/tests/L0/run_transformer/test_fused_softmax.py?

PR #1432, opened 2022-07-22
Improvements in distributed Adam optimizer for Megatron

timmoon10:
This PR has several optimizations to enable good memory scaling in Megatron:
- Add option to allocate gradient buckets out of one large buffer. This makes it easier to access these buffers externally, e.g. for model-parallel communication, at the expense of preventing ZeRO-2 memory scaling.
- Add option to manually initialize params instead of waiting for first backward pass.
- Support param allgathers with any dtype.
- Avoid allocating temporary buffers in side streams, since it made memory pool less efficient.
- When saving state for checkpointing, gather all state on root rank.
- Replace internal dicts with container classes.

The tests pass for me when running on 4 V100s.

### DIFF
--- a/apex/contrib/optimizers/distributed_fused_adam.py
+++ b/apex/contrib/optimizers/distributed_fused_adam.py
@@ -3,6 +3,7 @@ import contextlib
 import enum
 import importlib
 import inspect
+import io
 import math
 import threading
 
@@ -10,6 +11,10 @@ import torch
 import amp_C
 from apex.multi_tensor_apply import multi_tensor_applier
 from torch.distributed.distributed_c10d import _get_default_group, _get_global_rank
+
+def _round_to_multiple(number, multiple, round_up=True):
+    """Assumes arguments are positive integers"""
+    return (number+multiple-1 if round_up else number) // multiple * multiple
 
 class DistributedFusedAdam(torch.optim.Optimizer):
     """AdamW optimizer with ZeRO algorithm.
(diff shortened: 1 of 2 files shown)


apex/contrib/optimizers/distributed_fused_adam.py
```diff
@@ -86,6 +96,56 @@ class DistributedFusedAdam(torch.optim.Optimizer):
 
     """
 
+    class ParameterFragment():
```
crcrpar, line 99: we don't need an empty `()` per https://docs.python.org/3/tutorial/classes.html#class-definition-syntax.

```suggestion
    class ParameterFragment:
```

apex/contrib/optimizers/distributed_fused_adam.py
```diff
@@ -86,6 +96,56 @@ class DistributedFusedAdam(torch.optim.Optimizer):
...
+            # Whether fragment is in local shard of bucket
+            self.in_local_shard = in_local_shard
+            # Range within local shard
+            self.shard_range = shard_range
+            # Range of local fragment shard within bucket
+            self.shard_bucket_range = shard_bucket_range
+            # Range of local fragment shard within parameter
+            self.shard_param_range = shard_param_range
+
+    class StateBucket():
```
crcrpar, line 137:
```suggestion
    class StateBucket:
```

apex/contrib/optimizers/distributed_fused_adam.py
```diff
@@ -97,6 +157,26 @@ class GradientStatus(enum.Enum):
         # Asynchronous reduction is in progress
         SYNCING = enum.auto()
 
+    class GradientBucket():
```
crcrpar, line 160:
```suggestion
    class GradientBucket:
```

crcrpar (commented):
out of curiosity, is there any reason to define some classes inside `DistributedFusedAdam`?
It looks like we can have them as first citizen while assuming you follow how `GradientStatus` is defined.

timmoon10:
The helper classes are mostly for convenience and all functionality is provided by `DistributedFusedAdam` directly manipulating the class members. I prefer to think of them as internal state of `DistributedFusedAdam` and don't think they should be interacted with externally. That said, if we make them more standalone and encapsulated, it would be reasonable to bring them out of `DistributedFusedAdam` so we can use them to build other distributed optimizers.

PR #1416, opened 2022-07-06
update: mpu for t5 rpe

PerkzZheng:
support t5 relative position embedding mpu groups.

PR #1435, opened 2022-07-25
Fix bug when initializing model-parallel process groups for GPT-3

timmoon10:
PR #1416 causes me to experience runtime errors when initializing model-parallel groups for GPT-3 training. I think it is not properly initializing some T5-specific process groups (namely `encoder_relative_position_embedding_ranks`) and passing empty lists to `torch.distributed.new_group`, which is not supported.

This fix allows me to run the NeMo-Megatron GPT-3 implementation (8 V100s, 2-way pipeline, tensor, data parallelism). However, sanity-checking would be appreciated. Pinging @PerkzZheng and @crcrpar.

### DIFF
--- a/tests/L0/run_transformer/test_parallel_state.py
+++ b/tests/L0/run_transformer/test_parallel_state.py
@@ -132,6 +132,50 @@ class ParallelStateTestBase:
 
         parallel_state.destroy_model_parallel()
 
+    def test_initialize_model_parallel_decoder_only(self) -> None:
+        """Initialize model parallelism for decoder-only Transformers like GPT-3"""
+
+        self.assertFalse(parallel_state.model_parallel_is_initialized())
+
+        for tensor_model_parallel_world_size in range(1, self.world_size + 1):
+            with self.subTest(
+                tensor_model_parallel_world_size=tensor_model_parallel_world_size
+            ):
+                if self.world_size % tensor_model_parallel_world_size:
+                    continue
+
+                pipeline_model_parallel_world_size = (
+                    self.world_size // tensor_model_parallel_world_size
+                )
+
+                parallel_state.initialize_model_parallel(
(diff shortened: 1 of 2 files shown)


crcrpar:
do you have a minimal reproducible script?

crcrpar:
we have a so simple unittest which basically just calls `initialize_model_parallel` but we haven't seen such error. So ideally could you add a failing case which this PR would access?

timmoon10:
I've added a test that fails without this fix and succeeds with it. I also played around with a test for encoder-only Transformers like BERT, but it failed.

tests/L0/run_transformer/test_parallel_state.py
```diff
@@ -132,6 +132,50 @@ def test_initialize_model_parallel_with_virtual_and_split(self) -> None:
...
+                    continue
+
+                pipeline_model_parallel_world_size = (
+                    self.world_size // tensor_model_parallel_world_size
+                )
+
+                parallel_state.initialize_model_parallel(
+                    tensor_model_parallel_size_=tensor_model_parallel_world_size,
+                    pipeline_model_parallel_size_=pipeline_model_parallel_world_size,
+                    pipeline_model_parallel_split_rank_=0,
```
crcrpar, line 154: out of curiosity, when do you set split rank to 0? I'm not sure when GPT like models (= ModelType.encoder_or_decoder) would require this value to 0.

timmoon10: I see that the value is 0 when I print inside `initialize_model_parallel`. It seems that the default value is set in NeMo at https://github.com/NVIDIA/NeMo/blob/f1bf6c2731c6d4881b0063aa26564d9cd8300c4c/nemo/collections/nlp/models/language_modeling/megatron_base_model.py#L82.

PR #1439, opened 2022-07-28
Add BF16 support to FusedMixedPrecisionLamb

nv-joseli:
Added bfloat16 dispatch to kernels used in the FusedMixedPrecisionLamb optimizer.

### DIFF
--- a/tests/L0/run_optimizers/test_lamb.py
+++ b/tests/L0/run_optimizers/test_lamb.py
@@ -196,7 +196,7 @@ class TestLamb(unittest.TestCase):
 
         return max_abs_diff, max_rel_diff
 
-    def gen_single_type_test(self, param_type=torch.float, device="cuda"):
+    def gen_single_type_test(self, param_type=torch.float, device="cuda", iters=None):
         nelem = 278011
         tensor = torch.rand(nelem, dtype=param_type, device=device)
         weight_decay = [0, 0.01]
@@ -213,19 +213,14 @@ class TestLamb(unittest.TestCase):
                     # As long as this parameter is set before the first call to step(), 
                     # then it should act normally.
                     tst_optim.reduced_precision_dtype = param_type
-
-            for i in range(self.iters):
+            print(f"ITERS: {self.iters}")
+            for i in range(self.iters if iters is None else iters):
                 self.gen_grad(ref_param, tst_param)
                 ref_optim.step()
                 torch.cuda.synchronize()
                 tst_optim.step()
                 torch.cuda.synchronize()
-                torch.testing.assert_close(
-                    tst_param, 
-                    ref_param, 
-                    atol=self.max_abs_diff, 
-                    rtol=self.max_rel_diff
-                )
+                torch.testing.assert_close(tst_param, ref_param)
 
 class TestFusedLAMB(TestLamb):
     def __init__(self, *args, **kwargs):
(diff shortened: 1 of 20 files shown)


tests/L0/run_optimizers/test_lamb.py
```diff
@@ -200,16 +206,21 @@ def gen_single_type_test(self, param_type=torch.float, device="cuda"):
...
-            for i in range(self.iters):
+            if isinstance(tst_optim, apex.optimizers.FusedMixedPrecisionLamb):
+                if param_type != torch.float:
+                    # joseli: This parameter is usually passed into the constructor, 
+                    # but I do not want to change the testing interface.
+                    # As long as this parameter is set before the first call to step(), 
+                    # then it should act normally.
+                    tst_optim.reduced_precision_dtype = param_type
+            print(f"ITERS: {self.iters}")
+            for i in range(self.iters if iters is None else iters):
```
crcrpar, line 217: see the comment on `test_bfloat16`

```suggestion
            for i in range(self.iters):
```

tests/L0/run_optimizers/test_lamb.py
```diff
@@ -190,7 +196,7 @@ def get_max_diff(self, ref_param, tst_param):
 
         return max_abs_diff, max_rel_diff
 
-    def gen_single_type_test(self, param_type=torch.float, device="cuda"):
+    def gen_single_type_test(self, param_type=torch.float, device="cuda", iters=None):
```
crcrpar, line 199: see the comment on `test_bfloat16`

```suggestion
    def gen_single_type_test(self, param_type=torch.float, device="cuda"):
```

tests/L0/run_optimizers/test_lamb.py
```diff
@@ -281,6 +292,9 @@ def __init__(self, *args, **kwargs):
     def test_float(self):
         self.gen_single_type_test(param_type=torch.float)
 
+    def test_bfloat16(self):
+        self.gen_single_type_test(param_type=torch.bfloat16, iters=4)
```
crcrpar, line 296: Could you try this way and check the other member methods still see `self.iters` of 7?

ref: https://docs.python.org/3/library/unittest.html#unittest.TestCase.setUp

```suggestion
        self.iters = 4
        self.gen_single_type_test(param_type=torch.bfloat16)
```

tests/L0/run_optimizers/test_lamb.py
```diff
@@ -200,16 +206,21 @@ def gen_single_type_test(self, param_type=torch.float, device="cuda"):
...
                 self.gen_param_optim([tensor], lamb_option)
 
-            for i in range(self.iters):
+            if isinstance(tst_optim, apex.optimizers.FusedMixedPrecisionLamb):
+                if param_type != torch.float:
+                    # joseli: This parameter is usually passed into the constructor, 
+                    # but I do not want to change the testing interface.
+                    # As long as this parameter is set before the first call to step(), 
+                    # then it should act normally.
+                    tst_optim.reduced_precision_dtype = param_type
```
crcrpar, line 215: If we're to change the interface, could you tell me what changes would be needed?

tests/L0/run_optimizers/test_lamb.py
nv-joseli: This works, the other member methods still use `self.iters=7` .

tests/L0/run_optimizers/test_lamb.py
nv-joseli: We would need to split `self.gen_param_optim` so that it accepts a `ref_lamb_option` and `tst_lamb_option` and whenever we are testing a `FusedMixedPrecisionLamb`, then `tst_lamb_option` will get an extra entry like 

`tst_lamb_option['reduced_precision_dtype'] = param_type`

Issue #1430, opened 2022-07-22
fused_softmax kernel modifies input_grad tensor during backward step calculation.


yidong72:
**Describe the Bug**
In the current Apex fused softmax kernel implementation, I notice the backward pass modifies the `input_grad` tensor to be the same as the backward output tensor.
```C++
      dispatch_scaled_masked_softmax_backward<scalar_t, scalar_t, float>(
          reinterpret_cast<scalar_t*>(output_grads_ptr), 
	  reinterpret_cast<scalar_t*>(output_grads_ptr), 
	  reinterpret_cast<scalar_t const*>(softmax_results.data_ptr()),
	  scale_factor,
	  query_seq_len,
	  key_seq_len,
	  batches,
	  attn_heads);
```
However, according to the Pytorch document:
<img width="679" alt="image" src="https://user-images.githubusercontent.com/43824965/180485435-749e7745-c531-4f94-9504-e13a6114b539.png">
Pytorch document says explicitly that `It is important NEVER to modify these in-place`.  The current implementation break the contract which may cause some unpredicted problems. 
We should fix it and make it behave as the document says.

Issue #1390, opened 2022-05-26
ScaledMaskedSoftmax wrong behavior when all keys are masked.


yidong72:
In `ScaledMaskedSoftmax` , currently if all the keys are masked for a certain query element, the attention is uniformly distributed among all the key elements. It means it will average all the value vectors together which is not stable since value vectors can be different. The correct behavior should return zero probability for all the masked keys in this case.

PR #1440, opened 2022-07-28
Add a generic fused softmax

yidong72:
Originally created the PR for NeMo at https://github.com/NVIDIA/NeMo/pull/4594
@ericharper  suggested to move it here

# What does this PR do ?
Implemented a generic fused softmax GPU kernel that can be a drop-in replacement of the `FusedScaleMaskSoftmax`. It addresses the following issues:
1. Removed the sequence length limitation. It works with tensor of any length.
2. Fix the `grad_input` modification issue https://github.com/NVIDIA/apex/issues/1430
3. Fix the problem when all keys are masked https://github.com/NVIDIA/apex/issues/1390

It includes a unit test to cover a few cases.
The performance is comparable to the apex one. Here is a training step time graph when it is tested with training a RETRO model.
image
The blue curve is pytorch softmax. The bottom red curve is apex fused kernel and the middle red curve the fused kernel in this PR.

### DIFF
--- a/tests/L0/run_transformer/test_fused_softmax.py
+++ b/tests/L0/run_transformer/test_fused_softmax.py
@@ -227,167 +227,70 @@ class TestFusedScaleMaskSoftmax(common_utils.TestCase):
 
 class TestGenericFusedSoftmaxKernel(common_utils.TestCase):
 
+    def _setup_qk(self):
+        qlen = [1, 2, 1234, 2322, 2348]
+        klen = [1, 2, 3, 4, 5, 8, 10, 11, 13, 128, 256, 1200, 1234, 2048, 3123, 4096, 4128, 7234, 8192, 10232]
+        return itertools.product(qlen, klen)
+ 
     def test_forward(self):
         import generic_scaled_masked_softmax_cuda
-
         batch = 2
         attn = 16
-        qlen = 2348
-        klen = 3123
         scale_t = torch.tensor([1.0])
-        for qlen in [2348, 2322, 1234, 1, 2]:
-            for klen in [
-                3123,
-                1234,
-                2,
-                4,
-                8,
-                3,
-                1,
-                5,
-                10,
-                11,
-                13,
-                128,
-                256,
-                1200,
-                2048,
-                4096,
-                7234,
-                8192,
-                10232,
-                4128,
-            ]:
-                inputs = torch.normal(0, 2, (batch, attn, qlen, klen), dtype=torch.float16, device='cuda:0')
-                masks = torch.randint(0, 2, (batch, 1, qlen, klen), dtype=torch.bool, device='cuda:0')
-                softmax_results = generic_scaled_masked_softmax_cuda.forward(inputs, masks, scale_t[0].item())
-                softmax_results_torch = forward_torch_softmax(inputs, masks, scale_t[0].item())
-                error = (softmax_results_torch - softmax_results).abs().max()
-                assert error < 1e-3
+        for qlen, klen in self._setup_qk():
+            inputs = torch.normal(0, 2, (batch, attn, qlen, klen), dtype=torch.float16, device='cuda:0')
+            masks = torch.randint(0, 2, (batch, 1, qlen, klen), dtype=torch.bool, device='cuda:0')
+            softmax_results = generic_scaled_masked_softmax_cuda.forward(inputs, masks, scale_t[0].item())
+            softmax_results_torch = forward_torch_softmax(inputs, masks, scale_t[0].item())
+            self.assertEqual(softmax_results_torch.type(torch.float16), softmax_results, atol=1e-3, rtol=1e-3)
 
     def test_backward(self):
         import generic_scaled_masked_softmax_cuda
 
         batch = 2
         attn = 16
-        qlen = 2348
-        klen = 3123
         scale_t = torch.tensor([1.0])
-        for qlen in [2348, 2322, 1234, 1, 2]:
-            for klen in [
-                3123,
-                1234,
-                2,
-                4,
-                8,
-                3,
-                1,
-                5,
-                10,
-                11,
-                13,
-                128,
-                256,
-                1200,
-                2048,
-                4096,
-                7234,
-                8192,
-                10232,
-                4128,
-            ]:
-                inputs = torch.normal(0, 2, (batch, attn, qlen, klen), dtype=torch.float16, device='cuda:0')
-                backward = torch.rand_like(inputs, dtype=torch.float16, device='cuda:0')
-                masks = torch.randint(0, 2, (batch, 1, qlen, klen), dtype=torch.bool, device='cuda:0')
-                softmax_results = generic_scaled_masked_softmax_cuda.forward(inputs, masks, scale_t[0].item())
-                back_grad = generic_scaled_masked_softmax_cuda.backward(backward, softmax_results, scale_t[0].item())
+        for qlen, klen in self._setup_qk():
+            inputs = torch.normal(0, 2, (batch, attn, qlen, klen), dtype=torch.float16, device='cuda:0')
+            backward = torch.rand_like(inputs, dtype=torch.float16, device='cuda:0')
+            masks = torch.randint(0, 2, (batch, 1, qlen, klen), dtype=torch.bool, device='cuda:0')
+            softmax_results = generic_scaled_masked_softmax_cuda.forward(inputs, masks, scale_t[0].item())
+            back_grad = generic_scaled_masked_softmax_cuda.backward(backward, softmax_results, scale_t[0].item())
 
-                inputs.requires_grad = True
-                softmax_results_torch = forward_torch_softmax(inputs, masks, scale_t[0].item())
-                softmax_results_torch.backward(backward)
-                error = (back_grad - inputs.grad).abs().max()
-                assert error < 1e-3
+            inputs.requires_grad = True
+            softmax_results_torch = forward_torch_softmax(inputs, masks, scale_t[0].item())
+            softmax_results_torch.backward(backward)
+            self.assertEqual(back_grad, inputs.grad, atol=1e-3, rtol=1e-3)
 
     def test_allmasked(self):
         import generic_scaled_masked_softmax_cuda
-
         batch = 2
         attn = 16
-        qlen = 2348
-        klen = 3123
         scale_t = torch.tensor([1.0])
-        for qlen in [2348, 2322, 1234, 1, 2]:
-            for klen in [
-                3123,
-                1234,
-                2,
-                4,
-                8,
-                3,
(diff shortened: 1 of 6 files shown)


tests/L0/run_transformer/test_fused_softmax.py
```diff
@@ -216,5 +224,171 @@ def test_autocast_fused_upper_triangle_mask_softmax(self):
...
+                8192,
+                10232,
+                4128,
+            ]:
+                inputs = torch.normal(0, 2, (batch, attn, qlen, klen), dtype=torch.float16, device='cuda:0')
+                masks = torch.randint(0, 2, (batch, 1, qlen, klen), dtype=torch.bool, device='cuda:0')
+                softmax_results = generic_scaled_masked_softmax_cuda.forward(inputs, masks, scale_t[0].item())
+                softmax_results_torch = forward_torch_softmax(inputs, masks, scale_t[0].item())
+                error = (softmax_results_torch - softmax_results).abs().max()
+                assert error < 1e-3
```
crcrpar, line 266: I'd use `self.assertEqual` here which internally calls [`torch.testing.assert_close`](https://pytorch.org/docs/stable/testing.html#torch.testing.assert_close) and specify atol and rtol as personally it'll look cleaner. (this is only true because this test case class inherits `torch.testing._internal.common_utils.TestCase`)
```python
                self.assertEqual(softmax_results_torch, softmax_results, atol=1e-3, 0)
```

tests/L0/run_transformer/test_fused_softmax.py
```diff
@@ -216,5 +224,171 @@ def test_autocast_fused_upper_triangle_mask_softmax(self):
...
+    def test_forward(self):
+        import generic_scaled_masked_softmax_cuda
+
+        batch = 2
+        attn = 16
+        qlen = 2348
+        klen = 3123
+        scale_t = torch.tensor([1.0])
+        for qlen in [2348, 2322, 1234, 1, 2]:
+            for klen in [
```
crcrpar, line 239: it'd be cleaner to use [`itertools.product`](https://docs.python.org/3/library/itertools.html#itertools.product) instead for one less indentation

tests/L0/run_transformer/test_fused_softmax.py
```diff
@@ -216,5 +224,171 @@ def test_autocast_fused_upper_triangle_mask_softmax(self):
...
+                backward = torch.rand_like(inputs, dtype=torch.float16, device='cuda:0')
+                masks = torch.randint(0, 2, (batch, 1, qlen, klen), dtype=torch.bool, device='cuda:0')
+                softmax_results = generic_scaled_masked_softmax_cuda.forward(inputs, masks, scale_t[0].item())
+                back_grad = generic_scaled_masked_softmax_cuda.backward(backward, softmax_results, scale_t[0].item())
+
+                inputs.requires_grad = True
+                softmax_results_torch = forward_torch_softmax(inputs, masks, scale_t[0].item())
+                softmax_results_torch.backward(backward)
+                error = (back_grad - inputs.grad).abs().max()
+                assert error < 1e-3
```
crcrpar, line 309: ditto, maybe simpler to just call `self.assertEqual`

tests/L0/run_transformer/test_fused_softmax.py
```diff
@@ -216,5 +224,171 @@ def test_autocast_fused_upper_triangle_mask_softmax(self):
...
+                backward = torch.rand_like(inputs, dtype=torch.float16, device='cuda:0')
+                masks = torch.ones((batch, 1, qlen, klen), dtype=torch.bool, device='cuda:0')
+                softmax_results = generic_scaled_masked_softmax_cuda.forward(inputs, masks, scale_t[0].item())
+                back_grad = generic_scaled_masked_softmax_cuda.backward(backward, softmax_results, scale_t[0].item())
+
+                inputs.requires_grad = True
+                softmax_results_torch = forward_torch_softmax(inputs, masks, scale_t[0].item())
+                softmax_results_torch.backward(backward)
+                error = (back_grad - inputs.grad).abs().max()
+                assert error < 1e-3
```
crcrpar, line 390: ditto, `self.assertEqual`

tests/L0/run_transformer/test_fused_softmax.py
```diff
@@ -216,5 +224,171 @@ def test_autocast_fused_upper_triangle_mask_softmax(self):
...
+                2048,
+                4096,
+                7234,
+                8192,
+                10232,
+                4128,
+            ]:
+                inputs = torch.normal(0, 2, (batch, attn, qlen, klen), dtype=torch.float16, device='cuda:0')
+                masks = torch.randint(0, 2, (batch, 1, qlen, klen), dtype=torch.bool, device='cuda:0')
+                softmax_results = generic_scaled_masked_softmax_cuda.forward(inputs, masks, scale_t[0].item())
```
crcrpar, line 263: out of curiosity, why do you have a scale in Tensor and then extract and pass a Python scalar from `scale_t` to this function?

tests/L0/run_transformer/test_fused_softmax.py
```diff
@@ -216,5 +224,171 @@ def test_autocast_fused_upper_triangle_mask_softmax(self):
...
+                8192,
+                10232,
+                4128,
+            ]:
+                inputs = torch.normal(0, 2, (batch, attn, qlen, klen), dtype=torch.float16, device='cuda:0')
+                masks = torch.ones((batch, 1, qlen, klen), dtype=torch.bool, device='cuda:0')
+                softmax_results = generic_scaled_masked_softmax_cuda.forward(inputs, masks, scale_t[0].item())
+                softmax_results_torch = forward_torch_softmax(inputs, masks, scale_t[0].item())
+                error = (softmax_results_torch - softmax_results).abs().max()
+                assert error < 2e-3
```
crcrpar, line 347: ditto, `self.assertEqual` could be simpler.

tests/L0/run_transformer/test_fused_softmax.py
crcrpar, line 239: could you define these variables in `setUp` or define these as class variables?
if they are identical among the test methods of this class

tests/L0/run_transformer/test_fused_softmax.py
yidong72: Thanks. I change ti to use `self.assertEqual`

tests/L0/run_transformer/test_fused_softmax.py
yidong72: Good idea. I used `itertools`

tests/L0/run_transformer/test_fused_softmax.py
yidong72: fixed

tests/L0/run_transformer/test_fused_softmax.py
yidong72: fixed

tests/L0/run_transformer/test_fused_softmax.py
yidong72: yeah. passing a python scalar works. changed.

tests/L0/run_transformer/test_fused_softmax.py
yidong72: fixed

tests/L0/run_transformer/test_fused_softmax.py
yidong72: I used the setup

yidong72:
Thanks for the review. I cleaned up the unittest. Please take a look.

tests/L0/run_transformer/test_fused_softmax.py
```diff
@@ -216,5 +224,69 @@ def test_autocast_fused_upper_triangle_mask_softmax(self):
                 actual.backward(g0)
                 expected.backward(g1)
 
+
+class TestGenericFusedSoftmaxKernel(common_utils.TestCase):
+
+    def _setup_qk(self):
+        qlen = [1, 2, 1234, 2322, 2348]
+        klen = [1, 2, 3, 4, 5, 8, 10, 11, 13, 128, 256, 1200, 1234, 2048, 3123, 4096, 4128, 7234, 8192, 10232]
```
eqy, line 232: Is `10232` necessary here? Looks like it OOMs on GPUs even with 48GiB memory.

yidong72: It works on a 80G A100. 
It is not necessary though. Just to test some case with super long sequence length. I can remove it.

eqy: Sure, it should also be possible to separate the very large cases and gate them based on available device memory e.g.,
https://github.com/pytorch/pytorch/blob/f50a248a5eacb9a9aa475a9e610486aea136e4f5/torch/testing/_internal/common_device_type.py#L914

yidong72: Thanks for sharing.  I think current test cases are enough to cover this new function.

yidong72:
@crcrpar @eqy , are you waiting something from me for this PR?

tests/L0/run_transformer/test_fused_softmax.py
```diff
@@ -216,5 +224,69 @@ def test_autocast_fused_upper_triangle_mask_softmax(self):
...
+        return itertools.product(qlen, klen)
+
+    def _setup_batch_attn_scalar(self):
+        batch = 2
+        attn = 16
+        scale_t = 1.0
+        return batch, attn, scale_t
+
+    def test_forward(self):
+        import generic_scaled_masked_softmax_cuda
```
crcrpar, line 242: out of curiosity, why do you call the extension's forward and backward directly instead of `generic_scaled_masked_softmax` defined in `fused_softmax.py`?

crcrpar (commented):
just fyi; we could reduce `__syncthreads`, ref: https://github.com/NVIDIA/apex/pull/1444#issuecomment-1204880669

tests/L0/run_transformer/test_fused_softmax.py
yidong72: Either way works. As a unit test, I am focusing testing the low lever kennel implementation.

yidong72:
> just fyi; we could reduce `__syncthreads`, ref: [#1444 (comment)](https://github.com/NVIDIA/apex/pull/1444#issuecomment-1204880669)

Yes. I am aware of a lot of "__syncthreads" which has some performance impact.  Some of them are added for numerical instability issues. Without it, sometimes different calls produce different numbers. I want to make sure it is absolute safe to use. I guess the best performance is not the goal here. But it is nice to have the acceleration for arbitrary query and key lengths, which are useful for inference and maybe ASR projects.

crcrpar (approved):

PR #1441, opened 2022-08-02
Fix an issue where `child_name` can be None and make the overall ASP fail

mfuntowicz:
When attempting to sparsify a transformers model, it appears for some reason `child_name` can be `None` and thus `fx_graph.get(None)` returns `None` and make the overall process crash.

This PR attempts to hotfix this but do no investigate why the FX graph generates such `None` children.

I'm testing the output of the `fx_graph.get(...)` to be non-null rather than `child_name` being null because the first one will cover the second (`fx_graph.get(None) -> None`).

### DIFF
--- a/apex/contrib/sparsity/permutation_lib.py
+++ b/apex/contrib/sparsity/permutation_lib.py
@@ -661,7 +661,11 @@ class Permutation:
             while has_visit_children_num < len(node_children):
                 for child_name in node_children:
                     if child_name != 'output':    # 'output' node has no 'module_type'
-                        child_module_type = fx_graph.get(child_name).get('module_type')
+                        child = fx_graph.get(child_name)
+                        if child is None:
+                            has_visit_children_num += 1
+
+                        child_module_type = child.get('module_type')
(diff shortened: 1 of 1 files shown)


ptrblck:
CC @jpool-nv

ChongyuNVIDIA:
Thanks for report such issue and the fix. Can you also upload the transformer model structure to help us understand the failing case?

mfuntowicz:
@ChongyuNVIDIA sorry for the delay, please find below an example repro:

Please note you need transformers main branch in order to have FX support for BLOOM.

```python
from apex.contrib.sparsity import ASP
from transformers import BloomForCausalLM, BloomTokenizerFast
from torch.optim import AdamW

if __name__ == '__main__':
    model = BloomForCausalLM.from_pretrained("bigscience/bloom-350m")
    tokenizer = BloomTokenizerFast.from_pretrained("bigscience/bloom-350m")
    optimizer = AdamW(model.parameters(), amsgrad=True)

    ASP.prune_trained_model(model, optimizer)
```

PR #1444, opened 2022-08-03
support for full mask in fused softmax kernel + avoiding inplace operation in bwd pass

kvareddy:
This adds the following features to the softmax fused kernel

1. Full mask support
2. Avoiding in place backward operation as suggested in the Pytorch documentation

Background: The fused kernel outputs uniform distribution When the mask is all 1 (all values are masked). This led to wrong semantics during the backward pass. We implicitly obtain the mask using fwd softmax output (softmax_output == 0) during backward computation. This assumption does not hold true for the corner case of the mask being all 1. This change outputs 0 whenever the mask is all 1. Conceptually, it makes sense to output 0 if you are not attending to any of the keys. We now match PyTorch (non-fused flow) bwd pass with this change for the full mask case.

Fixes: https://github.com/NVIDIA/apex/issues/1430, https://github.com/NVIDIA/apex/issues/1390

co-authored by:  Vijay Korthikanti <vkorthikanti@nvidia.com> (ADLR team)

### DIFF
--- a/csrc/megatron/scaled_masked_softmax.h
+++ b/csrc/megatron/scaled_masked_softmax.h
@@ -182,6 +182,13 @@ __global__ void scaled_masked_softmax_warp_forward(
     }
     warp_reduce<acc_t, WARP_BATCH, WARP_SIZE, Max>(max_value);
 
+    // compute scale value to account for full mask
+    acc_t scale_value[WARP_BATCH];
+    #pragma unroll
+    for (int i = 0;  i < WARP_BATCH;  ++i) {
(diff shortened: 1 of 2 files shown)


crcrpar:
thanks for the PR

as referring to the same pair of issues, could you help us review #1440?

kvareddy:
#1440 changes look good. The only concern I have is the presence of too many __syncthreads calls in the kernel. It might slow down the kernel in some cases.

Can you please approve this MR. I already pushed these changes to Megatron Repo.

crcrpar (approved):

PR #1446, opened 2022-08-04
Store distributed Adam state_dict in uint8 tensors

timmoon10:
Checkpointing failed with large models since pickle doesn't support >4 GB byetarrays by default (requires pickle protocol >=4). pickle can handle large PyTorch tensors without problems though. With this change, I've been able to successfully checkpoint GPT-3 175B.

I've also fixed a device mismatch error that I was experiencing because `'cuda' != 'cuda:{device_id}'`. To get around this, I've forced the optimizer device to be in `'cuda:{device_id}'` format.

### DIFF
--- a/apex/contrib/optimizers/distributed_fused_adam.py
+++ b/apex/contrib/optimizers/distributed_fused_adam.py
@@ -223,12 +223,14 @@ class DistributedFusedAdam(torch.optim.Optimizer):
                 f'(dtype={dtype}, '
                 f'grad_sync_dtype={grad_sync_dtype}, '
                 f'param_sync_dtype={param_sync_dtype}))')
-        if device != 'cuda':
-            raise RuntimeError('DistributedFusedAdam only supports GPU')
+        device = torch.device(device)
+        if (device.type != 'cuda'
+            or device.index not in (None, torch.cuda.current_device())):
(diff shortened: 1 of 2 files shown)


apex/contrib/optimizers/distributed_fused_adam.py
```diff
@@ -223,12 +223,14 @@ def __init__(self,
                 f'(dtype={dtype}, '
                 f'grad_sync_dtype={grad_sync_dtype}, '
                 f'param_sync_dtype={param_sync_dtype}))')
-        if device != 'cuda':
-            raise RuntimeError('DistributedFusedAdam only supports GPU')
+        device = torch.device(device)
+        if (device.type != 'cuda'
+            or device.index not in (None, torch.cuda.current_device())):
+            raise RuntimeError(f'DistributedFusedAdam is only supported on current GPU')
```
crcrpar, line 229: qq: would it be eqiuvalent to check all the params are a device tensor?

timmoon10: We can actually handle params with any device or dtype, since we copy the values to the optimizer's internal tensors before calling any kernels. The real problem is that when I load a checkpoint, the device is `'cuda:{device_id}'` instead of `'cuda'`.

crcrpar (approved):

PR #1455, opened 2022-08-14
.cu files should not include torch/extension.h

lostmsu:
See https://github.com/facebookresearch/pytorch3d/issues/1127#issuecomment-1071206320

### DIFF
--- a/csrc/megatron/fused_weight_gradient_dense_16bit_prec_cuda.cu
+++ b/csrc/megatron/fused_weight_gradient_dense_16bit_prec_cuda.cu
@@ -3,7 +3,6 @@
 #include <cstdlib>
 #include <cstring>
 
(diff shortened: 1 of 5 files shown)


lostmsu:
I tried adding `libraries=['cublas']` argument, but it did not seem to affect linker flags at all.

crcrpar:
Could you confirm that with this change we can build `--cuda_ext` extension on Linux?

ksmaze:
Could you confirm if upgrading to use cuda 11.7 resolves the issue? It should be resolved in cuda based on comments:
https://github.com/pytorch/pytorch/issues/69460#issuecomment-1278305417

NiallHornFX:
Likewise, does anyone know if this is resolved with CUDA 11.7 ?

PR #1462, opened 2022-08-19
[ASP] Allocate the auxiliary tensors (e.g., mask) in the same device as the model itself

wangshangsam:
This PR fixes ASP such that, when ASP needs to allocate any auxiliary tensors (e.g., masks, `patterns`, etc.), those tensors will be allocated in the same device such as model. Two motivations for this PR:
- `prune_trained_model()` could be called before or after `model.cuda()`.
  - If `prune_trained_model()` is called before `model.cuda()`, those auxiliary tensors are allocated on the GPUs, while the model itself is not, and PyTorch does not permit any op to use tensors on different devices.
- One might bind multiple GPUs to the same progress, where part of the model is on one GPU and the other part on a different GPU.
- Or one might not have GPU at all.


wangshangsam:
@ChongyuNVIDIA let me know if those changes look good to you. Thanks!

PR #1463, opened 2022-08-23
Optimize distributed Adam kernels and implementation

timmoon10:
While mucking around with the kernels for `DistributedFusedAdam`, I've taken the opportunity to make internal improvements and add new features:

- Refactor kernels so they do not require GPU synchronization.
- Add option to directly use values from params instead of storing a copy of params as optimizer state. I've disabled this by default since NeMo Megatron expects that the optimizer operates on FP32 main params instead of BF16 model params. However, I'd like to enable this option by default in the future since a typical user will use the same dtype for their model and optimizer, so it's an easy memory saving.
- If optimizer is FP32 and params are BF16, add option to just store 16-bit param remainders and reconstruct FP32 params out of BF16 params. This is intended to reduce memory usage in NeMo Megatron.
- Add option to choose between Adam and AdamW, matching the behavior of `FusedAdam`.
- Avoid calling collectives in side streams. I noticed significant memory overheads when running with the 22.08 PyTorch container, and I think it's because PyTorch's NCCL collectives allocate some memory internally.

Running GPT-3 5B on a Selene node (4-way data parallelism, 2-way pipeline parallelism, local batch size 2, no micro-batching):

   | Optimizer                | Data types |Step time | Loss at step 50 | Allocated memory | Reserved memory |
   |-----------------------|---|--------|-----------------|--------------|-----------------|
   |  `FusedAdam`         | BF16 params, FP32 main params | 1.44 sec  |            6.53 | 60.4 GB     | 60.9 GB        |
   | `DistributedFusedAdam` | BF16 params, FP32 main params      | 1.37 sec  |            6.54 | 38.2 GB     | 39.8 GB        |
   | `DistributedFusedAdam` | BF16 params, 16-bit param remainders | 1.36 sec  |            6.53 | 36.8 GB     | 37.4 GB        |

### DIFF
--- a/apex/contrib/csrc/optimizers/multi_tensor_distopt_adam_kernel.cu
+++ b/apex/contrib/csrc/optimizers/multi_tensor_distopt_adam_kernel.cu
@@ -28,6 +28,12 @@ __device__ __forceinline__ void load_store(
   ((LT*)dst)[dst_offset] = ((const LT*)src)[src_offset];
 }
 
+// (1-t)*x + t*y
+__device__ __forceinline__ float lerp(float t, float x, float y) {
+  // See https://developer.nvidia.com/blog/lerp-faster-cuda/
+  return fma(t, y, fma(-t, x, x));
+}
+
 typedef enum{
   ADAM_MODE_0   =0, // L2 regularization mode
   ADAM_MODE_1   =1  // Decoupled weight decay mode(AdamW)
@@ -46,7 +52,7 @@ struct DistAdamFunctor
     T m[ILP],
     T v[ILP],
     const GRAD_T g[ILP],
-    const T grad_scale,
+    const float grad_scale,
     const float beta1,
     const float beta2,
     const float beta1_correction,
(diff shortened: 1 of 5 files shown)


apex/contrib/csrc/optimizers/multi_tensor_distopt_adam_kernel.cu
```diff
@@ -14,151 +14,305 @@
...
+    const float eps,
+    const float lr,
+    adamMode_t mode,
+    const float weight_decay) {
+    if (mode == ADAM_MODE_0) { // L2
+#pragma unroll
+      for (int ii = 0; ii < ILP; ii++) {
+        T scaled_grad = (g[ii] * grad_scale) + (weight_decay * p[ii]);
+        m[ii] = beta1*m[ii] + (1-beta1)*scaled_grad;
+        v[ii] = beta2*v[ii] + (1-beta2)*scaled_grad*scaled_grad;
```
crcrpar, line 63: https://developer.nvidia.com/blog/lerp-faster-cuda/ can be applied here

apex/contrib/csrc/optimizers/multi_tensor_distopt_adam_kernel.cu
```diff
@@ -14,151 +14,305 @@
...
+        T denom = sqrtf(next_v_unbiased) + eps;
+        T update = next_m_unbiased / denom;
+        p[ii] -= lr * update;
+      }
+    } else { // weight decay
+#pragma unroll
+      for (int ii = 0; ii < ILP; ii++) {
+        T scaled_grad = g[ii] * grad_scale;
+        m[ii] = beta1*m[ii] + (1-beta1)*scaled_grad;
+        v[ii] = beta2*v[ii] + (1-beta2)*scaled_grad*scaled_grad;
```
crcrpar, line 75: ditto, https://developer.nvidia.com/blog/lerp-faster-cuda/ can be applied here

apex/contrib/optimizers/distributed_fused_adam.py
```diff
@@ -1,26 +1,84 @@
...
+    _FOUND_DEPRECATED_FUSED_ADAM = True
+except ImportError:
+    import warnings
+    warnings.warn(
+        'Could not find recommended CUDA kernels when importing '
+        'apex.contrib.optimizers.distributed_fused_adam. '
+        'Apex should be installed with the following options: '
+        '`--cpp_exp`, `--cuda_ext`, '
+        '`--distributed_adam`, `--deprecated_fused_adam`.'
+    )
```
crcrpar, line 27: does this optimizer need `fused_adam_cuda`? I don't think this try-except raises when the extension is not found

timmoon10: `fused_adam_cuda` has a fused copy kernel that we use here:

https://github.com/timmoon10/apex/blob/2d5ebb793ca212773bc0af4b15afaee2cea5e82f/apex/contrib/optimizers/distributed_fused_adam.py#L66

If we can't find that kernel, then we fall back to performing the copies one at a time.

apex/contrib/csrc/optimizers/multi_tensor_distopt_adam_kernel.cu
timmoon10: That's a neat trick, I'll make the change.

apex/contrib/optimizers/distributed_fused_adam.py
crcrpar: I see, nit-picking though, "APEX should be installed with the following option" sounded a little bit misleading to me so how about clarifying this optimizer will be faster with `--deprecated_fused_adam`?

apex/contrib/csrc/optimizers/multi_tensor_distopt_adam_kernel.cu
```diff
@@ -14,151 +14,315 @@
...
+    const float eps,
+    const float lr,
+    adamMode_t mode,
+    const float weight_decay) {
+    if (mode == ADAM_MODE_0) { // L2
+#pragma unroll
+      for (int ii = 0; ii < ILP; ii++) {
+        float scaled_grad = (g[ii] * grad_scale) + (weight_decay * p[ii]);
+        float g2 = scaled_grad * scaled_grad;
+        // next_m = beta1 * m + (1-beta1) * g
```
crcrpar, line 63:
```suggestion
        // [note faster lerp] See https://developer.nvidia.com/blog/lerp-faster-cuda/
```

apex/contrib/csrc/optimizers/multi_tensor_distopt_adam_kernel.cu
```diff
@@ -14,151 +14,315 @@
...
+        float next_m = fma(beta1, m[ii], fma(-beta1, scaled_grad, scaled_grad));
+        // next_v = beta2 * v + (1-beta2) * g^2
```
crcrpar, line 65:
```suggestion
```

apex/contrib/csrc/optimizers/multi_tensor_distopt_adam_kernel.cu
```diff
@@ -14,151 +14,315 @@
...
+        m[ii] = next_m;
+        v[ii] = next_v;
+        p[ii] -= lr * update;
+      }
+    } else { // weight decay
+#pragma unroll
+      for (int ii = 0; ii < ILP; ii++) {
+        float scaled_grad = g[ii] * grad_scale;
+        float g2 = scaled_grad * scaled_grad;
+        // next_m = beta1 * m + (1-beta1) * g
```
crcrpar, line 80:
```suggestion
        // see [note faster lerp]
```

apex/contrib/csrc/optimizers/multi_tensor_distopt_adam_kernel.cu
```diff
@@ -14,151 +14,315 @@
...
+        float next_m = fma(beta1, m[ii], fma(-beta1, scaled_grad, scaled_grad));
+        // next_v = beta2 * v + (1-beta2) * g^2
```
crcrpar, line 82:
```suggestion
```

crcrpar (approved):
cosmetic nit-pickings, otherwise looks nice

crcrpar (approved):

PR #1469, opened 2022-08-30
fixes spurious illegal memaccess

yjk21:


crcrpar (approved):
thank you so much for your swift fix for batch size 56.
I confirmed the patch with nightly, NGC 22.08, and pytorch/pytorch:1.12.0-cuda11.3-cudnn8-devel on A100

PR #1475, opened 2022-09-07
Support overlapped grad sync with Megatron pipeline parallelism

timmoon10:
This PR adds functionality so that the distributed Adam optimizer can overlap grad reduce-scatters with backward compute in the first pipeline stage. Async grad reductions are disabled in the other pipeline stages since it slows down the backward pass, so the reduction should be done externally while waiting for the first stage to finish. Note that this is not compatible with `DistributedDataParallel` since I am not aware of an option to manually trigger a reduction after using `no_sync`.

I've also done some refactoring of distributed Adam to support NeMo-Megatron integration, mostly so that I can selectively disable async grad reductions for model-parallel operations:

- Each bucket independently keeps track of which grads it has received
- Exposed function that creates callback functions for launching async grad reductions, which I override in NeMo
- Perform collective communication for checkpointing in main stream in order to reduce memory pool overheads
- Change default value for grad norm function from `[]` to `None`. This allows us to pass in empty iterators

### DIFF
--- a/apex/contrib/optimizers/distributed_fused_adam.py
+++ b/apex/contrib/optimizers/distributed_fused_adam.py
@@ -260,6 +260,8 @@ class DistributedFusedAdam(torch.optim.Optimizer):
             self.status = DistributedFusedAdam.GradientStatus.READY
             # Request object for asynchronous communication
             self.sync_request = None
+            # Params that have generated grads
+            self.grads_generated = set()
 
         def sync_wait(self):
             """Wait for asynchronous communication to finish"""
@@ -420,7 +422,6 @@ class DistributedFusedAdam(torch.optim.Optimizer):
 
         # Objects for gradient synchronization
         self._grads_buckets = collections.defaultdict(self.GradientBucket)
-        self._grads_generated = set()
         self._pipeline_streams = [torch.cuda.Stream() for _ in range(self.pipeline_size)]
 
         # Scale by factor before optimizer step. Used for grad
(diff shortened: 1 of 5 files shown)


apex/transformer/pipeline_parallel/schedules/common.py
```diff
@@ -396,3 +397,11 @@ def backward_step(
 
     # timers("backward-compute").stop()
     return input_tensor_grad[0] if unwrap_input_tensor_grad else input_tensor_grad
+
+
+@contextmanager
```
crcrpar, line 402: I understand this is moved from "fwd_bwd_no_pipelining.py", but would you know whether or not [`contextlib.nullcontext`](https://docs.python.org/3/library/contextlib.html#contextlib.nullcontext) suffices?

apex/transformer/pipeline_parallel/schedules/fwd_bwd_pipelining_without_interleaving.py
```diff
@@ -456,6 +475,9 @@ def forward_backward_pipelining_without_interleaving(
     _logger.info("Cooldown phase")
     if not forward_only:
         for i in range(num_warmup_microbatches):
+            if i == num_warmup_microbatches-1 and rank == 0:
+                context.__exit__(None, None, None)
+                context = None
```
crcrpar, line 480: Question: why do we need to set `context` to `None`?

apex/transformer/pipeline_parallel/schedules/fwd_bwd_pipelining_without_interleaving.py
```diff
@@ -239,6 +240,7 @@ def forward_backward_pipelining_without_interleaving(
     deallocate_pipeline_outputs: bool = False,
     async_comm: bool = False,
     sequence_parallel_enabled: bool = False,
+    custom_sync_context_handler = None,
```
crcrpar, line 243: Can you add some test case(s) with a certain context to https://github.com/NVIDIA/apex/blob/master/tests/L0/run_transformer/test_pipeline_parallel_fwd_bwd.py#L283 once this gets "ready"?

apex/contrib/optimizers/distributed_fused_adam.py
```diff
@@ -928,11 +933,10 @@ def grad_sync(self):
                     self._grad_copy(param)
                     self._try_start_bucket_grad_sync(
                         params=[param],
-                        ignore_last_bucket=False,
```
crcrpar, line 931: I'd prefer to keep this one as is even though the default value is False

crcrpar (commented):
knowing this would be too early...

apex/transformer/pipeline_parallel/schedules/common.py
timmoon10: I like that better. I'll make that change.

apex/transformer/pipeline_parallel/schedules/fwd_bwd_pipelining_without_interleaving.py
timmoon10: I want to check at the end if the context has been exited, but I could also do it by setting a bool. Do you think that's easier to understand?

crcrpar: hmm, I cannot describe why but this value assignment doesn't feel nice

tests/L0/run_transformer/test_pipeline_parallel_fwd_bwd.py
```diff
@@ -442,6 +443,109 @@ def test_pipelining_without_interleaving_sequence_parallel_encoder_or_decoder(se
...
+        )
+        torch.cuda.synchronize()
+
+        # Check context behavior
+        assert has_entered_context, 'Has not entered custom sync context'
+        assert has_exited_context, 'Has not exited custom sync context'
+        assert no_grad_at_context_exit == parallel_state.is_pipeline_first_stage(), \
+            'Expected to exit custom sync context '\
+            'before backward pass in first pipeline stage ' \
+            'and after backward pass in other pipeline stages'
```
crcrpar, line 544: [nit-picking] Could you use `unittest.TestCase.assertTrue` and `unittest.TestCase.assertEqual` instead?
```suggestion
        self.assertTrue(has_entered_context, 'Has not entered custom sync context')
        self.assertTrue(has_exited_context, 'Has not exited custom sync context')
        self.assertEqual(no_grad_at_context_exit, parallel_state.is_pipeline_first_stage(),
            'Expected to exit custom sync context '\
            'before backward pass in first pipeline stage ' \
            'and after backward pass in other pipeline stages')
```
note: I haven't validate this suggestion.

crcrpar (commented):
SGTM

crcrpar (approved):
thank you :)

PR #1483, opened 2022-09-13
Skip test if related contrib modules are not found

crcrpar:
Also, add `__init__.py` so that we can run contrib tests with `python -m unittest/xmlrunner discover /path/to/apex/contrib/test` (see https://docs.python.org/3/library/unittest.html#test-discovery, https://github.com/xmlrunner/unittest-xml-reporting#command-line). You can specify junit xml report output directory with `-o` option (available only when `xmlrunner` (can be installed with `pip install unittest-xml-reporting`) is selected).

Signed-off-by: Masaki Kozuki <mkozuki@nvidia.com>


crcrpar:
looks okay, need to add some nccl version guard to `--nccl_p2p`, https://github.com/NVIDIA/apex/blob/ce9df7dfe90d5e8c1a46015fbcf6b13e4df7adbf/setup.py#L721

PR #1490, opened 2022-09-20
update fused_dense to use torch.autocast instead of apex amp

Fuzzkatt:
Remove last references to apex amp and replace with torch.autocast, since apex amp is being deprecated according to this issue: https://nvbugswb.nvidia.com/NvBugs5/SWBug.aspx?bugid=3781071&cmtNo=

### DIFF
--- a/apex/contrib/test/fused_dense/test_fused_dense.py
+++ b/apex/contrib/test/fused_dense/test_fused_dense.py
@@ -1,6 +1,7 @@
 import unittest
-import torch
 import os
+
+import torch
 from torch.testing._internal import common_utils
 from torch.testing._internal.common_device_type import instantiate_device_type_tests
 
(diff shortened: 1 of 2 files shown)


apex/contrib/test/fused_dense/test_fused_dense.py
```diff
@@ -1,6 +1,8 @@
 import unittest
-
 import torch
+import os
+from torch.testing._internal import common_utils
+from torch.testing._internal.common_device_type import instantiate_device_type_tests
```
crcrpar, line 5: could you follow https://peps.python.org/pep-0008/#imports regarding the order?

is `unittest` still used?

apex/contrib/test/fused_dense/test_fused_dense.py
```diff
@@ -10,39 +12,50 @@
...
-        self.tst_inputs = self.ref_inputs.clone().detach().requires_grad_(True)
-        self.dense = fused_dense.FusedDense(1024, 3072)
-        self.dense.half()
-        self.dense.cuda()
+        seq_length = 512
+        sequences = 3
+        hidden_dim = 1024
 
+        ref_inputs = torch.randn(sequences*seq_length, hidden_dim,
+                                 dtype=dtype, device=torch.device("cuda")).int().to(dtype=dtype).requires_grad_(True)
```
crcrpar, line 27: why casting to int then cast back to dtype?

Fuzzkatt: Good point, the original code did this but it doesn't make sense to me either, I'll get rid of the extra casts.

apex/contrib/test/fused_dense/test_fused_dense.py
Fuzzkatt: Will fix the order. `unittest` is still used to skip the test if for some reason apex can't import fused_dense, should I leave that there?

PR #1496, opened 2022-09-29
Replace usage of internal with public APIs

kumpera:
PyTorch c10d recently introduced some public functions that provides the same functionality as accessing those internals do.

 _get_global_rank is deprecated.

c10d._pg_group_ranks is slated for removal as part of some internals refactoring. See https://github.com/pytorch/pytorch/pull/84153 - it will be relanded soon.


crcrpar:
Thank you @kumpera

PR #1508, opened 2022-10-12
Empty CUDA cache in transformer unit tests

Aidyn-A:
Fixes an OOM issue when running `TORCH_ALLOW_TF32_CUBLAS_OVERRIDE=0 python /opt/pytorch/apex/tests/L0/run_test.py --include run_transformer` on A100x8.

Because `run_test.py` runs all test as a unified suite of all the tests in `run_transformer`. 
https://github.com/NVIDIA/apex/blob/806f9b0f6e321e105029a45f5cbab95f1d6d6434/tests/L0/run_test.py#L82
PyTorch uses the same CUDA cache in most of these tests and the amount of cached memory increases.

But why OOM failure is this happening on A100x8 (cards with the largest memory)? Because `test_fused_softmax` allocates a large amount of memory if a GPU has more than 40GB available:
https://github.com/NVIDIA/apex/blob/806f9b0f6e321e105029a45f5cbab95f1d6d6434/tests/L0/run_transformer/test_fused_softmax.py#L241-L244

The solution is to clear CUDA cache in `TearDown ` to prevent OOM.

### DIFF
--- a/apex/transformer/testing/distributed_test_base.py
+++ b/apex/transformer/testing/distributed_test_base.py
@@ -35,6 +35,7 @@ class DistributedTestBase(common_distributed.MultiProcessTestCase):
         self._spawn_processes()
 
     def tearDown(self) -> None:
+        torch.cuda.empty_cache()
         super().tearDown()
 
     @property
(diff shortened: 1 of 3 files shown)


crcrpar (approved):
Nice

PR #1512, opened 2022-10-14
Fix bug with grad clipping and distributed Adam

timmoon10:
I've gotten incorrect results using distributed Adam to train GPT-3 at FP16 because of a bug with gradient clipping and gradient scaling. In particular, there's an incorrect assumption that gradient clipping is applied before the gradients are unscaled, which results in extremely small gradients. The fix simply requires gradient clipping to handle the case where [`DistributedFusedAdam._grad_scale`](https://github.com/NVIDIA/apex/blob/2c3b8f9a0808e4c931a84def706885ff1d33e5f7/apex/contrib/optimizers/distributed_fused_adam.py#L436) != 1.0. Note that [`torch.cuda.amp.GradScaler.unscale_`](https://pytorch.org/docs/stable/amp.html#torch.cuda.amp.GradScaler.unscale_) does not natively support the distributed optimizer, so I have to unscale by directly manipulating `DistributedFusedAdam._grad_scale`.

While I was touching the code, I've also adopted some recommendations from @erhoo82:
- Avoid unnecessarily zeroing out GPU buffers.
- Use communication functions ([`reduce_scatter_tensor`](https://github.com/pytorch/pytorch/blob/f451e824f39516f503c2bdfd785d254b447b9557/torch/distributed/distributed_c10d.py#L2742) and [`all_gather_into_tensor`](https://github.com/pytorch/pytorch/blob/f451e824f39516f503c2bdfd785d254b447b9557/torch/distributed/distributed_c10d.py#L2274)) that do not allocate internal memory.

### DIFF
--- a/apex/contrib/optimizers/distributed_fused_adam.py
+++ b/apex/contrib/optimizers/distributed_fused_adam.py
@@ -12,12 +12,22 @@ from apex.multi_tensor_apply import multi_tensor_applier
 import amp_C
 import distributed_adam_cuda
 
-# Fallback to private function if using older PyTorch version
+# Fallback to private functions if using older PyTorch version
 try:
     from torch.distributed.distributed_c10d import get_global_rank
 except ImportError:
     from torch.distributed.distributed_c10d import _get_global_rank
     get_global_rank = _get_global_rank
+try:
+    from torch.distributed.distributed_c10d import reduce_scatter_tensor
+except ImportError:
+    from torch.distributed.distributed_c10d import _reduce_scatter_base
+    reduce_scatter_tensor = _reduce_scatter_base
+try:
+    from torch.distributed.distributed_c10d import all_gather_into_tensor
+except ImportError:
+    from torch.distributed.distributed_c10d import _all_gather_base
+    all_gather_into_tensor = _all_gather_base
 
 _FOUND_DEPRECATED_FUSED_ADAM = False
 try:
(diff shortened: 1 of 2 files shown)


crcrpar:
> Note that [`torch.cuda.amp.GradScaler.unscale_`](https://pytorch.org/docs/stable/amp.html#torch.cuda.amp.GradScaler.unscale_) does not natively support the distributed optimizer, so I have to unscale by directly manipulating `DistributedFusedAdam._grad_scale`.

qq: would https://github.com/NVIDIA/apex/blob/master/apex/transformer/amp/grad_scaler.py be useful?

timmoon10:
Interesting, incorporating the distributed optimizer into that is a good idea. I'm using the [NeMo grad scaler](https://github.com/NVIDIA/NeMo/blob/18940b3b32cff290cf70d4a251b0e2f7b08e1525/nemo/collections/nlp/parts/nlp_overrides.py#L395), which seems to be an extension of the Apex version.

timmoon10:
I've updated the Apex grad scaler to match the changes I've made in the NeMo grad scaler.

apex/contrib/optimizers/distributed_fused_adam.py
```diff
@@ -1050,9 +1055,22 @@ def clip_grad_norm(self, max_norm, parameters=None, norm_type=2.0):
...
+    def unscale_grads(self, inv_scale, *args):
+        """Custom unscale function for use by AMP gradient scaler
+
+        Overflow checking is deferred to optimization step.
+
+        Arguments:
+            inv_scale (torch.Tensor): factor to multiply gradients
+
+        """
+        self._grad_scale *= inv_scale.view([])
```
crcrpar, line 1071: qq: why `.view`?

timmoon10: The grad scaler initializes its scale factor as a 1D tensor of size 1 instead of as a scalar: https://github.com/pytorch/pytorch/blob/6e1764d806bc45e7c15c79f7e5f1a0bafb76ec73/torch/cuda/amp/grad_scaler.py#L146

crcrpar (approved):

PR #1514, opened 2022-10-20
Support overlapped grad sync with Megatron interleaved pipeline parallelism

timmoon10:
This PR makes similar changes as https://github.com/NVIDIA/apex/pull/1475. In particular, it provides an option to pass a custom `no_sync` context into [`apex.transformer.pipeline_parallel.schedules.fwd_bwd_pipelining_with_interleaving._forward_backward_pipelining_with_interleaving`](https://github.com/NVIDIA/apex/blob/2c3b8f9a0808e4c931a84def706885ff1d33e5f7/apex/transformer/pipeline_parallel/schedules/fwd_bwd_pipelining_with_interleaving.py#L26). In the first pipeline stage, gradient reductions will be overlapped with the final backward pass. In other pipeline stages, the caller is expected to manually perform gradient reductions (in the bubble overhead).

### DIFF
--- a/apex/transformer/pipeline_parallel/schedules/fwd_bwd_pipelining_with_interleaving.py
+++ b/apex/transformer/pipeline_parallel/schedules/fwd_bwd_pipelining_with_interleaving.py
@@ -1,5 +1,5 @@
 import contextlib
-from typing import List, Union, Optional, Sequence
+from typing import Any, List, Optional, Sequence, Union
 import warnings
 
 import torch
@@ -37,7 +37,7 @@ def _forward_backward_pipelining_with_interleaving(
     deallocate_pipeline_outputs: bool = False,
     async_comm: bool = False,
     sequence_parallel_enabled: bool = False,
-    custom_sync_context_handler = None,
+    custom_sync_context_handler: Optional[Any] = None,
     sync_batch_comm: bool = True,
     **kwargs,
 ) -> List[Union[torch.Tensor, Sequence[torch.Tensor]]]:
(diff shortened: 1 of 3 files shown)


apex/transformer/pipeline_parallel/schedules/fwd_bwd_pipelining_with_interleaving.py
```diff
@@ -71,11 +73,20 @@ def _forward_backward_pipelining_with_interleaving(
...
             When :obj:`True`, the sequence length on each tensor model parallel rank is updated
             to :math:`original\_sequence\_length / tensor\_model\_parallel\_world\_size`.
+        custom_sync_context_handler: Context manager to disable
+            asynchronous gradient reductions. In the first pipeline
+            stage, asynchronous gradient reductions are enabled in the
+            backward pass of the last microbatch. In other pipeline
+            stages, asynchronous gradient reductions are disabled and
+            gradients must be manually synchronized by the caller (in
+            practice the runtime is covered up by the bubble
+            overhead).
```
crcrpar, line 83: I should've asked in #1475 though how about clarifying the default behavior here? e.g. just mentioning that the default behavior is to have no effects.

timmoon10:
After discussion with @erhoo82, the right behavior is to enable overlapped grad syncs during the last backward pass. This is in all pipeline stages, not just the first. Since the interleaved schedule involves putting multiple model chunks in each pipeline stage, we must be capable of entering and exiting the `no_sync` context multiple times (e.g. to enable overlapped grad sync in the final backward pass of one model chunk, and then disable for forward/backward passes in another model chunk). I now see the expected overlap pattern when running GPT-3 5B on a DGX A100 node.

crcrpar:
@timmoon10 could you resolve conflicts?

crcrpar (approved):

PR #1513, opened 2022-10-20
Use all_gather_into_tensor and reduce_scatter_tensor

Aidyn-A:
This PR updates usage of old collective operations in favor of new `all_gather_into_tensor` and `reduce_scatter_tensor`. 
Updated `apex.contrib.optimizers.distributed_fused_lamb`
Updated `apex.transformer.utils`, `apex.transformer.tensor_parallel.layers` and `apex.transformer.tensor_parallel.mappings`

PR #1517, opened 2022-10-24
Remove weight gradients if model is frozen for linear layer

JimmyZhang12:
Changes the backward pass of the linear layer so that weight gradients for frozen models are not calculated

* Checks if weights have `requires_grad`
* if `requires_grad=False`, only calculate and return the data gradients
* Improves training step time for prompt tuning gpt3 by about 20%

### DIFF
--- a/apex/transformer/tensor_parallel/layers.py
+++ b/apex/transformer/tensor_parallel/layers.py
@@ -290,12 +290,19 @@ class LinearWithGradAccumulationAndAsyncCommunication(torch.autograd.Function):
         sequence_parallel_enabled: bool,
         use_16bit_in_wgrad_accum_fusion: bool = False,
     ):
-        ctx.save_for_backward(input, weight)
         ctx.use_bias = bias is not None
         ctx.gradient_accumulation_fusion = gradient_accumulation_fusion
         ctx.async_grad_allreduce = async_grad_allreduce
         ctx.sequence_parallel_enabled = sequence_parallel_enabled
         ctx.use_16bit_in_wgrad_accum_fusion = use_16bit_in_wgrad_accum_fusion
+        ctx.compute_weight_gradient = weight.requires_grad
+
+        if ctx.compute_weight_gradient:
+            ctx.save_for_backward(input, weight)
+        else:
+            ctx.save_for_backward(weight)
+            ctx.use_bias = False
+
 
         if ctx.sequence_parallel_enabled:
             world_size = get_tensor_model_parallel_world_size()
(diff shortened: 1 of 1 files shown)


apex/transformer/tensor_parallel/layers.py
```diff
@@ -313,45 +320,76 @@ def forward(
...
+                    sub_grad_input,
+                    grad_input,
+                    group=get_tensor_model_parallel_group(),
+                    async_op=True
+                )
+                handle.wait()
+                return sub_grad_input, None, None, None, None, None, None
+            if ctx.async_grad_allreduce:
+                handle.wait()
+            return grad_input, None, None, None, None, None, None
```
crcrpar, line 385: QQ: is this `grad_input` dependent on L.355 of `grad_input = grad_output.matmul(weight)` and L.362 when async_grad_allreduce? (I'm trying to understand why the view op on grad_output and total_input are moved

apex/transformer/tensor_parallel/layers.py
```diff
@@ -283,12 +283,19 @@ def forward(
         sequence_parallel_enabled: bool,
         use_16bit_in_wgrad_accum_fusion: bool = False,
     ):
-        ctx.save_for_backward(input, weight)
         ctx.use_bias = bias is not None
```
crcrpar, line 286: nit-picking

```suggestion
        ctx.use_bias = bias is not None and weight.requires_grad
```

apex/transformer/tensor_parallel/layers.py
```diff
@@ -283,12 +283,19 @@ def forward(
...
         ctx.async_grad_allreduce = async_grad_allreduce
         ctx.sequence_parallel_enabled = sequence_parallel_enabled
         ctx.use_16bit_in_wgrad_accum_fusion = use_16bit_in_wgrad_accum_fusion
+        ctx.compute_weight_gradient = weight.requires_grad
+
+        if ctx.compute_weight_gradient:
+            ctx.save_for_backward(input, weight)
+        else:
+            ctx.save_for_backward(weight)
+            ctx.use_bias = False
```
crcrpar, line 297:
```suggestion
```

apex/transformer/tensor_parallel/layers.py
```diff
@@ -313,45 +320,76 @@ def forward(
...
-                group=get_tensor_model_parallel_group(),
-                async_op=True,
-            )
-            total_input = all_gather_buffer
-        else:
-            total_input = input
         grad_input = grad_output.matmul(weight)
 
-        if ctx.sequence_parallel_enabled:
+        if handle is not None:
```
crcrpar, line 357: Is this condition equivalent to `ctx.sequence_parallel_enabled and ctx.compute_weight_gradient`?

apex/transformer/tensor_parallel/layers.py
JimmyZhang12: The view on grad_output and total input are only used for calculating the weight gradient so if `weight.requires_grad=False` then they are not needed so I moved them after the return on L.385 so that views would not be executed. Maybe it would be better to keep the views as they are but wrap them in an `if ctx.compute_weight_grad`

apex/transformer/tensor_parallel/layers.py
JimmyZhang12: Yes it is equivalent

apex/transformer/tensor_parallel/layers.py
crcrpar: understood. when I first saw "immediately return", I wondered if we could put the no weight grads logic before with weight grads.

crcrpar (approved):

crcrpar:
@JimmyZhang12 could you resolve conflicts, following https://github.com/NVIDIA/apex/pull/1513?

JimmyZhang12:
Conflicts should be resolved now.

crcrpar (approved):

PR #1518, opened 2022-10-25
add ada lovelace architecture (compute_89/sm_89) support

gregjhogan:
I only added support where `compute_86`/`sm_86` currently is supported (this excluded `fast_layer_norm` and `fmha`)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -85,7 +85,7 @@ if not torch.cuda.is_available():
     if os.environ.get("TORCH_CUDA_ARCH_LIST", None) is None and CUDA_HOME is not None:
         _, bare_metal_version = get_cuda_bare_metal_version(CUDA_HOME)
         if bare_metal_version >= Version("11.8"):
-            os.environ["TORCH_CUDA_ARCH_LIST"] = "6.0;6.1;6.2;7.0;7.5;8.0;8.6;9.0"
+            os.environ["TORCH_CUDA_ARCH_LIST"] = "6.0;6.1;6.2;7.0;7.5;8.0;8.6;8.9;9.0"
         elif bare_metal_version >= Version("11.1"):
             os.environ["TORCH_CUDA_ARCH_LIST"] = "6.0;6.1;6.2;7.0;7.5;8.0;8.6"
         elif bare_metal_version == Version("11.0"):
@@ -334,7 +334,7 @@ if "--cuda_ext" in sys.argv:
             cc_flag.append("arch=compute_86,code=sm_86")
         if bare_metal_version >= Version("11.8"):
             cc_flag.append("-gencode")
-            cc_flag.append("arch=compute_90,code=sm_90")
+            cc_flag.append("arch=compute_89,code=sm_89,compute_90,code=sm_90")
 
         ext_modules.append(
             CUDAExtension(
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -334,7 +334,7 @@ def check_cudnn_version_and_warn(global_option: str, required_cudnn_version: int
             cc_flag.append("arch=compute_86,code=sm_86")
         if bare_metal_version >= Version("11.8"):
             cc_flag.append("-gencode")
-            cc_flag.append("arch=compute_90,code=sm_90")
+            cc_flag.append("arch=compute_89,code=sm_89,compute_90,code=sm_90")
```
crcrpar, line 337:
```suggestion
            cc_flag.append("arch=compute_89,code=sm_89")
            cc_flag.append("-gencode")
            cc_flag.append("arch=compute_90,code=sm_90")
```

setup.py
```diff
@@ -617,7 +617,7 @@ def check_cudnn_version_and_warn(global_option: str, required_cudnn_version: int
         cc_flag.append("arch=compute_86,code=sm_86")
     if bare_metal_version >= Version("11.8"):
         cc_flag.append("-gencode")
-        cc_flag.append("arch=compute_90,code=sm_90")
+        cc_flag.append("arch=compute_89,code=sm_89,compute_90,code=sm_90")
```
crcrpar, line 620:
```suggestion
        cc_flag.append("arch=compute_89,code=sm_89")
        cc_flag.append("-gencode")
        cc_flag.append("arch=compute_90,code=sm_90")
```

crcrpar (commented):
thank you for your pull request.
lgtm except suggestions

gregjhogan:
@crcrpar I applied your suggestions, thanks!

cliffwoolley:
sm_80 and sm_86 code both run on sm_89 devices.  Building for every single compute capability version isn't necessary.

gregjhogan:
> sm_80 and sm_86 code both run on sm_89 devices. Building for every single compute capability version isn't necessary.

Is this why some modules are only built for sm_80 and sm_90 (sm_86 is missing), and sm_86 could be removed everywhere in `setup.py`?

I wasn't sure how I could know if I was missing out on some optimizations for sm_89 if it wasn't included in the target arch list.

cliffwoolley:
Good questions:

> Is this why some modules are only built for sm_80 and sm_90 (sm_86 is missing), and therefore sm_86 could be removed everywhere in setup.py?

We have a few places in Apex where kernels do end up being _specifically_ for say sm_80, but not sm_86/sm_89.

In terms of instruction set, code built for sm_80 will run on sm_86 or sm_89 GPUs like I said.  But there are a couple of opt-in areas such as the amount of shared memory used per CUDA thread block where an app can use some features of the "bigger" GPU (in this case sm_80) at the expense of binary compatibility with the other GPUs of the family.  Apex has a few spots that do actually use such opt-in features, and so for those cases, we intentionally don't build for the other GPUs in the family.

> I wasn't sure how I could know if I was missing out on some optimizations for sm_89 if it wasn't included in the target arch list.

Arguably we could have skipped building for sm_86 for the cases that _are_ binary compatible instead of building for both sm_80 and sm_86.  The real best way to decide whether to build for each sm version separately is to actually benchmark them; it's easy to just add every version across the board, but on the other hand it's most often unnecessary and doesn't actually buy much/any benefit besides just bigger binaries.

gregjhogan:
It is easy enough to set `TORCH_CUDA_ARCH_LIST` while building if you care about size, but the same argument could be used both for and against having `8.9` included in `TORCH_CUDA_ARCH_LIST` by default.

cliffwoolley:
Fair point.  (People do often report issues on our binary sizes, which is why I'm sensitive to that.)

PR #1539, opened 2022-11-15
Update megatron fused softmax follow megatron-lm

yaoyu-33:
We updated megatron fused softmax in the following aspects:
1. We updated the limit of sequence length from 2048 to 4096 according to https://github.com/NVIDIA/Megatron-LM/blob/main/megatron/model/fused_softmax.py#L171
2. We also enabled `mask=None` support in scaled_masked_softmax according to https://github.com/NVIDIA/Megatron-LM/blob/main/megatron/model/fused_softmax.py#L84

### DIFF
--- a/csrc/megatron/scaled_softmax.cpp
+++ b/csrc/megatron/scaled_softmax.cpp
@@ -1,0 +1,75 @@
+/* coding=utf-8
+ * Copyright (c) 2021, NVIDIA CORPORATION.  All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <cuda_fp16.h>
+#include <torch/extension.h>
+#include <vector>
+
+namespace multihead_attn {
+namespace fused_softmax {
+namespace scaled_softmax {
+
+torch::Tensor fwd_cuda(
+    torch::Tensor const& input, 
+    float scale_factor);
+
+torch::Tensor bwd_cuda(
+    torch::Tensor const& output_grads, 
+    torch::Tensor const& softmax_results,
+    float scale_factor);
+
+torch::Tensor fwd(
+    torch::Tensor const& input,
+    float scale_factor) {
+  AT_ASSERTM(input.dim() == 4, "expected 4D tensor");
+  AT_ASSERTM((input.scalar_type() == at::ScalarType::Half) ||
+	     (input.scalar_type() == at::ScalarType::BFloat16), 
+      "Only fp16 and bf16 are supported");
+
+  return fwd_cuda(input, scale_factor);
+}
+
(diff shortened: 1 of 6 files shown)


csrc/megatron/scaled_softmax.cpp
```diff
@@ -0,0 +1,75 @@
+/* coding=utf-8
+ * Copyright (c) 2021, NVIDIA CORPORATION.  All rights reserved.
```
crcrpar, line 2:
```suggestion
 * Copyright (c) 2022, NVIDIA CORPORATION.  All rights reserved.
```

crcrpar (commented):
how is the custom C++/CUDA extension of `scaled_softmax_cuda` built?
usually `setup.py` needs to be updated.
also, could you update https://github.com/NVIDIA/apex/blob/master/tests/L0/run_transformer/test_fused_softmax.py as well?

yaoyu-33:
@crcrpar comments addressed

tests/L0/run_transformer/test_fused_softmax.py
```diff
@@ -139,6 +139,84 @@ def test_autocast_fused_scale_mask_softmax(self):
...
+
+            expected = fused_fn(attention_scores_0, mask)
+            actual = torch_fn(attention_scores_1, mask)
+            self.assertEqual(actual, expected, msg=msg)
+
+            g0 = torch.rand_like(actual)
+            with torch.no_grad():
+                g1 = g0.clone()
+            expected.backward(g0)
+            actual.backward(g1)
```
crcrpar, line 188: feels like these lines are based off of some existing cases but how about either checking grads or remove these lines?

yaoyu-33: added grad check

yaoyu-33: I think the purpose in other tests was just to test backward can work correctly

tests/L0/run_transformer/test_fused_softmax.py
```diff
@@ -139,6 +139,86 @@ def test_autocast_fused_scale_mask_softmax(self):
...
+                actual = fused_fn(attention_scores_0, mask)
+                self.assertEqual(actual.dtype, dtype, msg=msg)
+            self.assertEqual(actual, expected, msg=msg)
+
+            g0 = torch.rand_like(actual)
+            with torch.no_grad():
+                g1 = g0.clone()
+            expected.backward(g0)
+            actual.backward(g1)
+            self.assertEqual(attention_scores_0.grad, attention_scores_1.grad, msg=msg)
```
crcrpar, line 220: I'm seeing a dtype mismatch here

tests/L0/run_transformer/test_fused_softmax.py
```diff
@@ -139,6 +139,86 @@ def test_autocast_fused_scale_mask_softmax(self):
...
+            expected = fused_fn(attention_scores_0, mask)
+            actual = torch_fn(attention_scores_1, mask)
+            self.assertEqual(actual, expected, msg=msg)
+
+            g0 = torch.rand_like(actual)
+            with torch.no_grad():
+                g1 = g0.clone()
+            expected.backward(g0)
+            actual.backward(g1)
+            self.assertEqual(attention_scores_0.grad, attention_scores_1.grad, msg=msg)
```
crcrpar, line 189: and this check complains that 1% elements have numerical mismatch

crcrpar (commented):
generally okay but could you maybe tweak tests a bit?

tests/L0/run_transformer/test_fused_softmax.py
yaoyu-33: lol, do we know why? I am not familiar with this

PR #1543, opened 2022-11-18
Optimize peer memory halo exchange kernel for latency

timmoon10:
This PR optimizes the peer memory halo exchange to minimize latency:
- Each CUDA thread performs its communication independently, which allows us to avoid cooperative group syncs. With enough threads, we avoid memory fences entirely. The downside is that we can only achieve 50% bandwidth since we pack data and flags into the same communication packets.
- Use an alternating double-buffer scheme, which allows push-only communication.
- The kernel does multiple rounds of communication in the cases where the number of SMs or the peer buffers are too small. This is important since the optimal peer memory requirements have increased by 4x, or more if the tensors are not nicely NHWC.  The only strict requirement is that the peer buffer sizes are aligned to 4096B.

Running on two DGX-1 V100s:
   | Version | Data type |   Kernel time |
   |-----------|-------|--------------------|
   | Baseline  | FP32  | 19.9 us |
   | Baseline  | FP16  |  17.2 us |
   | Optimized | FP32  | 15.2 us |
   | Optimized | FP16  | 11.4 us |

<details> <summary> Details </summary>

- I performed the halo exchange required for 3x3 convolution on a [1,336,336,64] NHWC tensor.
- To isolate the kernel time from the communication time, I delayed GPU 0 so it would always launch the kernel well after GPU 1.

</details>

Running the Mask R-CNN benchmark:

   | GPUs | Step time |
   |------|---------------|
   |    1 |       37.7 ms |
   |    2 |        34.8 ms |
   |    4 |        33.6 ms |

~The unit tests for the [halo exchanger](https://github.com/NVIDIA/apex/blob/master/apex/contrib/test/peer_memory/test_peer_halo_exchange_module.py) and the [bottleneck module](https://github.com/NVIDIA/apex/blob/master/apex/contrib/test/peer_memory/test_peer_halo_exchange_module.py) all pass for me on DGX-1 V100s.~ Pinging @thorjohnsen.

### DIFF
--- a/apex/contrib/bottleneck/halo_exchangers.py
+++ b/apex/contrib/bottleneck/halo_exchangers.py
@@ -88,14 +88,24 @@ class HaloExchangerSendRecv(HaloExchanger):
             inc.left_right_halo_exchange_inplace(self.handle, self.left_rank, self.right_rank, left_output_halo, right_output_halo, left_input_halo, right_input_halo)
 
 class HaloExchangerPeer(HaloExchanger):
-    def __init__(self, ranks, rank_in_group, peer_pool, explicit_nhwc, numSM=1):
+    def __init__(self, ranks, rank_in_group, peer_pool, explicit_nhwc, numSM=0):
         super(HaloExchangerPeer, self).__init__(ranks, rank_in_group)
         self.diagnostics = False
         self.explicit_nhwc = explicit_nhwc
         self.numSM = numSM
         self.peer_pool = peer_pool
-        self.signals = peer_pool.allocate_peer_tensors([2,4], torch.int32, False, False)
-        self.signals[self.rank_in_group].zero_()
+
+    def _allocate_peer_tensor(self, halo):
+
+        # Compute size in bytes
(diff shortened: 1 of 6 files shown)


timmoon10:
I've modified the bottleneck layer test to allocate more peer memory (since our peer memory requirement has grown 4x), and now it all passes for me.

I've also run some experiments on a DGX A100. Measuring the halo exchange kernel on two A100s:

   | Version | Data type |          Kernel time |
   |-----------|-------|--------------------|
   | Baseline  | FP32  | 28.3 us |
   | Baseline  | FP16  |   24.3 us |
   | Optimized | FP32  | 13.7 us |
   | Optimized | FP16  |  12.8 us |

Running the Mask R-CNN benchmark:

   | GPUs | Step time |
   |------|-----------|
   |    1 | 14.0 ms   |
   |    2 | 15.0 ms   |
   |    4 | 14.8 ms   |
   |    8 | 16.4 ms   |

It seems that with a sufficiently beefy GPU, we don't quite get strong scaling yet.

PR #1544, opened 2022-11-29
Coalesce reduce-scatters in distributed Adam

timmoon10:
When running the distributed Adam optimizer with `CUDA_DEVICE_MAX_CONNECTIONS=1`, @erhoo82 has observed poor overlapping between backward compute and gradient reduce-scatters. This happens when multiple reduce-scatters are launched at the same time and, from what I can tell, the serialization happens during inter-stream synchronization (namely between the calling stream and the internal NCCL communication stream).

The serialization vanishes if I [coalesce multiple reduce-scatters into a single operation](https://docs.nvidia.com/deeplearning/nccl/user-guide/docs/usage/groups.html#aggregated-operations-2-2-and-later), I think because it reduces the number of stream synchronizations. It should also have the benefit of reducing communication time by reducing latency costs.

The dist Adam unit tests pass for me and I don't observe a difference in learning behavior when GPT-2 124M for 20 steps.

### DIFF
--- a/apex/contrib/optimizers/distributed_fused_adam.py
+++ b/apex/contrib/optimizers/distributed_fused_adam.py
@@ -852,13 +852,8 @@ class DistributedFusedAdam(torch.optim.Optimizer):
         else:
             reduce_op = torch.distributed.ReduceOp.SUM
 
-        # Side stream for communication
-        main_stream = torch.cuda.current_stream()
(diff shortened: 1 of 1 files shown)


timmoon10:
Superseded by https://github.com/NVIDIA/apex/pull/1551.

PR #1557, opened 2022-12-12
Layernorm one_plus support in fast_layer_norm

anmolgupt:
In this implementation, gamma tensor will be initialized to 0 (instead of 1) and the +1 is numerically handled. It is enabled by setting use_1p=True (default: False)


apex/contrib/layer_norm/layer_norm.py
```diff
@@ -7,47 +7,56 @@
...
         gamma = gamma.contiguous()
         beta = beta.contiguous()
         hidden_size = gamma.numel()
         xmat = x.view((-1, hidden_size))
-        ymat, mu, rsigma = fast_layer_norm.ln_fwd(xmat, gamma, beta, epsilon)
+        if use_1p:
+            one_plus = 1.0
+        else:
+            one_plus = 0.0
+        ymat, mu, rsigma = fast_layer_norm.ln_fwd(xmat, gamma, beta, epsilon, one_plus)
```
mikolajblaz, line 20: I would suggest this for simplicity (assumes use_1p is a bool)
```suggestion
        ymat, mu, rsigma = fast_layer_norm.ln_fwd(xmat, gamma, beta, epsilon, float(use_1p))
```
(similarly in other places)

apex/contrib/test/layer_norm/test_fast_layer_norm.py
```diff
@@ -232,29 +239,38 @@ def test_all_configs(self):
...
-                self.assertAll(_test_impl(256, 2, h, bf16, bf16))
-                self.assertAll(_test_impl(256, 2, h, fp32, bf16))
+        use_1p = [
+            False,
+            True
+        ]
+        for is_1p in use_1p:
+            for h in hidden_sizes:
+                with self.subTest(f"hidden_size={h}, use_1p={is_1p}"):
+                    self.assertAll(_test_impl(256, 2, h, fp32, fp32, is_1p))
```
mikolajblaz, line 249: Is there an existing test (already in Apex) that checks that custom LayerNorm (like FastLayerNormFN) behaves exactly the same as e.g. PyT LayerNorm? It would be nice to double check that our modification with adding +1 is mathematically correct. It seems that `_test_impl` uses `backward_` to check correctness, which is using the same +1 logic as the kernels.

PR #1559, opened 2022-12-20
Distributed optimizer overlaps forward compute and param all-gather

timmoon10:
This PR adds experimental support for the distributed Adam optimizer to overlap its parameter all-gather with the forward compute. It does this by monkey-patching the PyTorch `Parameter` class's `__getattribute__` function. The optimizer step sets a flag in each of the parameters, and the first time a parameter is accessed in the forward pass it makes sure that its all-gather has finished.

This approach is delicate and prone to errors, hangs, or performance regressions (e.g. embeddings require special care since they pass the parameters into the C++ backend without activating our monkey-patched functions). In the long-term, it would be better to transition toward PyTorch's FSDP, which is more intrusive into the model but also more robust. That said, this functionality is disabled by default and I see some modest speedups (~2%) training GPT-3 175B on Selene.

This branch contains the work in https://github.com/NVIDIA/apex/pull/1551.


apex/contrib/optimizers/distributed_fused_adam.py
```diff
@@ -494,16 +582,51 @@ def __init__(self,
...
         self._register_post_backward_hooks()
+        # Attach hooks for param synchronization
+        if self.overlap_param_sync:
+            self._register_pre_forward_hooks()
+
+    def _broadcast_params(self):
+        """Broadcast parameter values from root rank"""
+        sync_requests = []
+        process_group = self.process_group
+        process_group._start_coalescing()
```
crcrpar, line 600: this part could require a similar change as https://github.com/NVIDIA/apex/pull/1551/commits/d42f78118047cb680da0aec77ee1e23124e631ef

crcrpar (changes requested):

apex/contrib/optimizers/distributed_fused_adam.py
timmoon10: Done

apex/contrib/optimizers/distributed_fused_adam.py
```diff
@@ -786,7 +1058,7 @@ def _init_param_state(
                     main_param_fragment = bucket.params_shard[shard_start:shard_end]
                     main_param_fragment.copy_(model_param_fragment)
 
-    def zero_grad(self, set_to_none=True):
+    def zero_grad(self, set_to_none=False):
```
crcrpar, line 1061: is this change to make use of the contiguous buffer for params & grads?

apex/contrib/optimizers/distributed_fused_adam.py
```diff
@@ -862,6 +1150,64 @@ def _grad_copy(self, param):
...
+        for param in params:
+            if 'fragments' in self.state[param]:
+                fragments.extend(
+                    fragment
+                    for fragment in self.state[param]['fragments']
+                    if fragment.bucket_id in self._params_buckets
+                )
+
+        # Make sure all needed buckets have been synchronized
+        buckets = collections.OrderedDict()
```
crcrpar, line 1169: fwiw recent Python keeps the insertion order as per https://mail.python.org/pipermail/python-dev/2017-December/151283.html

crcrpar (approved):

apex/contrib/optimizers/distributed_fused_adam.py
timmoon10: Yep, `set_to_none=False` avoids an extra memory allocation and memory copy if `contiguous_grad_buffer=True` and should have no impact if `contiguous_grad_buffer=False`. It also matches [`torch.optim.Adam`](https://pytorch.org/docs/stable/generated/torch.optim.Adam.html#torch.optim.Adam.zero_grad).

apex/contrib/optimizers/distributed_fused_adam.py
timmoon10: Huh, that's pretty neat. I'm a little scared to rely on it though.

PR #1560, opened 2022-12-22
Capturable Adam optimizer

gilshm:
This PR adds a "capturable" argument to enable the Adam optimizer to be captured by CUDA graphs. This enhancement also supports non-static LR strategies.

### DIFF
--- a/apex/contrib/test/optimizers/test_adam.py
+++ b/apex/contrib/test/optimizers/test_adam.py
@@ -1,0 +1,188 @@
+import copy
+import math
+import random
+import unittest
+
+import torch
+import torch.nn.functional as F
+from torch import nn
+
+try:
+    import apex
+except ImportError as e:
+    HAS_APEX = False
+else:
+    HAS_APEX = True
+
+
+class Model(torch.nn.Module):
+    def __init__(self):
+        super(Model, self).__init__()
+        self.conv1 = nn.Conv2d(1, 6, 5)
+        self.relu1 = nn.ReLU()
+        self.pool1 = nn.MaxPool2d(2)
+        self.conv2 = nn.Conv2d(6, 16, 5)
+        self.relu2 = nn.ReLU()
+        self.pool2 = nn.MaxPool2d(2)
+        self.fc1 = nn.Linear(256, 120)
+        self.relu3 = nn.ReLU()
+        self.fc2 = nn.Linear(120, 84)
+        self.relu4 = nn.ReLU()
+        self.fc3 = nn.Linear(84, 10)
+        self.relu5 = nn.ReLU()
+
+    def forward(self, x):
+        y = self.conv1(x)
+        y = self.relu1(y)
+        y = self.pool1(y)
+        y = self.conv2(y)
+        y = self.relu2(y)
+        y = self.pool2(y)
+        y = y.reshape(y.shape[0], -1)
+        y = self.fc1(y)
+        y = self.relu3(y)
+        y = self.fc2(y)
+        y = self.relu4(y)
+        y = self.fc3(y)
+        y = self.relu5(y)
+        return y 
+
+
+@unittest.skipIf(not HAS_APEX, "`apex` is not found.")
+class AdamTest(unittest.TestCase):
+    def setUp(self, seed=0):
+        super().setUp()
+        torch.manual_seed(seed)
+
+        self.model = Model().cuda()
+        self.model_ = Model().cuda()
+        self.model_.load_state_dict(copy.deepcopy(self.model.state_dict()))
+
+        self.lr = 0.00001
+        params = [p for p in self.model.parameters() if p.requires_grad]
+        self.optimizer = torch.optim.Adam(params, lr=self.lr)
+
+    def testGradScaler(self):
+        params_ = [p for p in self.model_.parameters() if p.requires_grad]
+        self.optimizer_ = apex.optimizers.FusedAdam(params_, lr=self.lr, capturable=False)
+        scaler = torch.cuda.amp.GradScaler(enabled=True)
+        scaler_ = torch.cuda.amp.GradScaler(enabled=True)
+
+        for i in range(100):
+            x = torch.rand([32, 1, 28, 28]).cuda().to(memory_format=torch.channels_last)
+            x_ = x.clone()
+            gt = torch.rand([32, 10]).cuda()
+            gt_ = gt.clone()
+
+            # Reference
+            with torch.cuda.amp.autocast(enabled=True):
+                y = self.model(x)
+                loss = ((gt - y) ** 2).mean()
+
+            scaler.scale(loss).backward()
+            scaler.step(self.optimizer)
+            scaler.update()
+            
+            # DUT
+            with torch.cuda.amp.autocast(enabled=True):
+                y = self.model_(x)
+                loss_ = ((gt_ - y) ** 2).mean()
+
+            scaler_.scale(loss_).backward()
+            scaler_.step(self.optimizer_)
+            scaler_.update()
+
+            for module in zip(self.model.modules(), self.model_.modules()):
+                m = module[0]
+                m_ = module[1]
+                if isinstance(m, nn.Conv2d) or isinstance(m_, nn.Linear):
+                    self.assertTrue(torch.allclose(m.weight, m_.weight, atol=1e-3, rtol=1e-3, equal_nan=True))
+
+            # Init for next iteration
+            self.optimizer.zero_grad()
+            self.optimizer_.zero_grad()
+
+            self.model_.load_state_dict(copy.deepcopy(self.model.state_dict()))
+    
+    def testGradScalerCapturable(self):
+        params_ = [p for p in self.model_.parameters() if p.requires_grad]
+        self.optimizer_ = apex.optimizers.FusedAdam(params_, lr=self.lr, capturable=True)
+        scaler = torch.cuda.amp.GradScaler(enabled=True)
+        scaler_ = torch.cuda.amp.GradScaler(enabled=True)
+
+        for i in range(100):
+            x = torch.rand([32, 1, 28, 28]).cuda().to(memory_format=torch.channels_last)
+            x_ = x.clone()
+            gt = torch.rand([32, 10]).cuda()
+            gt_ = gt.clone()
+
+            # Reference
+            with torch.cuda.amp.autocast(enabled=True):
+                y = self.model(x)
+                loss = ((gt - y) ** 2).mean()
+
+            scaler.scale(loss).backward()
+            scaler.step(self.optimizer)
+            scaler.update()
+            
+            # DUT
+            with torch.cuda.amp.autocast(enabled=True):
+                y = self.model_(x)
+                loss_ = ((gt_ - y) ** 2).mean()
+
+            scaler_.scale(loss_).backward()
+            scaler_.step(self.optimizer_)
+            scaler_.update()
+
+            for module in zip(self.model.modules(), self.model_.modules()):
+                m = module[0]
+                m_ = module[1]
+                if isinstance(m, nn.Conv2d) or isinstance(m_, nn.Linear):
+                    self.assertTrue(torch.allclose(m.weight, m_.weight, atol=1e-3, rtol=1e-3, equal_nan=True))
+
+            # Init for next iteration
+            self.optimizer.zero_grad()
+            self.optimizer_.zero_grad()
+
+            self.model_.load_state_dict(copy.deepcopy(self.model.state_dict()))
+    
+    def testNative(self):
+        params_ = [p for p in self.model_.parameters() if p.requires_grad]
+        self.optimizer_ = apex.optimizers.FusedAdam(params_, lr=self.lr, capturable=False)
+
+        for i in range(100):
+            x = torch.rand([32, 1, 28, 28]).cuda().to(memory_format=torch.channels_last)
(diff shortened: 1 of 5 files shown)


csrc/multi_tensor_adam.cu
```diff
@@ -126,6 +126,118 @@ struct AdamFunctor
...
+    TensorListMetadata<4>& tl,
+    const float beta1,
+    const float beta2,
+    const int* step,
+    const int bias_correction,
+    const float epsilon,
+    const float* lr,
+    adamMode_t mode,
+    const float decay,
+    const float* inv_scale)
```
crcrpar, line 144: Is this the argument which differentiates this functor from the existing `AdamFunctor`?

apex/optimizers/fused_adam.py
```diff
@@ -150,44 +166,104 @@ def step(self, closure=None, grads=None, output_params=None, scale=None, grad_no
...
+                        group['lr'],
+                        beta1,
+                        beta2,
+                        group['eps'],
+                        group['step'],
+                        self.adam_w_mode,
+                        bias_correction,
+                        group['weight_decay'],
+                    )
+                if(len(g_32) > 0):
```
crcrpar, line 256: nit-picking
```suggestion
                if len(g_32) > 0:
```

or 
```suggestion
                if g_32:
```

apex/optimizers/fused_adam.py
```diff
@@ -150,44 +166,104 @@ def step(self, closure=None, grads=None, output_params=None, scale=None, grad_no
...
+					 beta1,
+					 beta2,
+					 group['eps'],
+					 group['step'],
+					 self.adam_w_mode,
+					 bias_correction,
+					 group['weight_decay'],
+					 inv_scale)
+            else:
+                if(len(g_16) > 0):
```
crcrpar, line 230: nit-picking
```suggestion
                if len(g_16) > 0:
```
or
```suggestion
                if g_16:
```

apex/optimizers/fused_adam.py
```diff
@@ -150,44 +166,104 @@ def step(self, closure=None, grads=None, output_params=None, scale=None, grad_no
...
+					 group['lr'],
+					 beta1,
+					 beta2,
+					 group['eps'],
+					 group['step'],
+					 self.adam_w_mode,
+					 bias_correction,
+					 group['weight_decay'],
+					 inv_scale)
+            else:
```
crcrpar, line 229: the indentation doesn't look consistent. could you apply some formatter?

apex/optimizers/fused_adam.py
```diff
@@ -62,20 +62,35 @@ class FusedAdam(torch.optim.Optimizer):
...
-                 weight_decay=0., amsgrad=False, set_grad_none=True):
+                 weight_decay=0., amsgrad=False, set_grad_none=True,
+                 capturable=False):
 
         if amsgrad:
             raise RuntimeError('FusedAdam does not support the AMSGrad variant.')
+        # If the optimizer is capturable then LR should be a tensor (on GPU)
+        lr = torch.tensor(lr, dtype=torch.float32) if capturable else lr
         defaults = dict(lr=lr, bias_correction=bias_correction,
                         betas=betas, eps=eps, weight_decay=weight_decay)
```
crcrpar, line 73: I'm wondering if `load_state_dict` works as is when `capturable=True`

apex/contrib/test/optimizers/test_adam.py
```diff
@@ -0,0 +1,188 @@
+import copy
```
crcrpar, line 1: could you put this file into tests/L0/run_optimizers?

apex/contrib/test/optimizers/test_adam.py
```diff
@@ -0,0 +1,188 @@
...
+        self.model_ = Model().cuda()
+        self.model_.load_state_dict(copy.deepcopy(self.model.state_dict()))
+
+        self.lr = 0.00001
+        params = [p for p in self.model.parameters() if p.requires_grad]
+        self.optimizer = torch.optim.Adam(params, lr=self.lr)
+
+    def testGradScaler(self):
+        params_ = [p for p in self.model_.parameters() if p.requires_grad]
+        self.optimizer_ = apex.optimizers.FusedAdam(params_, lr=self.lr, capturable=False)
```
crcrpar, line 67: [trivial question] why making the optimizer an attribute of this class?

crcrpar, line 67: what do you think about testing both AdamW and Adam?

apex/contrib/test/optimizers/test_adam.py
```diff
@@ -0,0 +1,188 @@
...
+
+            # Init for next iteration
+            self.optimizer.zero_grad()
+            self.optimizer_.zero_grad()
+
+            self.model_.load_state_dict(copy.deepcopy(self.model.state_dict()))
+    
+    def testGradScalerCapturable(self):
+        params_ = [p for p in self.model_.parameters() if p.requires_grad]
+        self.optimizer_ = apex.optimizers.FusedAdam(params_, lr=self.lr, capturable=True)
```
crcrpar, line 109: ditto

apex/contrib/test/optimizers/test_adam.py
```diff
@@ -0,0 +1,188 @@
...
+
+            # Init for next iteration
+            self.optimizer.zero_grad()
+            self.optimizer_.zero_grad()
+
+            self.model_.load_state_dict(copy.deepcopy(self.model.state_dict()))
+    
+    def testNative(self):
+        params_ = [p for p in self.model_.parameters() if p.requires_grad]
+        self.optimizer_ = apex.optimizers.FusedAdam(params_, lr=self.lr, capturable=False)
```
crcrpar, line 151: ditto

apex/contrib/test/optimizers/test_adam.py
```diff
@@ -0,0 +1,188 @@
...
+
+            scaler_.scale(loss_).backward()
+            scaler_.step(self.optimizer_)
+            scaler_.update()
+
+            for module in zip(self.model.modules(), self.model_.modules()):
+                m = module[0]
+                m_ = module[1]
+                if isinstance(m, nn.Conv2d) or isinstance(m_, nn.Linear):
+                    self.assertTrue(torch.allclose(m.weight, m_.weight, atol=1e-3, rtol=1e-3, equal_nan=True))
```
crcrpar, line 99: nit-picking
```suggestion
                    torch.testing.assert_close(m.weight, m_.weight, atol=1e-3, rtol=1e-3, equal_nan=True)
```

apex/contrib/test/optimizers/test_adam.py
```diff
@@ -0,0 +1,188 @@
...
```
crcrpar, line 141: ditto
nit-picking
```suggestion
                    torch.testing.assert_close(m.weight, m_.weight, atol=1e-3, rtol=1e-3, equal_nan=True)
```

apex/contrib/test/optimizers/test_adam.py
```diff
@@ -0,0 +1,188 @@
...
+            loss_ = ((gt_ - y) ** 2).mean()
+
+            loss_.backward()
+            self.optimizer_.step()
+
+            for module in zip(self.model.modules(), self.model_.modules()):
+                m = module[0]
+                m_ = module[1]
+                if isinstance(m, nn.Conv2d) or isinstance(m_, nn.Linear):
+                    self.assertTrue(torch.allclose(m.weight, m_.weight, atol=1e-3, rtol=1e-3, equal_nan=True))
```
crcrpar, line 177: ditto, nit-picking
```suggestion
                    torch.testing.assert_close(m.weight, m_.weight, atol=1e-3, rtol=1e-3, equal_nan=True)
```

apex/optimizers/fused_adam.py
crcrpar: ah, there exists the check

csrc/multi_tensor_adam.cu
```diff
@@ -169,3 +281,42 @@ void multi_tensor_adam_cuda(
...
+      AdamCapturableFunctor<scalar_t_0>(),
+      beta1,
+      beta2,
+      step.DATA_PTR<int>(),
+      bias_correction,
+      epsilon,
+      lr.DATA_PTR<float>(),
+      (adamMode_t) mode,
+      weight_decay,
+      inv_scale.data_ptr<float>()); )
```
crcrpar, line 317: IMHO, it's okay to not use `DATA_PTR`

csrc/multi_tensor_adam.cu
gilshm: This functor is different from the original with the following:
1) const int* step - pointer, since it is now being used inside the kernel
2) const int bias_correction - same, see [here](https://github.com/NVIDIA/apex/blob/6588f6b03cf60ba65941d236a19dd5d046e7f90a/csrc/multi_tensor_adam.cu#L256-L261)
3) const float* lr - pointer, to support LR scheduling while captured
4) const float* inv_scale - same

apex/contrib/test/optimizers/test_adam.py
gilshm: Done

apex/contrib/test/optimizers/test_adam.py
gilshm: Fixed

gilshm: You are right, it is not really necessary. I modified that so the tested optimizer is _not_ an attribute of the class.

gilshm: I don't mind adding that to the testbench.
Does it only requires me to set `weight_decay` to non-zero value?

apex/contrib/test/optimizers/test_adam.py
gilshm: Fixed

apex/contrib/test/optimizers/test_adam.py
gilshm: Fixed

apex/contrib/test/optimizers/test_adam.py
gilshm: Fixed

apex/contrib/test/optimizers/test_adam.py
gilshm: Fixed

apex/contrib/test/optimizers/test_adam.py
gilshm: Fixed

apex/optimizers/fused_adam.py
gilshm: Fixed

apex/optimizers/fused_adam.py
gilshm: Fixed

apex/optimizers/fused_adam.py
gilshm: Fixed

csrc/multi_tensor_adam.cu
gilshm: The thing is `lr` and `step` are `at::Tensor` (see [here](https://github.com/NVIDIA/apex/blob/930bfb59b343b7dc2dccb0ee9fd29e41a54b9853/csrc/multi_tensor_adam.cu#L285-L297)). Will it still be correct if I will remove `DATA_PTR`?

crcrpar: oh I meant to say it's safe to use `data_ptr` instead like `inv_scale` as per https://github.com/NVIDIA/apex/blob/75f401e088ef88e7c85a57ecf70fb232235f0334/csrc/compat.h#L5-L9 (where `VERSION_GE_1_3` means the available pytorch's version >= 1.3)

gilshm: Got it. I see I also used the lowercase version in line 317...
I modified everything to be lowercase as you suggested.

csrc/multi_tensor_adam.cu
```diff
@@ -126,6 +126,118 @@ struct AdamFunctor
...
+      }
+#pragma unroll
+      for(int ii = 0; ii < ILP; ii++)
+      {
+        int i = i_start + threadIdx.x + ii*blockDim.x;
+        if(i < n && i < chunk_size)
+        {
+          p[i] = r_p[ii];
+          m[i] = r_m[ii];
+          v[i] = r_v[ii];
```
crcrpar, line 234: just to confirm, as this block doesn't seem to store the gradients, wouldn't you need unscaled gradient values outside of `optimizer.step` when GradientScaler is used?

crcrpar (commented):
I got a quick question

csrc/multi_tensor_adam.cu
gilshm: Perhaps I'm missing something... why do I need to store the gradients? as far as I understand, when gradient scaling is used and if overflow is detected, then the current iteration gradients are discarded. Doesn't it work like that?

crcrpar: while non-fused Adam such as `torch.optim.Adam` would unscale gradients (technically `GradScaler` unscales the registered gradients on behalf) in the in-place manner, this optimizer doesn't. So if we're seriously interested in the parity between the native Adam and the fused one here, it would make sense to store the unscaled gradients

gilshm: Fixed and added check to test.

PR #1564, opened 2023-01-17
Instance norm (and batch, layer norm) using NVFuser Python frontend only

jacobhinkle:
This implements a generic norm layer using the NVFuser Python frontend without using the CUDAExtension introduced in the target branch. I translated the pieces that were not exposed in `fd.ops`, namely the `instance_norm` and `instance_norm_backward` functions in `ops/normalization.cpp`.

The code is modified a bit to take a `stat_axes` parameter, which indicates the axes which are used to index the computed statistics. For instance norm, the stats are computed for each combination of instance (batch dim) and channel, so `stat_axes = [NamedAxis.BATCH, NamedAxis.CHANNEL]`. Batch and layer norm correspond to only the channel or batch dimension, respectively. The batch dimension is always reduced when computing running stats, but not the channel dim. I have not tested batch or layer norm using this approach yet as I have so far avoided modifying the test in the target branch at all.

Note that this PR requires using the `devel` branch of PyTorch (for `add_output(t, alias_input=s)`) which has undergone refactoring of nvfuser into `thirdparty/`.


apex/normalization/instance_norm.py
```diff
@@ -0,0 +1,808 @@
+from enum import Enum
+from typing import Any, Dict, List, Optional, Tuple
+
+import torch
+from torch.nn.modules.batchnorm import _NormBase
+
+from nvfuser._C import DataType, Fusion, FusionDefinition, Scalar, Tensor
```
jjsjann123, line 7: Sorry for breaking the news, but @kevinstephano had refactored python API and this part likely needs some patching for that.

jjsjann123: Also, let's now worry about it for now. IIUC, we also need the TOT devel branch for this to work?!

apex/normalization/instance_norm.py
```diff
@@ -0,0 +1,808 @@
...
+
+
+NamedAxis = Enum("NamedAxis", ["BATCH", "CHANNEL"])
+
+
+def torch2datatype(dt: torch.dtype) -> Optional[DataType]:
```
jjsjann123, line 13: Let's expose `aten_to_data_dtype` defined in `third_party/nvfuser/csrc/type.h` instead of wrapping it in python again.

Also, why is there an entry for `bool`, if we are only looking at torch.ScalarType, it should always be a `torch.bool` instead?!

apex/normalization/instance_norm.py
```diff
@@ -0,0 +1,808 @@
...
+    num_stats = 1
+    if NamedAxis.BATCH in stat_axes:
+        stat_dims.append(batch_dim)
+        num_stats *= extent[batch_dim]
+    if NamedAxis.CHANNEL in stat_axes:
+        stat_dims.append(channel_dim)
+        stat_dims_nobatch.append(channel_dim)
+        num_stats *= extent[channel_dim]
+    x_reduction_axes = [ax for ax in range(num_dims) if ax not in stat_dims]
+    num_features = extent.numel() // num_stats
```
jjsjann123, line 116: Looks like all shape specific things are hard coded as compile time constant.

Just want to point it out here that we are not supporting dynamic shapes? I'm wondering if there's a technical reason that we are not doing so.

apex/normalization/instance_norm.py
```diff
@@ -0,0 +1,808 @@
...
+        stat_axes: List[NamedAxis],
+    ) -> torch.Tensor:
+        channels_last = x.is_contiguous(
+            memory_format=torch.channels_last
+        ) or x.is_contiguous(memory_format=torch.channels_last_3d)
+        xorig = x
+        if channels_last:
+            order = [0] + [i for i in range(2, len(x.shape))] + [1]
+            x = x.permute(order)
+        assert x.is_contiguous()
```
jjsjann123, line 404: Nice~ Using `NamedAxis` allows us to basically ignore the semantic meaning of each dimensions when we handle reduction axes with permuted tensor.

apex/normalization/instance_norm.py
```diff
@@ -0,0 +1,808 @@
...
+
+        x_datatype = torch2datatype(x.dtype)
+
+        # The cache key for this Fusion definition. Misses will trigger redefinition and rescheduling.
+        key = (
```
jjsjann123, line 409: if the fusion definition above is shape specific, then input shape here should be part of the cache entry.

tests/L0/run_instance_norm_nvfuser/test_instance_norm_nvfuser.py
```diff
@@ -0,0 +1,102 @@
...
+
+import apex
+from apex.normalization import InstanceNorm3dNVFuser
+
+class TestInstanceNormNVFuser(unittest.TestCase):
+    dtype = torch.float
+    track_running_stats = False
+    channels_last = False
+    affine = False
+    batch_size = 5
```
jjsjann123, line 15: I think the one set shape test here hides the issue I posted above about the shape specific fusion definition vs absence of shape in cache key.

jjsjann123 (commented):
Took a quick look through this.

normalization behavior is always tricky to get right. I put some initial comment on this and will take a better look later.

apex/normalization/instance_norm.py
crcrpar: > IIUC, we also need the TOT devel branch for this to work?!

yes, I'm comfortable waiting for the refactoring pr on pytorch/pytorch

tests/L0/run_instance_norm_nvfuser/test_instance_norm_nvfuser.py
```diff
@@ -0,0 +1,102 @@
...
+            def __init__(self):
+                super(Model, self).__init__()
+                self.norm = InstanceNorm3dNVFuser(4)
+
+            def forward(self, x):
+                x = self.norm(x)
+                x = torch.sum(x, dim=(1, 2, 3, 4))
+                return x
+
+        device = torch.device(f"cuda:1")
```
crcrpar, line 95: nit-picking. no need to be an f-string

```suggestion
        device = torch.device("cuda:1")
```

apex/normalization/instance_norm.py
jacobhinkle: Removed the `bool` entry which as you mentioned has no place there. As for exposing `aten_to_data_dtype` I will do this and probably rename it as it will take pytorch as opposed to aten types.

apex/normalization/instance_norm.py
jacobhinkle: You're right. I've removed the cache anyway now though (see above).

tests/L0/run_instance_norm_nvfuser/test_instance_norm_nvfuser.py
jacobhinkle: Yeah, will add changing shape to the tests.

apex/normalization/instance_norm.py
jacobhinkle: One issue with dynamic shapes was that unlike in C++, I wasn't sure how to refer to tensor axes/extents in the frontend. @kevinstephano maybe I could take a crack at adding that to the frontend?

apex/normalization/instance_norm.py
jacobhinkle: After talking with Kevin, I'm removing the explicit cache altogether. It was based on `Fusion`s which are no longer exposed, and was probably misconceived by me to begin with since the `FusionDefinition` has its own cache mechanism, so the explicit cache here only saves the actual building of the FusionDefinition (stuff in the `with` block). `FusionDefinition` is lightweight so that part is fast anyway.

apex/normalization/instance_norm.py
jacobhinkle: Pushed 716fbc2e419a8ca307016e7ed8fa29995fb82a3e which enables dynamic shapes: sending any sized tensor does not change the number of cached fusions, even without the explicit cache. The tests are updated to check correctness and that the cache doesn't grow when passing different size inputs.

jacobhinkle: For the record: The trick is to use `fd.ops.tensor_sizes` to get a tuple of `Scalar`s describing the extent of a `Tensor`.

tests/L0/run_instance_norm_nvfuser/test_instance_norm_nvfuser.py
jacobhinkle: Added changing shapes to tests in 716fbc2e419a8ca307016e7ed8fa29995fb82a3e

tests/L0/run_instance_norm_nvfuser/test_instance_norm_nvfuser.py
```diff
@@ -0,0 +1,102 @@
...
+        model = Model().to(device)
+
+        x = torch.randn(2, 4, 128, 128, 128, device=device, requires_grad=True)
+        y = torch.randn(2, device=device)
+        pred = model(x)
+        loss = nn.functional.mse_loss(pred, y.float())
+        loss.backward()
```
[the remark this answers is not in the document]
jacobhinkle: Good suggestion. Done in 2c2a8efaa3fdb77c8fe7c2ae12de00887322f3ba

apex/normalization/instance_norm.py
jjsjann123: That's true. Luckily it looks like for various normalization, this seems to work out perfectly fine :clap:

jacobhinkle:
I did some benchmarking on a 3080. I did sweeps over batch, channel, and spatial dimensions, with track_running_stats&affine both false, but with half and full precision and contiguous, strided, and channels_last layouts. Full collection of plots inside a notebook along with the script used to collect benchmarks are in a gist at https://gist.github.com/jacobhinkle/a8051c150c51f0505d97862557290344#file-instancenormbenchmarks-ipynb. Some highlights:
- The version in this PR (apex) is similar or faster than eager mode in most cases, particularly in the forward direction.
- We handle strided and channels_last inputs better than eager mode `InstanceNorm3d`. In some cases this is pretty drastic:
image
It seems to be specific to instance norm, as I've shown eager batchnorm for comparison.
- Eager batchnorm is often much slower than either instance norm:
image
I think eager is slow with both strided and channels_last inputs in forward direction and only for channels_last in the backward direction.
- I haven't found cases where our instance norm is much slower than the others.
- Half precision matches general trends of float32.

eqy:
Is there any expected performance difference for cases covered by the old C++ implementation? If those look good and the tests are passing, then I think we can try to land.

What do you think @crcrpar ?

jacobhinkle:
I don't think so, as I tried to translate it verbatim. Note that we depend on `alias_input` arg to `add_output` which is on TOT of csarofeen/pytorch/devel but is not yet in pytorch/pytorch/master.

crcrpar:
getting ready... https://github.com/NVIDIA/Fuser/pull/59

PR #1565, opened 2023-01-17
[FMHA] Remove zero fill for softmax output

vasunvidia:


eqy (approved):

crcrpar:
let me merge this and let's investigate some race conditions reported by `compute-sanitizer --tool=racecheck` which I didn't see with CUDA11.8

PR #1567, opened 2023-01-19
[Transformer] Update test routines

Aidyn-A:
The main reason for existence of this PR is that `torch.distributed.<communication>()` without process group being specified (like fixed barriers) can do some nasty things. On single node there are no visible issues whatsoever. However, on multiple nodes with non-IB connections (like Ethernet) the `torch.distributed.<communication>()` hangs causing the timeout-failure. 
Why is this happening on multi node with Ethernet? Because `torch.distributed.<communication>()` attempts to use default PG with NCCL backend over all available ranks. Using all ranks would be only possible if all communications of NCCL were gone through TCP, which we are targeting to avoid by using UCC to do the TCP job.

Please let me know if there are other communications without specified process groups. 
cc @crcrpar @eqy @Fuzzkatt

### DIFF
--- a/apex/transformer/testing/commons.py
+++ b/apex/transformer/testing/commons.py
@@ -289,10 +289,8 @@ def initialize_distributed(backend="nccl"):
 
 
 def print_separator(message):
-    torch.distributed.barrier()
     filler_len = (78 - len(message)) // 2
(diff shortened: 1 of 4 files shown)


crcrpar (approved):
thanks

PR #1568, opened 2023-01-20
[fused softmax] integer overflow and non-contiguous memory

cyanguwa:
This PR includes two commits. One to promote certain integers to long ints so when there are more than 2B elements in the input tensor, the fused softmax doesn't crash. The other commit is to convert non-contiguous input/mask tensors to contiguous format. Currently, the code produces random results when the tensors are non-contiguous.


crcrpar (approved):

crcrpar:
note to future me: could be mitigated by using some `TensorAccessor` e.g .`PackedTensorAccessor32` and `PackedTensorAccessor64`

PR #1575, opened 2023-02-04
Decouple optimizer state and grad dtypes in distributed Adam optimizer

timmoon10:
This PR generalizes the supported dtypes for distributed Adam. In particular, it can now handle FP32, FP16, or BF16 gradients, independent of the optimizer state or param dtypes.

Depends on https://github.com/NVIDIA/apex/pull/1559.


apex/contrib/csrc/optimizers/multi_tensor_distopt_adam_kernel.cu
```diff
@@ -402,27 +401,30 @@ void multi_tensor_fused_adam_with_param_remainders_cuda(
   // Expect p_in, p_rem, m, v, g, p_out
   size_t tl_sz = tensor_lists.size();
   TORCH_CHECK(tl_sz == 6, "expected tensor lists of size 6");
+  auto g_type = tensor_lists[4][0].scalar_type();
```
crcrpar, line 404: nit-picking
```suggestion
  const auto g_type = tensor_lists[4][0].scalar_type();
```

apex/contrib/csrc/optimizers/multi_tensor_distopt_adam_kernel.cu
```diff
@@ -402,27 +401,30 @@ void multi_tensor_fused_adam_with_param_remainders_cuda(
...
-    (adamMode_t) mode,
-    weight_decay);
+  DISPATCH_FLOAT_HALF_AND_BFLOAT(g_type, 0, "dist_adam_with_param_remainders_cuda_kernel",
+    multi_tensor_apply<6>(
+      BLOCK_SIZE,
+      chunk_size,
+      noop_flag,
+      tensor_lists,
+      DistAdamWithParamRemaindersFunctor<scalar_t_0>(),
+      grad_scale.DATA_PTR<float>(),
```
crcrpar, line 419: nit-picking
```suggestion
      grad_scale.data_ptr<float>(),
```

apex/transformer/tensor_parallel/layers.py
```diff
@@ -520,7 +501,7 @@ def __init__(
         params_dtype=torch.float32,
         use_cpu_initialization=False,
         gradient_accumulation_fusion=False,
-        accumulation_in_fp16: bool = False,
+        accumulation_in_fp16: bool = False, # Deprecated
```
crcrpar, line 504: just a comment, no need to be accessed.
if this is "deprecated" (well even this transformer module itself is going to be deprecated), I'd recommend making this argument "Optional" and throwing a warning if a value is set

```suggestion
        accumulation_in_fp16: Optional[bool] = None,
```

apex/transformer/tensor_parallel/layers.py
```diff
@@ -708,7 +685,7 @@ def __init__(
...
```
crcrpar, line 688: Ditto

crcrpar (commented):
it's totally optional to access all the comments

apex/transformer/tensor_parallel/layers.py
timmoon10: I think this is a nicer approach. I don't see this option used anywhere in Apex, NeMo, or Megatron-LM, so we could even remove it if we're feeling brave.

PR #1578, opened 2023-02-06
replace torch.Tensor with torch.empty

NouamaneTazi:
Use  `torch.empty` to initialize tensors instead of `torch.Tensor` which uses the new constructor.

One benefit for this is that we can pass device and dtype directly to  `torch.empty` whereas we can't for `torch.Tensor`


crcrpar (commented):
Thank you, looks good
Just so you know, we're going to remove apex/amp and apex/RNN.

MozerWang:
BUG!
torch.empty() must have params.

PR #1579, opened 2023-02-06
Fix FusedAdam.zero_grad(set_to_none=True)

NouamaneTazi:


apex/optimizers/fused_adam.py
```diff
@@ -94,8 +94,8 @@ def __init__(self, params, lr=1e-3, bias_correction=True,
         else:
             raise RuntimeError('apex.optimizers.FusedAdam requires cuda extensions')
 
-    def zero_grad(self):
-        if self.set_grad_none:
+    def zero_grad(self, set_to_none=None):
```
crcrpar, line 97: QQ: why is the default `None`? Given that the default of `set_grad_none`' and `torch.optim.Optimizer.zero_grad`'s `set_to_none` is `True`, it'd make sense to make it default to `True` to me

thomasw21: Also would it make sense to check that `set_to_none` and `self.set_grad_none` are the same? Raise a `RunTimeError` if they are not?

thomasw21: Also note that this is a breaking change from `torch`. It defaults to False in `1.13` and True in `master`. Not sure which one would make more sense here.

apex/optimizers/fused_adam.py
```diff
@@ -94,8 +94,8 @@ def __init__(self, params, lr=1e-3, bias_correction=True,
         else:
             raise RuntimeError('apex.optimizers.FusedAdam requires cuda extensions')
 
-    def zero_grad(self):
-        if self.set_grad_none:
+    def zero_grad(self, set_to_none=True):
+        if self.set_grad_none or set_to_none:
```
NouamaneTazi, line 98: gently pinging @crcrpar
Does this look good to you? :)

PR #1580, opened 2023-02-06
Return input-loader iterator function in Transformer data loader

erhoo82:
Return the non-list input to the training function as is. This is to support the per-micro-batch input loader for Megatron Transformer.

### DIFF
--- a/apex/transformer/pipeline_parallel/utils.py
+++ b/apex/transformer/pipeline_parallel/utils.py
@@ -14,7 +14,7 @@
 # limitations under the License.
 
 """Utilities for pipeline model parallel."""
-from typing import Optional, List, Union
+from typing import Optional, List, Union, Tuple
 
 import torch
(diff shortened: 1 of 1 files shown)


apex/transformer/pipeline_parallel/utils.py
```diff
@@ -125,7 +125,7 @@ def get_kth_microbatch(batch: Optional[List[torch.Tensor]], k: int) -> List[torc
     This function creates a list of `k`th microbatches from a list of local minibatches.
     `a local minibatch` consists of `global_batch_size / data_parallel_size` samples.
     """
-    if batch is None:
+    if batch is None or not isinstance(batch, (List, Tuple)):
```
crcrpar, line 128: need to import `Tuple` (iiuc, `(list, tuple)` would also work but not tested)

crcrpar (changes requested):

PR #1581, opened 2023-02-07
decrease the number of warnings

crcrpar:
cc @xwang233

### DIFF
--- a/apex/contrib/csrc/conv_bias_relu/conv_bias_relu.cpp
+++ b/apex/contrib/csrc/conv_bias_relu/conv_bias_relu.cpp
@@ -19,8 +19,8 @@
 #define DEBUG_CUDNN_MSG(buf, str) do { } while ( false )
 #endif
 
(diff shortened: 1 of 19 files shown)


xwang233:
Thanks for the fix!

I was also wondering if we can add extra NVCC flags here from some environment variables
https://github.com/NVIDIA/apex/blob/a32d7a6dddcf4e39d241b0d139c222a97c91887d/setup.py#L53-L58

For example

```diff
def append_nvcc_threads(nvcc_extra_args):
+   nvcc_extra_args += os.getenv('APEX_NVCC_FLAGS', '').split()
    _, bare_metal_version = get_cuda_bare_metal_version(CUDA_HOME)
    if bare_metal_version >= Version("11.2"):
        return nvcc_extra_args + ["--threads", "4"]
    return nvcc_extra_args
```

cc @ptrblck

crcrpar:
I rather was once thinking of removing that function and making use of either `NVCC_PREPEND_FLAGS` or `NVCC_APPEND_FLAGS` (ref: https://docs.nvidia.com/cuda/cuda-compiler-driver-nvcc/index.html#nvcc-environment-variables)

PR #1584, opened 2023-02-14
quick fix: torch.empty() must have args

NouamaneTazi:
Fixes small bug introduced in my last PR https://github.com/NVIDIA/apex/pull/1578

credits to @MozerWang


crcrpar (approved):
thank you @NouamaneTazi  and @MozerWang

PR #1585, opened 2023-02-14
Use a separate communicator for AMAX reduction

erhoo82:
This is to enable the overlap of AMAX reduction and the other DP-communications e.g. gradient reduction and parameter gathering.

### DIFF
--- a/apex/transformer/parallel_state.py
+++ b/apex/transformer/parallel_state.py
@@ -48,6 +48,8 @@ _ENCODER_RELATIVE_POSITION_EMBEDDING_GROUP = None
 _DECODER_RELATIVE_POSITION_EMBEDDING_GROUP = None
 # Data parallel group that the current rank belongs to.
 _DATA_PARALLEL_GROUP = None
+# FP8 AMAX reduction group that the current rank belongs to.
+_DATA_PARALLEL_GROUP_AMAX_REDUCTION = None
 
 _VIRTUAL_PIPELINE_MODEL_PARALLEL_RANK = None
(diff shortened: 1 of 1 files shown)


apex/transformer/parallel_state.py
```diff
@@ -84,6 +86,7 @@ def initialize_model_parallel(
     pipeline_model_parallel_size_: int = 1,
     virtual_pipeline_model_parallel_size_: Optional[int] = None,
     pipeline_model_parallel_split_rank_: Optional[int] = None,
+    use_fp8_: Optional[bool] = None,
```
crcrpar, line 89: would the behavior with `use_fp8_=None` be different from that with `use_fp8_=False`?
I'm not sure about the need to make this `Optional` at glance

crcrpar: If the behavior will be the same, then I'd use 
```suggestion
    use_fp8_: bool = False,
```

apex/transformer/parallel_state.py
```diff
@@ -96,6 +99,7 @@ def initialize_model_parallel(
         pipeline_model_parallel_size: number of GPUs used to parallelize model pipeline.
         virtual_pipeline_model_parallel_size: number of virtual stages (interleaved pipeline).
         pipeline_model_parallel_split_rank: for models with both encoder and decoder, rank in pipeline with split point.
+        use_fp8: FP8 training that needs AMAX reduction across data-parallel ranks.
```
crcrpar, line 102: nit-picking
```suggestion
        use_fp8_: FP8 training that needs AMAX reduction across data-parallel ranks.
```

erhoo82:
Thanks for the comments. I have applied the suggested changes.

PR #1596, opened 2023-02-23
check for empty params in group

arendu:
A Recent update fails for parameter efficient fine-training (PEFT) setup. In this setup, a small subset of params are trained and the rest of the network is frozen. When using PEFT with pipeline parallel mode, it is possible that an entire rank has no trainable params. See this error log:
Screen Shot 2023-02-23 at 12 20 18 PM

This PR checks for `len(group['params'])` and defaults to 'cuda' if params are empty.

### DIFF
--- a/tests/L0/run_optimizers/test_fused_optimizer.py
+++ b/tests/L0/run_optimizers/test_fused_optimizer.py
@@ -130,7 +130,7 @@ class TestFusedAdam(TestFusedOptimizer):
             max_abs_diff, max_rel_diff = self.get_max_diff(ref_param, tst_param)
             self.assertLessEqual(max_abs_diff, self.max_abs_diff)
             self.assertLessEqual(max_rel_diff, self.max_rel_diff)
-
+            
     @unittest.skip('No longer support fuse scaling')
     def test_scale(self):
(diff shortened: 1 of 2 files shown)


crcrpar (commented):
could you add a test to https://github.com/NVIDIA/apex/tree/master/tests/L0/run_optimizers?

JimmyZhang12:
Hi @crcrpar I added a simple test that adds an empty param group to the test optimizer, which exposes the bug, can you please review? Thanks

tests/L0/run_optimizers/test_fused_optimizer.py
```diff
@@ -130,7 +130,7 @@ def test_multi_params(self):
             max_abs_diff, max_rel_diff = self.get_max_diff(ref_param, tst_param)
             self.assertLessEqual(max_abs_diff, self.max_abs_diff)
             self.assertLessEqual(max_rel_diff, self.max_rel_diff)
-
+
```
crcrpar, line 133: could you remove white spaces in an empty line?

JimmyZhang12: done!

PR #1597, opened 2023-02-24
Single communicator for amax reduction in TP and DP domains

erhoo82:
Add AMAX communicator for the reduction in both TP and DP domains.


crcrpar (approved):
confirmed no negative effects on the existing tests. lgtm

PR #1601, opened 2023-02-27
Update layers.py to ensure grad_output is contiguous

markelsanz14:
Depending on how the forward pass is implemented, grad_output in this function will not be contiguous, so it crashes when calling .view(). By adding this extra line, we ensure it's contiguous. And it's a no-op when it already is contiguous, so performance shouldn't suffer.


crcrpar (approved):
could you provide me with an example where `grad_output` is not contiguous?

PR #1611, opened 2023-03-13
Align distopt communication in interleaved pipeline parallelism

timmoon10:
While running NeMo-Megatron using the distributed optimizer and interleaved pipeline parallelism, we have observed excessive GPU idle time. The root cause is that asynchronous NCCL collectives slow down some GEMMs, so microbatches with overlapped reduce-scatters or all-gathers will take longer and other pipeline-parallel ranks have to wait to receive data. Our current implementation gets the worst-case behavior and the GEMM perf degradation is magnified by the total number of virtual pipeline stages. The fix is for all pipeline-parallel ranks to launch NCCL at the same time, so they're all similarly slowed down and the GEMM perf degradation doesn't compound.

### DIFF
--- a/apex/contrib/optimizers/distributed_fused_adam.py
+++ b/apex/contrib/optimizers/distributed_fused_adam.py
@@ -1407,37 +1407,60 @@ class DistributedFusedAdam(torch.optim.Optimizer):
                 # Cached gradient norm has been invalidated
                 self._grad_norm = None
 
-    def _try_start_bucket_param_sync(self):
+    def _try_start_bucket_param_sync(
+            self,
+            params=None,
+    ):
         """Attempt to launch parameter synchronization
 
-        Launches parameter synchronization if there is at least one
-        unsynchronized bucket and there are no other synchronizations
-        in progress. Parameter synchronization is asynchronous.
+        Launches parameter synchronization for buckets corresponding
+        to provided parameters, if needed. If parameters are not
+        provided and no other synchronizations are in progress,
+        attempts to find a parameter that still requires
+        synchronization. Parameter synchronization is asynchronous.
+
+        Arguments:
+            params (iterable, optional): parameters to synchronize
 
         """
 
-        # Only launch param sync if there is an unsynchronized bucket
-        # and no other syncs are in progress
-        if not any(bucket.status == self.ParameterStatus.SHARDED
+        # Default behavior: only launch param sync if no other syncs
+        # are in progress
+        if params is None:
+            params = []
+            if any(bucket.status == self.ParameterStatus.SYNCING
                    for bucket in self._params_buckets.values()):
-            return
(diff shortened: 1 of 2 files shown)


apex/contrib/optimizers/distributed_fused_adam.py
```diff
@@ -1407,37 +1407,60 @@ def _finish_bucket_grad_sync(self):
...
 
         """
 
-        # Only launch param sync if there is an unsynchronized bucket
-        # and no other syncs are in progress
-        if not any(bucket.status == self.ParameterStatus.SHARDED
+        # Default behavior: only launch param sync if no other syncs
+        # are in progress
+        if params is None:
+            params = []
```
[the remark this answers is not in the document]
timmoon10: The `None` and `[]` cases are slightly different:
```python
# Case with params=None
optim._try_start_bucket_param_sync()  # Want to try launching one bucket

# Case with params=[]
optim._try_start_bucket_param_sync(
  [p for p in model.parameters() if not is_disabled_for_some_reason(p)]
)  # Want no-op if list is empty
```
There's an argument for splitting into two functions, but the difference is pretty subtle.

apex/transformer/pipeline_parallel/schedules/fwd_bwd_pipelining_with_interleaving.py
```diff
@@ -284,6 +329,19 @@ def backward_step_helper(microbatch_id: int) -> torch.Tensor:
...
+        # assumes all ranks have the same compute time. However,
+        # asynchronous communication tends to slow down compute. Thus,
+        # we launch asynchronous communication at the same time across
+        # the pipeline-parallel group.
+        if custom_grad_sync_func is not None:
+            grad_sync_microbatch_id = microbatch_id - pipeline_parallel_rank
+            if grad_sync_microbatch_id >= 0 and is_last_microbatch_for_model_chunk(grad_sync_microbatch_id):
+                grad_sync_chunk_id = get_model_chunk_id(grad_sync_microbatch_id, forward=False)
+                enable_grad_sync()
+                custom_grad_sync_func(model[grad_sync_chunk_id].parameters())
```
crcrpar, line 344: just to confirm: it's user's responsibility to take care of `custom_grad_sync_func` appropriately (= create a list of gradients from the list of parameters). is this correct?

crcrpar (approved):

apex/transformer/pipeline_parallel/schedules/fwd_bwd_pipelining_with_interleaving.py
timmoon10: Yep, that's right. A minimal implementation would be:
```python
def grad_sync(params):
  for param in params:
    torch.distributed.all_reduce(param.grad)
```

PR #1620, opened 2023-03-17
[Transformer] Manipulate NCCL_NET for hybrid IB/Socket setups

Aidyn-A:


apex/transformer/parallel_state.py
```diff
@@ -80,6 +79,56 @@ def is_unitialized():
...
+    environment variable NUM_GPUS_PER_BLOCK is defined it looks up the ranks
+    and determines whether they all belong to the same computational block or not.
+    If all ranks are in the same blocks the process group will use NCCL_NET=IB 
+    communication otherwise it will use NCCL_NET=Socket. 
+    
+    If NCCL_NET=Socket is to be used, the user must set NCCL_SOCKET_IFNAME.
+    Additionally, it is recommended to set NCCL_SOCKET_NTHREADS
+    and NCCL_NSOCKS_PERTHREAD before running the job.
+    """
+    if backend us None:
```
crcrpar, line 118:
```suggestion
    if backend is None:
```

apex/transformer/parallel_state.py
```diff
@@ -80,6 +79,56 @@ def is_unitialized():
...
+        backend = "nccl"
+
+    compute_block_size = os.getenv("NUM_GPUS_PER_BLOCK")
+    if backend == "nccl" and compute_block_size is not None:
+        compute_block_size = int(compute_block_size)
+        blocks = [rank // compute_block_size for rank in ranks]
+        use_ib = all(block==blocks[0] for block in blocks)
```
crcrpar, line 125:
```suggestion
        use_ib = all(block == blocks[0] for block in blocks)
```

apex/transformer/parallel_state.py
```diff
@@ -126,13 +175,13 @@ def initialize_model_parallel(
     with a total of 16 GPUs, rank 0 to 7 belong to the first box and
     ranks 8 to 15 belong to the second box.
     """
+    default_nccl_net = os.getenv("NCCL_NET")
```
crcrpar, line 178: Q: is this variable used other than the epilogue of this function?

Aidyn-A: Not, it should not be used anywhere else than in epilogue

crcrpar: then could you move this definition to right before the epilogue for a little bit better clarity?

Aidyn-A: I can move it down a little bit, but no further than the first `new_process_group` where `NCCL_NET` is going to be modified.

apex/transformer/parallel_state.py
```diff
@@ -80,6 +79,73 @@ def is_unitialized():
...
+    and determines whether the set of ranks belong to the same computational block where
+    GPUs are interconnected via IB type of connection or not.
+    If all ranks are in the same block, the process group will use NCCL_NET=IB for
+    communication otherwise it will use NCCL_NET=Socket. 
+    
+    If NCCL_NET=Socket is ever to be used, the user must set NCCL_SOCKET_IFNAME.
+    Additionally, it is recommended to set NCCL_SOCKET_NTHREADS and
+    NCCL_NSOCKS_PERTHREAD before running the job.
+    See: https://docs.nvidia.com/deeplearning/nccl/user-guide/docs/env.html
+    for more info
```
crcrpar, line 130: I have one newbie question. What will happen if Socket is used without `NCCL_SOCKET_IFNAME` specified by a user?

Aidyn-A: It is some sort of an undefined behavior. It may work if there is only one interface. However, if there are more than one interfaces (eg. `eth0`, `eth1`, `eth2` ...) it will hang or throw an error in best scenario.

crcrpar: hmm, sounds like it (= Socket is used but the env var is missing) deserves RuntimeError if it could lead to unknown status

apex/transformer/parallel_state.py
```diff
@@ -17,11 +17,10 @@
 """Model and data parallel groups."""
 from typing import Tuple, Optional
 import warnings
-
+import os
 import torch
 
 from apex.transformer.log_util import get_transformer_logger
-from apex.transformer._ucc_util import HAS_UCC
```
crcrpar, line 24: oh I think this shouldn't be removed.
Could you revert this deletion?

Aidyn-A: Oops, sorry, I missed that one :sweat_smile:  Thanks!

crcrpar (approved):

PR #1621, opened 2023-03-17
Overlap first param sync with optimizer step in distributed Adam

timmoon10:
When you try overlapping forward compute with distopt param all-gathers, one all-gather is fully exposed. This PR overlaps that all-gather with the optimizer step.

### DIFF
--- a/apex/contrib/optimizers/distributed_fused_adam.py
+++ b/apex/contrib/optimizers/distributed_fused_adam.py
@@ -1794,29 +1794,61 @@ class DistributedFusedAdam(torch.optim.Optimizer):
                     device=self.device,
                 )
 
-        # Apply optimizer step
+        # Apply optimizer step and synchronize params
         self.state['step'] += 1
-        if self.store_param_remainders:
-            self._local_step_with_param_remainders()
-        else:
-            self._local_step()
+        if self.distributed_size > 1 and self.overlap_param_sync and self.state['buckets']:
+            # Local step and non-blocking param sync
+            # Note: Overlap param sync of first buckets with optimizer
+            # step of remaining buckets.
 
-        # Make sure all parameters are synchronized if needed
-        if self.distributed_size > 1 and self.overlap_param_sync:
-            self._try_start_bucket_param_sync()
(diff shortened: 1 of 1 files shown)


apex/contrib/optimizers/distributed_fused_adam.py
```diff
@@ -1794,29 +1794,66 @@ def step(self, closure=None, *, grad_scaler=None):
...
-            self._local_step_with_param_remainders()
-        else:
-            self._local_step()
+        if self.distributed_size > 1 and self.overlap_param_sync and self.state['buckets']:
+            # Local step and non-blocking param sync
+            # Note: Overlap param sync of first buckets with optimizer
+            # step of remaining buckets.
+
+            # Get buckets containing "first" parameter
+            fragment = self.state['buckets'][-1].fragments[-1]
```
crcrpar, line 1805: excuse me for my lack of understanding but why the last bucket and the last fragment?

apex/contrib/optimizers/distributed_fused_adam.py
```diff
@@ -1794,29 +1794,66 @@ def step(self, closure=None, *, grad_scaler=None):
...
+
+            # Local step and launch param sync for first buckets
+            self._local_step(first_bucket_ids)
+            self._start_bucket_param_sync(
+                self._params_buckets[bucket_id]
+                for bucket_id in first_bucket_ids
+            )
+
+            # Local step for remaining buckets
+            first_bucket_ids = set(first_bucket_ids)
```
crcrpar, line 1822: this seems to allow duplicated bucket ids in the call of `_lcoal_step` above but would it be the case in the use case?

crcrpar (approved):
my two quick questions

apex/contrib/optimizers/distributed_fused_adam.py
timmoon10: By default, each param initializes its optimizer state in the first backward pass, so the first param to show up in the forward pass should be the last one to initialize.

PR #1623, opened 2023-03-22
Add support for FP32 master weights in FusedAdam optimizer

minitu:
This PR adds support for FP32 master weights in FusedAdam optimizer, to improve performance without affecting convergence when the model is trained with PyTorch AMP and some layers are manually cast to FP16.

Currently being debugged as there seems to be a slight regression in convergence.

### DIFF
--- a/apex/optimizers/fused_adam.py
+++ b/apex/optimizers/fused_adam.py
@@ -146,10 +146,11 @@ class FusedAdam(torch.optim.Optimizer):
             p_16_model = []
 
             for pi, p in enumerate(group['params']):
-                if p.grad is None and not self.use_master:
-                    continue
-                if p.grad.data.is_sparse:
-                    raise RuntimeError('FusedAdam does not support sparse gradients, please consider SparseAdam instead')
+                if not self.use_master:
+                    if p.grad is None:
+                        continue
+                    if p.grad.data.is_sparse:
+                        raise RuntimeError('FusedAdam does not support sparse gradients, please consider SparseAdam instead')
 
                 state = self.state[p]
                 # State initialization
--- a/apex/optimizers/fused_adam.py
+++ b/apex/optimizers/fused_adam.py
@@ -160,8 +160,8 @@ class FusedAdam(torch.optim.Optimizer):
                         p_16_master.append(self.master_params[pi])
                     g_16.append(p.grad.data)
                     p_16.append(p.data)
-                    m_16.append(state['exp_avg']
-                    v_16.append(state['exp_avg_sq']
+                    m_16.append(state['exp_avg'])
+                    v_16.append(state['exp_avg_sq'])
                 elif p.dtype == torch.bfloat16:
                     g_bf.append(p.grad)
                     p_bf.append(p)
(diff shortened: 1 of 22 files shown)


apex/optimizers/fused_adam.py
```diff
@@ -144,15 +151,17 @@ def step(self, closure=None, grads=None, output_params=None, scale=None, grad_no
                 # State initialization
                 if len(state) == 0:
                     # Exponential moving average of gradient values
-                    state['exp_avg'] = torch.zeros_like(p.data)
+                    state['exp_avg'] = torch.zeros_like(p.data).float()
```
minitu, line 154: I found that the m and v tensors also need to be in full precision for training to work properly, hence the changes here.

apex/optimizers/fused_adam.py
```diff
@@ -144,15 +151,17 @@ def step(self, closure=None, grads=None, output_params=None, scale=None, grad_no
...
+                    state['exp_avg_sq'] = torch.zeros_like(p.data).float()
 
                 if p.dtype == torch.float16:
+                    if self.use_master:
+                        p_16_master.append(self.master_params[pi])
                     g_16.append(p.grad.data)
                     p_16.append(p.data)
-                    m_16.append(state['exp_avg'])
-                    v_16.append(state['exp_avg_sq'])
+                    m_16.append(state['exp_avg'].float())
```
minitu, line 163: Same as above

csrc/multi_tensor_adam.cu
```diff
@@ -189,11 +189,10 @@ struct AdamCapturableFunctor
         int i = i_start + threadIdx.x + ii*blockDim.x;
         if(i < n && i < chunk_size)
         {
-	  g[i] = g[i] * (*inv_scale);
```
minitu, line 192: If `g[i]` is in FP16 and the FP32 multiplication result is out of FP16 range, it may cause issues when the result is cast to FP16 to be stored in `g[i]`.

csrc/multi_tensor_adam.cu
```diff
@@ -269,7 +386,7 @@ void multi_tensor_adam_cuda(
       chunk_size,
       noop_flag,
       tensor_lists,
-      AdamFunctor<scalar_t_0>(),
+      AdamFunctor<scalar_t_0, float>(),
```
crcrpar, line 389: could be regarded as backward compatibility breaking but this makes the kernel safer.

apex/optimizers/fused_adam.py
minitu: @crcrpar You're right, this is not necessary and works without it. Removed

csrc/multi_tensor_adam.cu
minitu: @crcrpar I can limit my changes to only affect the `use_master=True` case, although I do think the original code was broken in that it doesn't train at all when I pass in FP16 model parameters as it that will try to use FP16 m and v values.

crcrpar: yeah I'm comfortable with this functional improvement. appreciate your change :)

minitu:
@crcrpar We're actually looking into some degradation in convergence with this change, should we revert this PR or submit a new one with fixes?

crcrpar:
which works better to you?

minitu:
@crcrpar I'm fine with either but I'd prefer reverting it so that we keep all the relevant changes in one place.

PR #1626, opened 2023-03-25
Create mpi proc group for userbuffer communication kernels

erhoo82:


erhoo82:
I create the mpi process group in parallel_stage, where other process groups are also created and initialized.

crcrpar (approved):

PR #1628, opened 2023-03-28
Grouped conv permutations

jpool-nv:
This is a large refactoring of the channel permutation application code (the permutation _search_ code is not affected) with the following changes, motivated by the first point below:

- Grouped convolution support is improved (mobilenets, etc.)
- MHA support is improved (transformers)
- Module parameters can be permuted
- Many robustness improvements
- Parallelized search with multiple GPUs
- Regression tests added
- Improved readability/maintainability

All torchvision reference models are supported without errors.


jpool-nv:
@ChongyuNVIDIA , can you confirm you've reviewed this PR?

ChongyuNVIDIA:
> @ChongyuNVIDIA , can you confirm you've reviewed this PR?

Yes. I have reviewed this PR offline. It is in good shape now.

apex/contrib/sparsity/permutation_lib.py
```diff
@@ -808,112 +1467,200 @@ def build_fx_graph(cls, model, dump_fx_graph=False, save_dumped_fx_graph='./mode
             torch_version_minimum = int(torch_version.split('.')[2])
         except ValueError:    # support the none standard version
             torch_version_minimum = torch_version.split('.')[2]
-        print("[build_fx_graph] The torch version is: {}, version major is: {}, version minor is: {}, version minimum is: {}".format(torch_version, torch_version_major, torch_version_minor, torch_version_minimum))
+        if cls.__verbosity > 2:
+            print("[build_fx_graph] The torch version is: {}, version major is: {}, version minor is: {}, version minimum is: {}".format(torch_version, torch_version_major, torch_version_minor, torch_version_minimum))
+
         if torch_version_major >= 1 and torch_version_minor >= 8:
```
crcrpar, line 1473: Would you know if this module works with PyTorch 2.0 and later?
If so, I suppose the appropriate condition is

```python
torch_version_major >= 2 or (torch_version_major >= 1 and torch_version_minor >= 8)`
```

or use packaging which is a dependency of apex to do comparison of versions (ref: https://packaging.pypa.io/en/stable/version.html)

jpool-nv: I just tested with NGC's pytorch:23.03-py3, which includes PyTorch 2.0, and the regression tests and some full network tests complete successfully with your suggested modification.  I added it and a workaround for the missing 'tabulate' library in 437ce1f.

PR #1634, opened 2023-04-05
Using regular layernorm for ONNX and TS export

borisfom:
Here, we can't use Autograd.Function and Apex nodes when exporting. 
There may be better way to do that - but that at least let onnx.export, torch.jit.script(0 and torch.trace_module() be used. 
We used to replace Apex norm nodes with regular norms in Nemo - better to address at the source.

### DIFF
--- a/tests/L0/run_fused_layer_norm/test_fused_layer_norm.py
+++ b/tests/L0/run_fused_layer_norm/test_fused_layer_norm.py
@@ -233,8 +233,6 @@ class TestFusedLayerNorm(common_utils.TestCase):
 
         tols = {'rtol': None, 'atol': None} if dtype == torch.half else bf16_bwd_thresholds
         torch.testing.assert_close(native_x.grad, fused_x.grad, **tols, check_dtype=False)
-        # check that export() is working
-        verify(fused, fused_x)
         
     @common_utils.parametrize(
         "dtype, elementwise_affine",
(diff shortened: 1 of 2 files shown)


crcrpar (commented):
basically sounds okay to me but would you mind adding tests to https://github.com/NVIDIA/apex/blob/89cc215a49b0e99263a8184f17f17275879015aa/tests/L0/run_fused_layer_norm/test_fused_layer_norm.py in order to verify that this change covers the cases with which you've bothered?

borisfom:
@crcrpar : Added unit tests.

tests/L0/run_fused_layer_norm/test_fused_layer_norm.py
crcrpar, line 1: Would you know if this introduces the dependency on `onnx`? If `onnx` is required by `verify`, could you make the call of `verify` conditional according to the availability of it?

borisfom: Well, that would make the whole test conditional ... 
The problem is that I can't just use torch.onnx.export() for unit test - as it does succeed, but produces wrong graph (one constant node instead of whole LayerNorm math). I guess I can try 'verbose' arg to export() and analyze its output to avoid onnx dependency.

crcrpar: I mean, we'd better list up the modules the new test require.

Currently, it seems like onnx and onnxruntime are needed for this test but they are not mentioned.

borisfom: Changed the tests to not depend on new packages.

tests/L0/run_fused_layer_norm/test_fused_layer_norm.py
```diff
@@ -269,5 +269,41 @@ def test_autocast_fused_rms_norm(self, dtype, elementwise_affine):
...
+    normalized_shape = [32, 16]
+    fused = FusedLayerNorm(
+        normalized_shape=normalized_shape, elementwise_affine=True
+    ).cuda()
+    fused_m = MixedFusedLayerNorm(
+        normalized_shape=normalized_shape, elementwise_affine=True
+    ).cuda()
+    native_x, fused_x = _prep_inputs(batch_size, normalized_shape, torch.float32)
+    _verify_export(fused, fused_x)
+    _verify_export(fused_m, fused_x)
```
crcrpar, line 306: Could you implement these as [`unittest.TestCase`](https://docs.python.org/3/library/unittest.html#unittest.TestCase)'s methods so that an environment w/o pytest can also run?

tests/L0/run_fused_layer_norm/test_fused_layer_norm.py
```diff
@@ -269,5 +269,41 @@ def test_autocast_fused_rms_norm(self, dtype, elementwise_affine):
 
 instantiate_device_type_tests(TestFusedLayerNorm, globals(), only_for=("cuda",))
 
+def _verify_export(fused, fused_x):
+    # check that export() is working
+    onnx_str = torch.onnx.export_to_pretty_string(fused, (fused_x,),
+                                                  input_names=['x_in'],
+    )
+    assert('x_in' in onnx_str)
+    assert('ReduceMean' in onnx_str)
```
crcrpar, line 278: [nit-picking] [assert expression](https://docs.python.org/3/reference/simple_stmts.html?highlight=assert#the-assert-statement)

```suggestion
    assert 'x_in' in onnx_str
    assert 'ReduceMean' in onnx_str
```

crcrpar (commented):
Thank you for coming up with this test w/o the dependency on onnx & onnxruntime

tests/L0/run_fused_layer_norm/test_fused_layer_norm.py
borisfom: Done.

tests/L0/run_fused_layer_norm/test_fused_layer_norm.py
```diff
@@ -266,8 +265,41 @@ def test_autocast_fused_rms_norm(self, dtype, elementwise_affine):
...
+        fused_m = MixedFusedLayerNorm(
+            normalized_shape=normalized_shape, elementwise_affine=True
+        ).cuda()
+        native_x, fused_x = _prep_inputs(batch_size, normalized_shape, torch.float32)
+        self._verify_export(fused, fused_x)
+        self._verify_export(fused_m, fused_x)
+        
 instantiate_device_type_tests(TestFusedLayerNorm, globals(), only_for=("cuda",))
-
+
```
crcrpar, line 303:
```suggestion

```

tests/L0/run_fused_layer_norm/test_fused_layer_norm.py
```diff
@@ -233,7 +232,7 @@ def test_autocast_fused_layer_norm(self, dtype, elementwise_affine):
 
         tols = {'rtol': None, 'atol': None} if dtype == torch.half else bf16_bwd_thresholds
         torch.testing.assert_close(native_x.grad, fused_x.grad, **tols, check_dtype=False)
-
+
```
crcrpar, line 235:
```suggestion

```

crcrpar (approved):

PR #1651, opened 2023-04-26
[Transformer][test/L0] Increase max world_size for BERT and GPT

Aidyn-A:
Fixes a mismatch between `num_devices` and `world_size` that causes an incorrect setup and subsequently a test failure for 8 GPUs.


crcrpar (approved):
that was a great finding. LGTM

PR #1660, opened 2023-05-12
get compatible with the latest `torch.distributed.distributed_c10d._coalescing_manager`

crcrpar:


timmoon10:
I have some modifications to ensure backwards compatibility: https://github.com/timmoon10/apex/tree/contrib-optimizer/coalescing_manager
It runs fine when I run with the 23.02 NGC container (PyTorch 1.14.0), but I haven't tried with the latest PyTorch yet.

PR #1664, opened 2023-05-12
Support for FP32 input dtype

khj94:
The Megatron code was recently updated to use apex's layernorm code.
However, the layernorm code in the previous Megatron was considered for fp32, but the apex code was not considered, so I made a PR.
In the case of code using `--fp32-residual-connection` in Megatron, fp32 dtype is used for layernorm input, but when learning with bf16 or fp16, there is a problem that layernorm of apex cannot be used.


khj94:
> Quick question: why isn't `layer_norm_affine_mixed_dtypes` used in such situation?

Thank you for quick response.
I confirmed that the function I thought was implemented.
I closed the PR as it felt more appropriate to PR Megatron-LM to use that function.

PR #1667, opened 2023-05-16
Update setup.py

RedaGrace:


ptrblck:
The error is valid and suggests you can comment it out at your own risk. 
Your PR doesn't explain anything, in particular why it's not needed, so I'll close it.

Issue #1645, opened 2023-04-19
Error installing Apex


410011max:
**Describe the Bug**
```
 csrc/fused_dense.cpp: In lambda function:
  csrc/fused_dense.cpp:159:94: warning: ‘at::DeprecatedTypeProperties& at::Tensor::type() const’ is deprecated: Tensor.type() is deprecated. Instead use Tensor.options(), which in many cases (e.g. in a constructor) is a drop-in replacement. If you were using data from type(), that is now available from Tensor itself, so instead of tensor.type().scalar_type(), use tensor.scalar_type() instead and instead of tensor.type().backend() use tensor.device(). [-Wdeprecated-declarations]
    159 |   AT_DISPATCH_FLOATING_TYPES_AND2(at::ScalarType::Half, at::ScalarType::BFloat16, input.type(), "linear_bias_backward", [&] {
        |                                                                                              ^
  /home/max410011/.conda/envs/nvidia/lib/python3.10/site-packages/torch/include/c10/util/Exception.h:260:39: note: in definition of macro ‘C10_EXPAND_MSVC_WORKAROUND’
    260 | #define C10_EXPAND_MSVC_WORKAROUND(x) x
        |                                       ^
  /home/max410011/.conda/envs/nvidia/lib/python3.10/site-packages/torch/include/c10/util/Exception.h:465:9: note: in expansion of macro ‘TORCH_CHECK_MSG’
    465 |         TORCH_CHECK_MSG(cond, "", ##__VA_ARGS__)); \
        |         ^~~~~~~~~~~~~~~
  /home/max410011/.conda/envs/nvidia/lib/python3.10/site-packages/torch/include/c10/util/Exception.h:637:32: note: in expansion of macro ‘TORCH_CHECK’
    637 |     C10_EXPAND_MSVC_WORKAROUND(TORCH_CHECK(false, ::c10::str(__VA_ARGS__))); \
        |                                ^~~~~~~~~~~
  /home/max410011/.conda/envs/nvidia/lib/python3.10/site-packages/torch/include/ATen/Dispatch.h:330:9: note: in expansion of macro ‘AT_ERROR’
    330 |         AT_ERROR(#NAME, " not implemented for '", toString(TYPE), "'");     \
        |         ^~~~~~~~
  csrc/fused_dense.cpp:159:3: note: in expansion of macro ‘AT_DISPATCH_FLOATING_TYPES_AND2’
[...]

PR #1672, opened 2023-05-31
current code cannot build due to tensor.type()

ycsos:
when building current code, it will failed for toString() not found, like [issus1645](https://github.com/NVIDIA/apex/issues/1645)

the root cause this problem is:
apex/csrc/fused_dense.cpp:154:72: warning: ‘at::DeprecatedTypeProperties& at::Tensor::type() const’ is deprecated: Tensor.type() is deprecated. Instead use Tensor.options(), which in many cases (e.g. in a constructor) is a drop-in replacement. If you were using data from type(), that is now available from Tensor itself, so instead of tensor.type().scalar_type(), use tensor.scalar_type() instead and instead of tensor.type().backend() use tensor.device(). [-Wdeprecated-declarations]

this PR change type() to scalar_type() will fix this


crcrpar (approved):
thank you for this fix

crcrpar (approved):
thank you for this fix

PR #1673, opened 2023-06-01
[Test][Transformer] Pre-parse container version

Aidyn-A:
The test `tests/L0/run_transformer/test_pipeline_parallel_fwd_bwd.py` is failing in PyTorch container with non-`"YY.MM"` format of `NVIDIA_PYTORCH_VERSION`. This PR adds a workaround by preprocessing the version string.

cc @crcrpar


Aidyn-A:
>removing the check can be an alternative as well

I am kinda paranoid about backward compatibility, but I guess it is now okay to completely remove this check.
Should we remove it just in the test or in the transformer as well?

PR #1684, opened 2023-06-20
Tkurth/mplamb fixed

azrael417:
This MR fixes the Mixed Precision LAMB optimizer: note that param_groups is not set up before the optimizer module init was called. Therefore I swapped the order of obtaining device info and the super module init around. I tested it and it seems to work. This is critical for MLPerf HPC, please review and merge asap.

### DIFF
--- a/apex/optimizers/fused_mixed_precision_lamb.py
+++ b/apex/optimizers/fused_mixed_precision_lamb.py
@@ -26,12 +26,10 @@ class FusedMixedPrecisionLamb(torch.optim.Optimizer):
 
         # init base module
         super(FusedMixedPrecisionLamb, self).__init__(params, defaults)
-        
-        # The learning rate (lr) and optimizer step (step) should be located on device
-        device = self.param_groups[0]['params'][0].device
 
         # The learning rate (lr) and optimizer step (step) should be located on device
-	# in order to faciliated device sync free execution  
+        # in order to faciliated device sync free execution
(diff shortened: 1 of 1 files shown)


apex/optimizers/fused_mixed_precision_lamb.py
```diff
@@ -12,22 +12,27 @@ def __init__(self, params, lr=1e-3, step=0, bias_correction=True,
...
-        super(FusedMixedPrecisionLamb, self).__init__(params, defaults)
 
+        # init base module
+        super(FusedMixedPrecisionLamb, self).__init__(params, defaults)
+        
+        # The learning rate (lr) and optimizer step (step) should be located on device
         device = self.param_groups[0]['params'][0].device
 
+        # The learning rate (lr) and optimizer step (step) should be located on device
+	# in order to faciliated device sync free execution
```
Aidyn-A, line 34: nit: please fix indentation

apex/contrib/csrc/cudnn-frontend
Aidyn-A, line 1: I do not think updating the `cudnn-frontend` is in the scope of this PR, is it?

Aidyn-A (changes requested):
The changes in `fused_mixed_precision_lamb.py` make sense, just needs a tiny indentation fix. However, I believe the `cudnn-frontend` is not-related and should not be modified at least in this PR.

Aidyn-A (approved):

PR #1685, opened 2023-06-22
Tkurth/sgbn fixes

azrael417:
This PR fixes the single node group batch norm in APEX to work with cuda 12.2 and RTC.

### DIFF
--- a/apex/contrib/csrc/cudnn_gbn/cudnn_gbn.cpp
+++ b/apex/contrib/csrc/cudnn_gbn/cudnn_gbn.cpp
@@ -57,7 +57,7 @@ at::Tensor gbn_forward(const at::Tensor& x,
   std::vector<int64_t> fv = {(int64_t)BN_FWD, N, C, H, W, bn_group, (int64_t)CUDNN_DATA_HALF};
   if ( gbn_plan_cache.find(fv) == gbn_plan_cache.end() ) {
     auto plan = run_batch_norm_forward(tensorDims, perChannelDims, epsilonDims, peerDims, CUDNN_DATA_HALF);
-    gbn_plan_cache.insert(std::make_pair(fv, plan));
+    gbn_plan_cache.emplace(fv, std::move(plan));
   }
 
   // get plan and handle
@@ -130,7 +130,7 @@ std::vector<at::Tensor> gbn_backward(
   std::vector<int64_t> fv = {(int64_t)BN_BWD, N, C, H, W, bn_group, (int64_t)CUDNN_DATA_HALF};
   if ( gbn_plan_cache.find(fv) == gbn_plan_cache.end() ) {
     auto plan = run_batch_norm_backward(tensorDims, perChannelDims, epsilonDims, peerDims, CUDNN_DATA_HALF);
-    gbn_plan_cache.insert(std::make_pair(fv, plan));
+    gbn_plan_cache.emplace(fv, std::move(plan));
   }
   
   // get plan and handle
--- a/apex/contrib/csrc/cudnn_gbn/cudnn_gbn.cpp
+++ b/apex/contrib/csrc/cudnn_gbn/cudnn_gbn.cpp
@@ -7,6 +7,12 @@
 
 #include "norm_sample.h"
 
+// define this enum:
+enum bn_type { BN_FWD, BN_BWD };
+
+// this is a global variable
+static std::map<std::vector<int64_t>, cudnn_frontend::ExecutionPlan> gbn_plan_cache;
+
 at::Tensor gbn_forward(const at::Tensor& x,
                        const at::Tensor& scale,
                        const at::Tensor& bias,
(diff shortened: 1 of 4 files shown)


apex/contrib/csrc/cudnn_gbn/norm_sample.cpp
```diff
@@ -27,448 +27,453 @@
...
-        }
-    } else {
-        // Here we assume that the format is CUDNN_TENSOR_NHWC
-        strideA[1]          = 1;
-        strideA[nbDims - 1] = strideA[1] * dimA[1];
-        for (int64_t d = nbDims - 2; d >= 2; d--) {
-            strideA[d] = strideA[d + 1] * dimA[d + 1];
-        }
-        strideA[0] = strideA[2] * dimA[2];
+void generateStrides(const int64_t* dimA, int64_t* strideA, int64_t nbDims, cudnnTensorFormat_t filterFormat) {
```
eqy, line 53: Is it possible to simply use the Tensor's existing `.strides()` rather than relying on another helper function? AFAIK it would respect the NHWC vs. NCHW convention.

apex/contrib/csrc/cudnn_gbn/norm_sample.cpp
```diff
@@ -27,448 +27,453 @@
...
+  int64_t peer_stride[4];
+
+  // NHWC format. GenerateStrides() takes care of this. Howeever, tensor dims should still be NCHW
+  generateStrides(tensorDims, tensor_stride, (int64_t)4, CUDNN_TENSOR_NHWC);
+  generateStrides(peerDims, peer_stride, (int64_t)4, CUDNN_TENSOR_NHWC);
+
+  auto tensor_create = [&tensor_stride, &tensorDims](cudnnDataType_t type,
+  int64_t id) {
+    return cudnn_frontend::TensorBuilder()
+      .setDim(4, tensorDims)
```
eqy, line 95: Similarly, can we use the existing `.sizes()` instead of creating another `tensorDims` array?

apex/contrib/csrc/cudnn_gbn/cudnn_gbn.cpp
```diff
@@ -38,28 +44,43 @@ at::Tensor gbn_forward(const at::Tensor& x,
...
-    rank_id
-  );
 
+  // check if plan already exists
+  std::vector<int64_t> fv = {(int64_t)BN_FWD, N, C, H, W, bn_group, (int64_t)CUDNN_DATA_HALF};
+  if ( gbn_plan_cache.find(fv) == gbn_plan_cache.end() ) {
+    auto plan = run_batch_norm_forward(tensorDims, perChannelDims, epsilonDims, peerDims, CUDNN_DATA_HALF);
+    gbn_plan_cache.insert(std::make_pair(fv, plan));
+  }
+
```
eqy, line 62: It looks like some of the code makes assumptions about the input tensor(s)' memory layout. If so, there should be checks like `is_contiguous(at::MemoryFormat::ChannelsLast)`.

apex/contrib/csrc/cudnn_gbn/norm_sample.cpp
```diff
@@ -27,448 +27,453 @@
...
+        .setStrides(4, tensor_stride)
+          .setId(id)
+            .setAlignment(16)
```
eqy, line 98: Manually setting alignment without checking the actual tensor address seems dangerous.

eqy (changes requested):
Took a first look, but given the size of this PR I believe @crcrpar  should get the final say

crcrpar (commented):
wouldn't this require any changes to https://github.com/NVIDIA/apex/blob/30a7ad3974b32f7ce68cefabc38374fb4520a35e/apex/contrib/test/cudnn_gbn/test_cudnn_gbn_with_two_gpus.py?

apex/contrib/csrc/cudnn_gbn/norm_sample.cpp
azrael417: this is how it was implemented before though, I basically based the new implementation on the old one. We could of course use strides and pass the stride tensors to the routine. Do we want to change that?

apex/contrib/csrc/cudnn_gbn/cudnn_gbn.cpp
azrael417: This is done on the python frontend. That check it [here](https://github.com/azrael417/apex/blob/96b961f38128cce0452d53c8b501e28e220b28e2/apex/contrib/cudnn_gbn/batch_norm.py#L122)

apex/contrib/csrc/cudnn_gbn/norm_sample.cpp
azrael417: the existing code has that too: this is just a refactor of the code which is already present:

https://github.com/NVIDIA/apex/blob/master/apex/contrib/csrc/cudnn_gbn/norm_sample.cpp

This is somewhat urgent since it is fixing a showstopper bug for mlperf hpc 3.0. I am fine with rewriting this but I want to move fast on this. Is there an example of how this should be done?

eqy: Yes, e.g., https://github.com/pytorch/pytorch/blob/004ff536e87c9586064fb49c4e581f185f3a9d47/aten/src/ATen/native/cudnn/Conv_v8.cpp#L55

eqy:
@azrael417 we can defer addressing the issues I brought up to a later PR if @crcrpar is content to merge the fix given the urgency

crcrpar:
rel: https://github.com/NVIDIA/apex/issues/1689

PR #1688, opened 2023-06-30
A FasterRMSNorm implementation (based on FasterLayerNorm)

Njuapp:
A FasterRMSNorm implementation (similar to FasterLayerNorm).

### DIFF
--- a/apex/contrib/test/layer_norm/test_fast_rms_norm.py
+++ b/apex/contrib/test/layer_norm/test_fast_rms_norm.py
@@ -1,0 +1,265 @@
+import unittest
+
+import torch
+
+SKIP_TEST = None
+try:
+    from apex.contrib.layer_norm.layer_norm import FastRMSNorm
+    import fast_layer_norm as fln
+except ImportError as e:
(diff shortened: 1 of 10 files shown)


apex/contrib/test/layer_norm/test_fast_rms_norm.py
```diff
@@ -0,0 +1,265 @@
...
+
+        ms_bwd = timer.millis() / runs
+
+        print(
+            "[BWD] Time: {:.4f}ms Throughput: {:.4f} GB/sec".format(
+                ms_bwd, total_bytes_bwd * 1e-6 / ms_bwd
+            )
+        )
+
+def _test_impl(S, B, hidden_size, itype, wtype, ctype=fp32):
```
crcrpar, line 125: I understand this test is based on the existing one for faster layer norm but could you access the following two items?
1. make this function a member method of the test class below
2. use `torch.testing.assert_close` instead of `check_err` and `TestFastRMSNorm.assertAll`

crcrpar (changes requested):

PR #1690, opened 2023-07-05
Fix installation command

janEbert:
`--build-option` is not the correct config setting to use; we need `--global-option`.

Ref #1679.


janEbert:
Actually this command suffers from the same deprecation problems that the previous command tried to solve, see https://github.com/NVIDIA/apex/issues/1679#issuecomment-1622863439.

I assume the flags specified in `--build-options` need to be accessed differently in `setup.py`.

janEbert:
See #1692.

PR #1695, opened 2023-07-13
Fast CUDA NHWC Group Norm

alpha0422:
This PR creates efficient CUDA NHWC Group Norm implementation, which is required for a lot of diffusion models like Stable Diffusion. We tested with Stable Diffusion and Imagen, it improves the performance a lot.

The CUDA kernels were originally developed by Julien Demouth and Nikita Korobov.


alpha0422:
Hi @ptrblck , could you take a review and merge the PR? This is required by various diffusion models, thanks!

apex/contrib/csrc/group_norm/group_norm_nhwc.cpp
```diff
@@ -0,0 +1,1100 @@
+/***************************************************************************************************
+ * Copyright (c) 2011-2022, NVIDIA CORPORATION.  All rights reserved.
```
crcrpar, line 2: Shouldn't 2023 be included as well?

apex/contrib/group_norm/group_norm.py
```diff
@@ -0,0 +1,209 @@
...
+    y = F.group_norm(x, g, w, b, eps)
+    if act in ["silu", "swish"]:
+        y = F.silu(y)
+    if xdtype != wdtype and y.dtype != xdtype:
+        y = y.to(dtype=xdtype)
+    return y
+
+
+@torch.compile
+def inductor_group_norm(x, g, w, b, eps, act=""):
```
crcrpar, line 50: when would this be used?

apex/contrib/group_norm/group_norm.py
```diff
@@ -0,0 +1,209 @@
...
+from torch import Tensor
+from torch.nn.parameter import Parameter
+from torch._dynamo import disable
+from functools import partial
+
+__all__ = [
+    'GroupNormOpt',
+    'cuda_group_norm_nhwc',
+    'cuda_group_norm_nhwc_one_pass',
+    'cuda_group_norm_nhwc_two_pass',
```
crcrpar, line 32: would you expect these to be used directly?
or would `GroupNormOpt` be the one-stop shop?

apex/contrib/group_norm/group_norm.py
```diff
@@ -0,0 +1,209 @@
...
+
+cuda_group_norm_nhwc = GroupNormNHWC.apply
+cuda_group_norm_nhwc_one_pass = GroupNormOnePass.apply
+cuda_group_norm_nhwc_two_pass = GroupNormTwoPass.apply
+
+
+# We do not direct inherit from torch.nn.GroupNorm since several fusers don't
+# support inheritance. Extends:
+# https://github.com/pytorch/pytorch/blob/main/torch/nn/modules/normalization.py
+class GroupNormOpt(torch.nn.Module):
```
crcrpar, line 140: QQ: what would `Opt` mean here?

crcrpar (changes requested):
Could you please add a test that illustrates how to use this feature and the supported input shapes to https://github.com/NVIDIA/apex/tree/master/apex/contrib/test?

apex/contrib/group_norm/group_norm.py
alpha0422: This is only used during testing. Removed in this file.

apex/contrib/group_norm/group_norm.py
alpha0422: I would expect user only use `GroupNormOpt`. We export two different operators here mainly due to testing requirements. `GroupNormOpt` contains a heuristic to choose one pass or two pass algorithms. During testing, we want to test every algorithms.

apex/contrib/group_norm/group_norm.py
alpha0422: It's short for 'Optimized'. To make it clear, I removed `Opt`, so that users will only import `GroupNorm` from `apex`.

alpha0422:
> 

I just added a couple of tests to `apex/contrib/test/group_norm/test_group_norm.py`.

apex/contrib/test/group_norm/test_group_norm.py
```diff
@@ -0,0 +1,191 @@
...
+        dx_ref, dw_ref, db_ref = [_.grad.clone() for _ in [x, weight, bias]]
+        x.grad.zero_()
+        weight.grad.zero_()
+        bias.grad.zero_()
+        y_tst.backward(dy, retain_graph=True)
+        dx_tst, dw_tst, db_tst = [_.grad.clone() for _ in [x, weight, bias]]
+
+        # compare
+        fprop_close = torch.allclose(y_tst, y_ref, atol=4e-2, rtol=0)
+        self.assertTrue(fprop_close)
```
crcrpar, line 94:
```suggestion
        torch.testing.assert_close(y_tst, y_ref, atol=4e-2, rtol=0)
```

apex/contrib/test/group_norm/test_group_norm.py
```diff
@@ -0,0 +1,191 @@
...
+                print(y_ref[0, 0, 0, :peep])
+                print(y_tst[0, 0, 0, :peep])
+
+        dgrad_close = torch.allclose(dx_tst, dx_ref, atol=1e-2, rtol=0)
+        self.assertTrue(dgrad_close)
+        if not dgrad_close:
+            print('\nResult mis-match: dgrad')
+            if peep > 0:
+                print(dx_ref[0, 0, 0, :peep])
+                print(dx_tst[0, 0, 0, :peep])
```
crcrpar, line 107:
```suggestion
        torch.testing.assert_close(dx_tst, dx_ref, atol=1e-2, rtol=0)
```

crcrpar (commented):
`torch.testing.assert_close` would work well for numerical checks -- https://pytorch.org/docs/stable/testing.html

alpha0422:
Done, thanks for your suggestions.

apex/contrib/group_norm/group_norm.py
crcrpar: I know it's a mere nit picking but how about 
- making `__all__` to include only `GroupNorm`
- importing `cuda_group_norm_nhwc_(one|two)_pass` from `apex.contrib.group_norm.group_norm` in the test
?

apex/contrib/test/group_norm/test_group_norm.py
```diff
@@ -0,0 +1,153 @@
...
+# STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+import torch
+import unittest
+
+SKIP_TEST = None
+try:
+    from apex.contrib.group_norm import cuda_group_norm_nhwc_one_pass
+    from apex.contrib.group_norm import cuda_group_norm_nhwc_two_pass
```
crcrpar, line 24: The comment above basically means
```
    from apex.contrib.group_norm.group_norm import cuda_group_norm_nhwc_one_pass
    from apex.contrib.group_norm.group_norm import cuda_group_norm_nhwc_two_pass
```

crcrpar (approved):

apex/contrib/test/group_norm/test_group_norm.py
alpha0422: Make sense, I just modified and now we only export `GroupNorm` only.
Thanks for you valuable feedback on this, please see if there's further issues.

apex/contrib/test/group_norm/test_group_norm.py
```diff
@@ -0,0 +1,153 @@
...
+        super().setUp()
+        torch.manual_seed(seed)
+
+    def verify_group_norm(self,
+                          tst_func,
+                          N=32,
+                          C=128,
+                          H=256,
+                          W=256,
+                          G=32,
```
crcrpar, line 55: One last question: would we know the limitation of supported input shapes?
at glance, this is the only input shape tested

nekorobov: N/H/W do not have lower (except >0) and upper bound limitations. Number of groups G is limited to 32. Supported number of channels C are: 128, 256, 320, 448, 512, 640, 768, 896, 960, 1024, 1280, 1344, 1536, 1792, 1920, 2048, 2240, 2560, 2688, 3072, 3136, 3584, 4096. Two pass algorithm might automatically support some other channels as well, but it is not tested. One pass algorithm supports only channels mentioned above.

apex/contrib/test/group_norm/test_group_norm.py
```diff
@@ -0,0 +1,153 @@
...
+        x = x.to(memory_format=memory_format)
+        dy = .1 * torch.randn_like(x)
+        x.requires_grad_(True)
+
+        # forward pass
+        y_ref = ref_func(x, G, weight, bias, eps, act)
+        y_tst = tst_func(x, G, weight, bias, eps, act)
+        # backward pass
+        y_ref.backward(dy, retain_graph=True)
+        dx_ref, dw_ref, db_ref = [_.grad.clone() for _ in [x, weight, bias]]
```
crcrpar, line 84: this is my taste, but I'd not use `_` if it's really used
```suggestion
        dx_ref, dw_ref, db_ref = [t.grad.clone() for t in [x, weight, bias]]
```

apex/contrib/test/group_norm/test_group_norm.py
```diff
@@ -0,0 +1,153 @@
...
+        x.grad.zero_()
+        weight.grad.zero_()
+        bias.grad.zero_()
+        y_tst.backward(dy, retain_graph=True)
+        dx_tst, dw_tst, db_tst = [_.grad.clone() for _ in [x, weight, bias]]
```
crcrpar, line 89: ditto
```suggestion
        dx_tst, dw_tst, db_tst = [t.grad.clone() for t in [x, weight, bias]]
```

apex/contrib/test/group_norm/test_group_norm.py
crcrpar, line 1: could you add at least one test case that shows this group norm raises an exception for inputs with unsupported group values?

alpha0422: I tried to add some of the unsupported group values but find that it can raise SEGFAULT, which is not capturable by unittest. So I added explicit check in the Python code to use PyTorch version GroupNorm if it doesn't meet the conditions. Please take a look and see if that makes sense, thanks!

apex/contrib/csrc/group_norm/group_norm_nhwc_op.cpp
```diff
@@ -0,0 +1,243 @@
...
+    }                                                                 \
+  } while (0)
+
+#define CHECK_CUDA(x) \
+  AT_ASSERTM(x.type().is_cuda(), #x " must be a CUDA tensor")
+#define CHECK_CONTIGUOUS(x) \
+  AT_ASSERTM(x.is_contiguous(), #x " must be contiguous")
+#define CHECK_CHANNELS_LAST(x)                                \
+  AT_ASSERTM(x.is_contiguous(at::MemoryFormat::ChannelsLast), \
+             #x " must be channels last")
```
crcrpar, line 45: Macros that start with `AT_` have been sort of deprecated.

```suggestion
#define CHECK_CUDA(x) \
  TORCH_CHECK(x.type().is_cuda(), #x " must be a CUDA tensor")
#define CHECK_CONTIGUOUS(x) \
  TORCH_CHECK(x.is_contiguous(), #x " must be contiguous")
#define CHECK_CHANNELS_LAST(x)                                \
  TORCH_CHECK(x.is_contiguous(at::MemoryFormat::ChannelsLast), \
             #x " must be channels last")
```

apex/contrib/csrc/group_norm/group_norm_nhwc_op.cpp
```diff
@@ -0,0 +1,243 @@
...
+std::vector<torch::Tensor> group_norm_fwd(torch::Tensor input, int groups,
+                                          torch::Tensor weight,
+                                          torch::Tensor bias, float eps,
+                                          Algo algo, 
+                                          bool with_swish = false) {
+  if (!initialized) {
+    CHECK_CUDA_STATUS(cudaGetDeviceProperties(&props, 0));
+    initialized = true;
+  }
+  CHECK_NHWC_INPUT(input);
```
crcrpar, line 67: I prefer to explicitly throw an error if we're sure about the supported values of `groups`

```c++
  TORCH_CHECK(supported_groups_values.count(groups), "`groups` of 32 is only supported but ", groups, " is passed");
  CHECK_NHWC_INPUT(input);
```

apex/contrib/csrc/group_norm/group_norm_nhwc_op.cpp
```diff
@@ -0,0 +1,243 @@
...
+  CHECK_CUDA(x);       \
+  CHECK_CONTIGUOUS(x)
+#define CHECK_NHWC_INPUT(x) \
+  CHECK_CUDA(x);            \
+  CHECK_CHANNELS_LAST(x)
+
+enum Algo { OnePass, TwoPass };
+
+static bool initialized = false;
+static cudaDeviceProp props;
```
crcrpar, line 56: I prefer to spell out the supported `C` and `groups` and check the arguments of fwd and bwd like
```c++
static cudaDeviceProp props;
constexpr std::unordered_set<int> supported_c_values = {128, 256, 320, 448, 512, 640, 768, 896, 960, 1024, 1280, 1344, 1536, 1792, 1920, 2048, 2240, 2560, 2688, 3072, 3136, 3584, 4096};
constexpr std::unordered_set<int> supported_groups_values = {32};
```

apex/contrib/csrc/group_norm/group_norm_nhwc_op.cpp
```diff
@@ -0,0 +1,243 @@
...
+  if (!initialized) {
+    CHECK_CUDA_STATUS(cudaGetDeviceProperties(&props, 0));
+    initialized = true;
+  }
+  CHECK_NHWC_INPUT(input);
+  auto stream = at::cuda::getCurrentCUDAStream();
+
+  // Achieve group norm arguments
+  int n = input.size(0);
+  int c = input.size(1);
```
crcrpar, line 72: likewise `groups`
```c++
  int c = input.size(1);
  TORCH_CHECK(supported_c_values.count(c), "`c` of ", c, " is not included in ", supported_c_values);
```

apex/contrib/csrc/group_norm/group_norm_nhwc_op.cpp
```diff
@@ -0,0 +1,243 @@
...
+                                          torch::Tensor input, int groups,
+                                          torch::Tensor weight,
+                                          torch::Tensor bias, float eps,
+                                          Algo algo, 
+                                          bool with_swish = false) {
+  if (!initialized) {
+    CHECK_CUDA_STATUS(cudaGetDeviceProperties(&props, 0));
+    initialized = true;
+  }
+  CHECK_NHWC_INPUT(grad_output);
```
crcrpar, line 152: likewise fwd
```c++
 TORCH_CHECK(supported_groups_values.count(groups), "`groups` of 32 is only supported but ", groups, " is passed");
  CHECK_NHWC_INPUT(grad_output);
```

apex/contrib/csrc/group_norm/group_norm_nhwc_op.cpp
```diff
@@ -0,0 +1,243 @@
...
+  auto stream = at::cuda::getCurrentCUDAStream();
+
+  // Achieve group norm arguments
+  int n = input.size(0);
+  int c = input.size(1);
```
crcrpar, line 157: ditto

```c++
  int c = input.size(1);
  TORCH_CHECK(supported_c_values.count(c), "`c` of ", c, " is not included in ", supported_c_values);
```

apex/contrib/test/group_norm/test_group_norm.py
crcrpar: let's throw an error explicitly

alpha0422: Added test for `GroupNorm`, all the cases other than what Nikita mentioned will be forwarded to PyTorch implementation.

apex/contrib/csrc/group_norm/group_norm_nhwc_op.cpp
alpha0422: Thanks, all the checking has been added to the Python side and unsupported cases will be forwarded to PyTorch implementation.

apex/contrib/csrc/group_norm/group_norm_nhwc_op.cpp
alpha0422: Ditto.

apex/contrib/csrc/group_norm/group_norm_nhwc_op.cpp
alpha0422: Ditto.

apex/contrib/csrc/group_norm/group_norm_nhwc_op.cpp
alpha0422: Ditto.

apex/contrib/group_norm/group_norm.py
```diff
@@ -0,0 +1,258 @@
...
+# https://github.com/pytorch/pytorch/blob/main/torch/nn/modules/normalization.py
+class GroupNorm(torch.nn.Module):
+    """Optimized GroupNorm for NHWC layout with optional Swish/SiLU fusion.
+
+    There are two version of CUDA kernels under the hood: one pass and two
+    passes. This operator contains a simple heuristic to choose algorithm.
+
+    Limitations:
+
+    * Designed for 32 groups, though some other number of groups can also work;
```
crcrpar, line 135: Shouldn't 16 be mentioned here as well?

crcrpar: if this is intended for 16/32, then I'd mention those values and also mention that this will call pytorch based implementation for the unsupported groups values

crcrpar (approved):

apex/contrib/group_norm/group_norm.py
alpha0422: You're right, that makes sense and I just updated the doc for the module.

apex/contrib/csrc/group_norm/group_norm_nhwc_op.cpp
crcrpar: @alpha0422 That's great but still I think it'd be better to have a check in C++ side as well

felixdae:
great work!
i have a question. if i want to support more channels, what should i do

alpha0422:
> great work! i have a question. if i want to support more channels, what should i do

You can create a PR if you want to contribute and support more channels for this GroupNorm implementation. Meanwhile, this function will be moved into cuDNN, and more channels will be supported there.

GHGmc2:
> cuDNN

May I know the status of cuDNN support of GroupNorm?
Thanks!

PR #1699, opened 2023-07-25
Add type hints to distributed Adam optimizer

timmoon10:
This PR makes some minor stylistic changes to the distributed Adam optimizer:
- Add type hints for functions and member variables.
- Remove unused functions, mainly the deprecated synchronization logic for async collectives in `GradientBucket.sync_wait` and `ParameterBucket.sync_wait`,
- Make the `ParameterFragment` class a dataclass.

None of these changes should affect functionality.

### DIFF
--- a/apex/contrib/optimizers/distributed_fused_adam.py
+++ b/apex/contrib/optimizers/distributed_fused_adam.py
@@ -6,7 +6,7 @@ import inspect
 import io
 import itertools
 import threading
-from typing import Any, Callable, Iterable, List, Optional, Set, Tuple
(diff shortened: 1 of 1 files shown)


crcrpar (commented):
LGTM. how about applying a formatter such as black with the line length of maybe 100 or 120 rather than 80 to only this file as well?

we can later "ignore" purely stylistic changes by adding a file called `.git-blame-ignore-revs` e.g. https://github.com/pytorch/pytorch/blob/3eef86dbf47be29a813fc51e4248bea79afd4252/.git-blame-ignore-revs

PR #1714, opened 2023-08-23
Add the warning of distributed_fused_adam low bucket usage

shjwudp:
If **not** set a fit bucket size in the distributed optimizer, memory waste will result. Memory loss is sometimes high but invisible, for example, a 10GB memory penalty for each GPU on a misconfigured gpt-7b. I think reporting a warning when the bucket utilization is low is a solution, and I submitted my code as reference.

### DIFF
--- a/apex/contrib/optimizers/distributed_fused_adam.py
+++ b/apex/contrib/optimizers/distributed_fused_adam.py
@@ -393,6 +393,8 @@ class DistributedFusedAdam(torch.optim.Optimizer):
             self.shard_size: int = shard_size
             # Size of the filled region in the bucket
             self.filled_size: int = 0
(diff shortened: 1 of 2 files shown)


timmoon10 (changes requested):
This would be a very useful check. I just have a suggestion to handle the case where the parameters are initialized in multiple stages, e.g. if the user uses `init_params_bucket` to manually configure buckets.

shjwudp:
Hi Tim, It's great to see your reply! I appreciate your help in correcting it; thanks!

timmoon10 (approved):
LGTM. Thanks!

crcrpar (approved):

PR #1715, opened 2023-08-24
Massively reduce LayerNorm/RMSNorm GPU memory usage in modern networks by tricking torch autograd

RuiWang1998:
Hi,

# Content/motivation
This modifies the fused layernorm/rmsnorm implmentation s.t. the output tensor is saved and the input is supposedly freed. The motivation comes from the observation that the output is going to be saved somewhere anyway. For example, in pre-norm Transformers (GPT, Palm, T5, and many others), the normalization is directly followed by a linear projection either in attention or in mlp.  In post-norm models, the output is additionally used in residual connection directly. However, in both scenarios, the input of the normalization is left unused, making its original usage here pretty redundant as we can save the output tensor that is going to be saved anyway.

# Effect
We observe a substantial reduction in memory usage when training large transformer models: a 80GB-run on one A100 now only requires 70GB memory, which is 1/8 of a reduction in total memory cost. We use a configuration where fewer normalizations are applied and the benefits will be more pronounced in canonical models.

# Notes on numerical precision
Since we save the output instead of the input tensor, the _normalied_ tensor $\hat{x}$ needed in the backward is recomputed from output, i.e., $\hat{x}=(y-\beta)/\gamma$ instead of $\hat{x}=\frac{x-E(x)}{\sqrt{VAR(x)-E(x)}}$. As a result we are seeing numerical differences between the current implementation and the proposed change. However, note that the reduction part of the computation happens in the forward pass s.t. we would not have any stability issue whatsoever (we also clamp by magnitude of $\gamma$ for division).

Currently we are failing at two test runs due to numerical discrepancies when using `float16`:

```bash
======================================================================
FAIL: test_autocast_fused_layer_norm_float16_elementwise_affine_False_cuda_float16 (__main__.TestFusedLayerNormCUDA.test_autocast_fused_layer_norm_float16_elementwise_affine_False_cuda_float16)
----------------------------------------------------------------------
  File "/home/rui/anaconda3/envs/layer-norm/lib/python3.11/site-packages/torch/testing/_internal/common_utils.py", line 2081, in wrapper
    method(*args, **kwargs)
  File "/home/rui/anaconda3/envs/layer-norm/lib/python3.11/site-packages/torch/testing/_internal/common_device_type.py", line 401, in instantiated_test
    result = test(self, **param_kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/rui/opensource/apex/tests/L0/run_fused_layer_norm/test_fused_layer_norm.py", line 234, in test_autocast_fused_layer_norm
    torch.testing.assert_close(native_x.grad, fused_x.grad, **tols, check_dtype=False)
  File "/home/rui/anaconda3/envs/layer-norm/lib/python3.11/site-packages/torch/testing/_comparison.py", line 1511, in assert_close
    raise error_metas[0].to_error(msg)
AssertionError: Tensor-likes are not close!

Mismatched elements: 20 / 8192 (0.2%)
Greatest absolute difference: 4.57763671875e-05 at index (1, 21, 12) (up to 1e-05 allowed)
Greatest relative difference: 0.01692047377326565 at index (7, 25, 8) (up to 0.001 allowed)

======================================================================
FAIL: test_autocast_fused_layer_norm_float16_elementwise_affine_True_cuda_float16 (__main__.TestFusedLayerNormCUDA.test_autocast_fused_layer_norm_float16_elementwise_affine_True_cuda_float16)
----------------------------------------------------------------------
  File "/home/rui/anaconda3/envs/layer-norm/lib/python3.11/site-packages/torch/testing/_internal/common_utils.py", line 2081, in wrapper
    method(*args, **kwargs)
  File "/home/rui/anaconda3/envs/layer-norm/lib/python3.11/site-packages/torch/testing/_internal/common_device_type.py", line 401, in instantiated_test
    result = test(self, **param_kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/rui/opensource/apex/tests/L0/run_fused_layer_norm/test_fused_layer_norm.py", line 234, in test_autocast_fused_layer_norm
    torch.testing.assert_close(native_x.grad, fused_x.grad, **tols, check_dtype=False)
  File "/home/rui/anaconda3/envs/layer-norm/lib/python3.11/site-packages/torch/testing/_comparison.py", line 1511, in assert_close
    raise error_metas[0].to_error(msg)
AssertionError: Tensor-likes are not close!

Mismatched elements: 20 / 8192 (0.2%)
Greatest absolute difference: 4.57763671875e-05 at index (1, 21, 12) (up to 1e-05 allowed)
Greatest relative difference: 0.01692047377326565 at index (7, 25, 8) (up to 0.001 allowed)
```

### DIFF
--- a/apex/contrib/layer_norm/layer_norm.py
+++ b/apex/contrib/layer_norm/layer_norm.py
@@ -29,9 +29,9 @@ class FastLayerNormFN(torch.autograd.Function):
         dy = dy.contiguous()  # this happens!
         x_or_y_mat, gamma, mu, rsigma, beta = ctx.saved_tensors
         dymat = dy.view(x_or_y_mat.shape)
-        dxmat, dgamma, dbeta, _, _ = fast_layer_norm.ln_bwd(dymat, x_or_y_mat, mu, rsigma, gamma, beta)
+        dxmat, dgamma, dbeta, _, _ = fast_layer_norm.ln_bwd(dymat, x_or_y_mat, mu, rsigma, gamma, beta, ctx.memory_efficient)
         dx = dxmat.view(ctx.x_shape)
-        return dx, dgamma, dbeta, None
+        return dx, dgamma, dbeta, None, None
 
 
 def _fast_layer_norm(x, weight, bias, epsilon, memory_efficient):
--- a/apex/contrib/layer_norm/layer_norm.py
+++ b/apex/contrib/layer_norm/layer_norm.py
@@ -7,40 +7,44 @@ import fast_layer_norm
 
 class FastLayerNormFN(torch.autograd.Function):
     @staticmethod
-    def forward(ctx, x, gamma, beta, epsilon):
+    def forward(ctx, x, gamma, beta, epsilon, memory_efficient):
+        ctx.x_shape = x.shape
+        ctx.memory_efficient = memory_efficient
+
         x = x.contiguous()
         gamma = gamma.contiguous()
         beta = beta.contiguous()
         hidden_size = gamma.numel()
         xmat = x.view((-1, hidden_size))
         ymat, mu, rsigma = fast_layer_norm.ln_fwd(xmat, gamma, beta, epsilon)
-        ctx.save_for_backward(x, gamma, mu, rsigma)
+        if ctx.memory_efficient:
+            ctx.save_for_backward(ymat, gamma, None, rsigma, beta)
+        else:
+            ctx.save_for_backward(xmat, gamma, mu, rsigma, None)
         return ymat.view(x.shape)
 
     @staticmethod
     def backward(ctx, dy):
         # assert dy.is_contiguous()
         dy = dy.contiguous()  # this happens!
-        x, gamma, mu, rsigma = ctx.saved_tensors
-
-        hidden_size = gamma.numel()
-        xmat = x.view((-1, hidden_size))
-        dymat = dy.view(xmat.shape)
-        dxmat, dgamma, dbeta, _, _ = fast_layer_norm.ln_bwd(dymat, xmat, mu, rsigma, gamma)
-        dx = dxmat.view(x.shape)
+        x_or_y_mat, gamma, mu, rsigma, beta = ctx.saved_tensors
+        dymat = dy.view(x_or_y_mat.shape)
+        dxmat, dgamma, dbeta, _, _ = fast_layer_norm.ln_bwd(dymat, x_or_y_mat, mu, rsigma, gamma, beta)
+        dx = dxmat.view(ctx.x_shape)
         return dx, dgamma, dbeta, None
 
 
-def _fast_layer_norm(x, weight, bias, epsilon):
-    args = _cast_if_autocast_enabled(x, weight, bias, epsilon)
+def _fast_layer_norm(x, weight, bias, epsilon, memory_efficient):
+    args = _cast_if_autocast_enabled(x, weight, bias, epsilon, memory_efficient)
     with torch.cuda.amp.autocast(enabled=False):
         return FastLayerNormFN.apply(*args)
 
 
 class FastLayerNorm(torch.nn.Module):
-    def __init__(self, hidden_size, eps=1e-5):
+    def __init__(self, hidden_size, eps=1e-5, memory_efficient=False):
         super().__init__()
         self.epsilon = eps
+        self.memory_efficient = memory_efficient
         self.weight = torch.nn.Parameter(torch.empty(hidden_size))
         self.bias = torch.nn.Parameter(torch.empty(hidden_size))
         self.reset_parameters()
@@ -50,4 +54,4 @@ class FastLayerNorm(torch.nn.Module):
         init.zeros_(self.bias)
 
     def forward(self, x):
-        return _fast_layer_norm(x, self.weight, self.bias, self.epsilon)
+        return _fast_layer_norm(x, self.weight, self.bias, self.epsilon, self.memory_efficient)
--- a/apex/contrib/test/layer_norm/test_fast_layer_norm.py
+++ b/apex/contrib/test/layer_norm/test_fast_layer_norm.py
@@ -1,3 +1,4 @@
+import itertools
 import unittest
 
 import torch
@@ -106,7 +107,7 @@ def benchmark_(S, B, hidden_size, itype, wtype, runs=100):
 
         timer.start()
         for r in range(runs):
-            dx, dgamma, dbeta, dbp, dgp = fln.ln_bwd(dz, x, mu, rsigma, gamma)
+            dx, dgamma, dbeta, dbp, dgp = fln.ln_bwd(dz, z, mu, rsigma, gamma, beta, True)
         timer.stop()
         timer.sync()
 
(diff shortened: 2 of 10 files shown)


crcrpar (commented):
First of all, thank you very much for sharing your awesome finding and implementation. This looks so exciting! 

I however think it'd be better if we can switch the behavior via an argument of the two normalizations given the numerical stability, e.g. `LayerNorm(..., memory_efficient)`.

What do you think?

RuiWang1998:
Hi @crcrpar,
 Thanks for the comment!

It's very doable to add an additional flag to the class. However I believe this could be the new _default_ behavior simply due to its benefits. For the stability issue however, I would argue that it's as stable as the original since the reduction happens in the forward pass and we do not modify that. 

The difference between the two is small enough that the memory cost of the original in my opinion pales a bit.

Nevertheless if you still think it is better to add an additional flag, I'll be very glad to do that!

crcrpar:
I hear you and appreciate this improvement very much, but I want to be conservative here. These normalization layers are used by a certain number of models out there. Even if we make this path the default behavior in the end, I do think we should have some period to let the users get aware of the behavioral change is coming by adding an argument and an appropriate warning message

RuiWang1998:
Hi @crcrpar ,

Yeah sure that makes sense!
 I'll add a flag very soon for this!

RuiWang1998:
Hi @crcrpar ,

I believe these two commits allows an argument for switching behaviors!

Do we need a warning somehow now?

On a side note, I was trying to do the same thing to `FastLayerNorm` but keep getting weird bugs even doing something as simple as changing the name of `params.x` to `params.z` for `FwdParams`. Anyways I would really appreciate it if you take a look at it and give me some pointers here: https://github.com/RuiWang1998/apex/blob/f658a37deebf0c5b11402326b8e24cce3b58b04a/apex/contrib/csrc/layer_norm/ln_bwd_kernels.cuh#L96

Best,
Rui

RuiWang1998:
Hi @crcrpar ,

It seems I have successfully get `fast_layer_norm` version to work as well! It turned out to be just some cache issues during compiling.

apex/contrib/test/layer_norm/test_fast_layer_norm.py
```diff
@@ -184,7 +188,7 @@ def _test_impl(S, B, hidden_size, itype, wtype, ctype=fp32):
     print(f"db: relerr={re_db:.4e} mse={mse_db:.4e}")
 
     def check_err(x, relerr):
-        tol = 1e-3 if x.dtype == torch.float16 else 5e-6
+        tol = 2e-2 if x.dtype in (torch.float16, torch.bfloat16) else 5e-6
```
crcrpar, line 191: is `2e-2` for gradients check in `mem_eff` pass?

RuiWang1998: It passes after I change it to this.

For `float16`, it works with `2e-3` but `bfloat16` somehow results in higher error.

RuiWang1998: All these are for `dg` For all other values, the original config passes no problem.

eqy:
Since this is changing the computation done on the backward pass somewhat, is there any intuition or data on the expected performance (speed) difference compared to the current implementation?

RuiWang1998:
Hi @eqy ,

Thanks for the comment! 

As you can probably see, the total number of FLOPs is almost identical and the read/write tensors have almost the same shape albeit slightly different. For this reason, we believe they should be almost the same performance speed wise, which is consistent with what we have been observing.

RuiWang1998:
Hi @crcrpar ,

I additionally moved `memory_efficient` to template, which should in theory make it costless.

tests/L0/run_fused_layer_norm/test_fused_layer_norm.py
```diff
@@ -123,87 +129,87 @@ def _test_fused_rms_norm(
...
                                 fwd_thresholds=dict(rtol=1.6e-2, atol=3e-4), bwd_thresholds=dict(rtol=1.6e-2, atol=3e-2))
 
     @common_utils.parametrize(
-        "dtype, elementwise_affine",
-        list(product(autocast_dtypes, (True, False)))
+        "dtype, elementwise_affine, memory_efficient",
+        list(product(autocast_dtypes, (True, False), (True, False)))
     )
-    def test_autocast_fused_layer_norm(self, dtype, elementwise_affine):
+    def test_autocast_fused_layer_norm(self, dtype, elementwise_affine, memory_efficient):
```
crcrpar, line 212: on V100 I saw the following failures

```
FAILED test_fused_layer_norm.py::TestFusedLayerNormCUDA::test_autocast_fused_layer_norm_float16_elementwise_affine_False_memory_efficient_True_cuda_float16 - AssertionError: Tensor-likes are not close!
FAILED test_fused_layer_norm.py::TestFusedLayerNormCUDA::test_autocast_fused_layer_norm_float16_elementwise_affine_True_memory_efficient_True_cuda_float16 - AssertionError: Tensor-likes are not close!
```

RuiWang1998: Hi,

Are those the two tests referenced at the initial post? https://github.com/NVIDIA/apex/pull/1715#issue-1864532937

I have not modify those two tests yet.

RuiWang1998:
Hi @crcrpar ,

I have relaxed the test tolerance a tiny bit for the half precision to pass.

Best.

apex/contrib/layer_norm/layer_norm.py
```diff
@@ -7,40 +7,44 @@
...
-        x, gamma, mu, rsigma = ctx.saved_tensors
-
-        hidden_size = gamma.numel()
-        xmat = x.view((-1, hidden_size))
-        dymat = dy.view(xmat.shape)
-        dxmat, dgamma, dbeta, _, _ = fast_layer_norm.ln_bwd(dymat, xmat, mu, rsigma, gamma)
-        dx = dxmat.view(x.shape)
+        x_or_y_mat, gamma, mu, rsigma, beta = ctx.saved_tensors
+        dymat = dy.view(x_or_y_mat.shape)
+        dxmat, dgamma, dbeta, _, _ = fast_layer_norm.ln_bwd(dymat, x_or_y_mat, mu, rsigma, gamma, beta)
```
crcrpar, line 32: doesn't `ln_bwd` need `memory_efficient`?
```suggestion
        dxmat, dgamma, dbeta, _, _ = fast_layer_norm.ln_bwd(dymat, x_or_y_mat, mu, rsigma, gamma, beta, ctx.memory_efficient)
```

apex/contrib/layer_norm/layer_norm.py
```diff
@@ -7,40 +7,44 @@
...
+        dx = dxmat.view(ctx.x_shape)
         return dx, dgamma, dbeta, None
```
crcrpar, line 34: the number of return values shouldn't be 5?
```suggestion
        return dx, dgamma, dbeta, None, None
```

crcrpar (changes requested):

crcrpar:
I had been on vacation, excuse me for the delay

RuiWang1998:
Hi @crcrpar ,

Thanks so much! I don't know how I missed those. Anyways they are fixed now.

Best

crcrpar (approved):

PR #1716, opened 2023-08-25
Include format version in distopt checkpoints

timmoon10:
https://github.com/NVIDIA/apex/pull/1704 introduced a bug where the distributed optimizer fails when loading old checkpoints in the deprecated v1 format. This PR includes the checkpoint format in the checkpoint. If the distributed optimizer can't find the format when loading a checkpoint, it falls back to v1. This should help with backwards compatibility if we change the format again in the future.

### DIFF
--- a/apex/contrib/optimizers/distributed_fused_adam.py
+++ b/apex/contrib/optimizers/distributed_fused_adam.py
@@ -2187,9 +2187,8 @@ class DistributedFusedAdam(torch.optim.Optimizer):
         returns None on non-root ranks.
 
         Arguments:
-            state_dict_format (optional): Tag for state dict format
-                (default: 2). Typically used to output in deprecated
-                formats.
+            state_dict_format (optional): Tag for custom or deprecated
+                state dict format.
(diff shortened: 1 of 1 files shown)


apex/contrib/optimizers/distributed_fused_adam.py
```diff
@@ -2178,31 +2178,46 @@ def _local_step_with_param_remainders(
     def state_dict(
         self,
         *,
-        v1_format: bool = False,
+        state_dict_format: Optional[Any] = None,
```
crcrpar, line 2181: I'm not quite sure but shouldn't the annotation be

```suggestion
        state_dict_format: Optional[int] = None,
```
?

crcrpar (approved):

apex/contrib/optimizers/distributed_fused_adam.py
timmoon10: My thought is that the format could also be a `str`, but I may be overthinking it.

PR #1718, opened 2023-08-28
Adjusting test for ONNX opset 18 (now default)

borisfom:
Fixes errors in unit test that happened when upset 18 was used.

### DIFF
--- a/tests/L0/run_fused_layer_norm/test_fused_layer_norm.py
+++ b/tests/L0/run_fused_layer_norm/test_fused_layer_norm.py
@@ -270,7 +270,6 @@ class TestFusedLayerNorm(common_utils.TestCase):
                                                       input_names=['x_in'],
                                                       opset_version=18,
         )
-        print(onnx_str)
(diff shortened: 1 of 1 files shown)


tests/L0/run_fused_layer_norm/test_fused_layer_norm.py
```diff
@@ -268,9 +268,11 @@ def _verify_export(self, fused, fused_x):
         # check that export() is working
         onnx_str = torch.onnx.export_to_pretty_string(fused, (fused_x,),
                                                       input_names=['x_in'],
+                                                      opset_version=18,
         )
+        print(onnx_str)
```
crcrpar, line 273: I don't think this is needed

crcrpar (approved):

tests/L0/run_fused_layer_norm/test_fused_layer_norm.py
crcrpar, line 273:
```suggestion
```

PR #1721, opened 2023-09-01
Distributed optimizer support for multiple dtypes

timmoon10:
This PR adds logic so that the parameters can be configured with different dtypes for the grad reduce-scatters and param all-gathers. I have two NeMo use-cases in mind:
- For GPT, most grads can be reduced in BF16 but embedding grads need to be reduced in FP32 to avoid learning issues.
- For FP8 support, weight matrices can be stored in FP8 while most other parameters (e.g. biases, layernorm params, embeddings) are in BF16. We would like to handle FP8 and BF16 param all-gathers in the same optimizer.

This also includes changes from https://github.com/NVIDIA/apex/pull/1719, which returns the state dict on all ranks and not just rank 0. We can either merge that first and rebase, or merge this and close https://github.com/NVIDIA/apex/pull/1719.


apex/contrib/test/optimizers/test_dist_adam.py
```diff
@@ -1,17 +1,13 @@
...
 
 import torch
 from torch.testing._internal import common_utils
 
-SKIP_TEST = None
-try:
-    from apex.contrib.optimizers.distributed_fused_adam import DistributedFusedAdam
-except ImportError as e:
-    SKIP_TEST = e
+from apex.contrib.optimizers.distributed_fused_adam import DistributedFusedAdam
```
crcrpar, line 10: I'm not sure why it's okay to remove the guard here off the top of my head?

crcrpar (approved):
lgtm

apex/contrib/test/optimizers/test_dist_adam.py
timmoon10: On second thought, it would be good to restore the guard in case Apex has been built without the distopt CUDA kernels.

PR #1723, opened 2023-09-06
Distributed optimizer infrastructure for FP8 parameters

timmoon10:
This PR does some refactoring that will enable distributed optimizer support for FP8 parameters in NeMo. It adds the option to do parameter all-gathers in integer dtypes and adds two member functions - `_check_params_shard_dtypes` and `_param_copy_fragments` - to handle casting into and out of the all-gather buffer. For now these functions will either do a direct cast for floating-point dtypes or copy the most significant bytes for other dtypes. I plan to override these functions in the [NeMo derived class](https://github.com/NVIDIA/NeMo/blob/03465b0d5224a5e82f259d49ba69f2d02b036d94/nemo/core/optim/distributed_adam.py#L51) so that it casts to FP8, performs the all-gather in UINT8, and unpacks into a custom FP8 tensor class.

This PR depends on https://github.com/NVIDIA/apex/pull/1719 and https://github.com/NVIDIA/apex/pull/1721.

### DIFF
--- a/apex/contrib/optimizers/distributed_fused_adam.py
+++ b/apex/contrib/optimizers/distributed_fused_adam.py
@@ -1081,7 +1081,7 @@ class DistributedFusedAdam(torch.optim.Optimizer):
             raise TypeError(
                 "Expected input types are "
                 "[int, int] or [DistributedFusedAdam.ParameterFragment], "
-                "but found {[type(arg).__name__ for arg in args]}"
+                f"but found {[type(arg).__name__ for arg in args]}"
             )
         return self.param_groups[param_group_id]["params"][param_id]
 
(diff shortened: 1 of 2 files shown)


apex/contrib/optimizers/distributed_fused_adam.py
```diff
@@ -1032,6 +1052,39 @@ def parameters(self) -> Iterable[torch.nn.Parameter]:
...
+            param_id = args[1]
+        elif len(args) == 1 and isinstance(args[0], self.ParameterFragment):
+            fragment = args[0]
+            param_group_id = fragment.param_group_id
+            param_id = fragment.param_id
+        else:
+            raise TypeError(
+                "Expected input types are "
+                "[int, int] or [DistributedFusedAdam.ParameterFragment], "
+                "but found {[type(arg).__name__ for arg in args]}"
```
crcrpar, line 1084: I'm not sure but would this happen to miss `f`?
```suggestion
                f"but found {[type(arg).__name__ for arg in args]}"
```

crcrpar (approved):

Issue #1728, opened 2023-09-13
FusedAdam doesn't allocate master weights for bfloat16


cbcase:
If you look at https://github.com/NVIDIA/apex/blob/master/apex/optimizers/fused_adam.py#L184, the FusedAdam code doesn't allocate master weights when the parameter dtype is bfloat16, even if you set `master_weights=True` (and subsequently no master weights are passed to the kernel at https://github.com/NVIDIA/apex/blob/master/apex/optimizers/fused_adam.py#L235).

Is there a specific reason for this, or is it simply an oversight?

(Seeing as bfloat has fewer mantissa bits than fp16, fp32 master weights are even more important for bfloat16 -- though, really, they are a necessity for both.)

Aidyn-A:
@crcrpar, do you remember why we skip `master_weights` for `bfloat16`?

crcrpar:
> @crcrpar, do you remember why we skip `master_weights` for `bfloat16`? 

I'm unsure but I vaguely remember there wasn't master weights usage in fused adam so it feels like more of a fallout when fp16 - fp32 things were added

PR #1731, opened 2023-09-22
Use master weights for bfloat16 FusedAdam when master_weights=True

cbcase:
As mentioned in #1728, the FusedAdam optimizer ignores `master_weights=True` for bfloat16 parameters. This PR fixes that oversight. I have confirmed that the behavior now matches a "by hand" implementation of master weights (hand-copying) along with vanilla torch.optim.AdamW on the fp32 copy.


cbcase:
Ping @minitu, looks like you added this support originally -- could you take a look? Thanks

minitu:
LGTM, we only looked at adding master weights for FP16 AMP at the time of the original PR.
@crcrpar Could you review this as well?

PR #1740, opened 2023-10-12
Loop through all available engines for cuDNN heuristics search

minitu:
This PR updates the cuDNN heuristics search to loop through all available engines and find the first successful one for a given operation graph, instead of trying a set number of times.

### DIFF
--- a/apex/contrib/csrc/conv_bias_relu/conv_bias_relu.cpp
+++ b/apex/contrib/csrc/conv_bias_relu/conv_bias_relu.cpp
@@ -143,6 +143,7 @@ cudnn_frontend::ExecutionPlan& getOrCreatePlan(cudnnHandle_t handle_,
     DEBUG_CUDNN_MSG(log_buf, "Found plan in cache");
     return it->second;
   } else {
+    DEBUG_CUDNN_MSG(log_buf, "No plan in cache");
     if (use_heuristic) {
       // TODO: confirm which mode to use
       auto heuristics = cudnn_frontend::EngineHeuristicsBuilder()
@@ -168,7 +169,6 @@ cudnn_frontend::ExecutionPlan& getOrCreatePlan(cudnnHandle_t handle_,
         }
       }
     } else {
-      DEBUG_CUDNN_MSG(log_buf, "No plan in cache");
       // How many engines support this operation graph ?
       auto total_engines = opGraph.getEngineCount();
       DEBUG_CUDNN_MSG(log_buf, opGraph.describe() << " has " << total_engines << " engines.");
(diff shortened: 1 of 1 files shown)


apex/contrib/csrc/conv_bias_relu/conv_bias_relu.cpp
```diff
@@ -149,8 +149,8 @@ cudnn_frontend::ExecutionPlan& getOrCreatePlan(cudnnHandle_t handle_,
         .setOperationGraph(opGraph)
         .setHeurMode(CUDNN_HEUR_MODE_INSTANT)
         .build();
-      // try 3 times for now as WAR for no heuristic training
-      int max_tries = 3, count = 0;
+      // try 10 times for now as WAR for no heuristic training
+      int max_tries = 10, count = 0;
```
eqy, line 153: Could we flip this around, where instead the number of "successful" engines is capped rather than the number of "failures?" I don't see a reason why trying to limit  the number of failing engines is helpful---we don't do this in upstream code for vanilla cuDNN convs.

apex/contrib/csrc/conv_bias_relu/conv_bias_relu.cpp
```diff
@@ -143,49 +143,53 @@ cudnn_frontend::ExecutionPlan& getOrCreatePlan(cudnnHandle_t handle_,
...
-    for (auto it = std::begin(knobs); it != std::end(knobs); ++it) {
-      DEBUG_CUDNN_MSG(log_buf, it->describe());
-    }
-    if (knobs.begin() != knobs.end()) {
-      DEBUG_CUDNN_MSG(log_buf, "Updated knob choice");
-      knobs.begin()->setChoice(knobs.begin()->getMinValue() + 1);
-      DEBUG_CUDNN_MSG(log_buf, knobs.begin()->describe());
-    }
+    } else {
+      DEBUG_CUDNN_MSG(log_buf, "No plan in cache");
```
eqy, line 171: Nit: this is isn't from this PR, but I'm guessing that this should be moved outside this branch to just after the `return it->second` line, as both this case and the `use_heuristic` branch are for when there is no plan in the cache.

eqy (approved):

eqy:
Also noting that the `plan_cache` should be made `thread_local`, but this doesn't have to be done here and isn't necessary if we assume no multithreaded interactions (expected if we are only calling this layer from Python).

apex/contrib/csrc/conv_bias_relu/conv_bias_relu.cpp
minitu: Moved accordingly

PR #1742, opened 2023-10-17
Cleanup usage of `self.assertTrue(torch.allclose(...`

eqy:
Cleanup so that we have better signal (i.e., are numerical mismatches small or large?) in logs of failing tests.

Some files appear to be mangled because they were in Windows? format somehow? I've converted them to unix in this PR.

CC @crcrpar @Fuzzkatt


crcrpar (approved):
Thank you very much!

PR #1743, opened 2023-10-19
Fix `rtol` in `assert_close` cleanup

eqy:
Fix broken call in #1742
CC @crcrpar


crcrpar (approved):
Thanks. Sorry for having missed this

PR #1751, opened 2023-11-20
Avoid `.contiguous()` in fused RoPE

yaox12:
Support different memory formats to avoid `.contiguous` calls.


crcrpar (approved):
looks good

PR #1767, opened 2024-01-12
Update test_transducer_joint.py

nWEIdia:
Increase tolerance to workaround unit test failures 

torch.testing.assert_close(f_grad_ref, f_grad_tst, atol=1e-5, rtol=1e-5)
Mismatched elements: 1 / 205636 (0.0%)
Greatest absolute difference: 3.0517578125e-05 at index (3, 27, 390) (up to 1e-05 allowed) Greatest relative difference: 0.000492095947265625 at index (3, 27, 390) (up to 1e-05 allowed)

 torch.testing.assert_close(g_grad_ref, g_grad_tst, atol=1e-4, rtol=1e-4)
Mismatched elements: 1 / 51200 (0.0%)
Greatest absolute difference: 0.0009765625 at index (0, 15, 280) (up to 0.0001 allowed) Greatest relative difference: 0.0008397102355957031 at index (0, 15, 280) (up to 0.0001 allowed)

### DIFF
--- a/apex/contrib/test/transducer/test_transducer_joint.py
+++ b/apex/contrib/test/transducer/test_transducer_joint.py
@@ -119,8 +119,8 @@ class TransducerJointTest(unittest.TestCase):
         g_grad_tst = self.g_tst.grad
 
         torch.testing.assert_close(h_ref, h_tst, atol=1e-5, rtol=1e-5)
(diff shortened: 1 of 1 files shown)


crcrpar (approved):
Would all the cases need this change, out of curiosity?

nWEIdia:
> Would all the cases need this change, out of curiosity?

These two mismatches were from two different tests,  test_transducer_joint_pack_relu, and test_transducer_joint_vec_pack.

PR #1772, opened 2024-01-23
[64bit indexing][Adam] Add annotation to large tensor test

eqy:
Memory reserved reaches ~56GiB here so going with 60GiB to be "safe"

### DIFF
--- a/tests/L0/run_optimizers/test_adam.py
+++ b/tests/L0/run_optimizers/test_adam.py
@@ -6,6 +6,7 @@ import unittest
 import torch
 import torch.nn.functional as F
 from torch import nn
+from torch.testing._internal.common_device_type import largeTensorTest
 
 try:
     import apex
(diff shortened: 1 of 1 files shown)


tests/L0/run_optimizers/test_adam.py
```diff
@@ -232,6 +233,7 @@ def testNative(self):
             
             self.model_.load_state_dict(copy.deepcopy(self.model.state_dict()))
 
+    @largeTensorTest('60GB', 'cuda')
```
nWEIdia, line 236: How about testing on a 2x 40GB configuration where each process had at most 40GB? These tests would be auto-skipped, correct?

eqy: Yes, it should be current-device only IIUC

nWEIdia (approved):
LGTM, thanks for the fix!

PR #1794, opened 2024-04-11
Enhance Distributed Fused Adam

alpha0422:
This PR enhances distributed fused adam by:

1. Support NHWC layout (required by some Conv related models, e.g. Diffusion models);
2. Fix the gradient clipping bug;
3. Support CUDA graph;

@timmoon10 @crcrpar Please help review, thanks.

### DIFF
--- a/apex/contrib/csrc/optimizers/multi_tensor_distopt_adam_kernel.cu
+++ b/apex/contrib/csrc/optimizers/multi_tensor_distopt_adam_kernel.cu
@@ -194,6 +194,170 @@ struct DistAdamFunctor
   }
 };
 
+/* Multi-tensor Adam with CUDA Graph Support
+ *
+ * Updates params in-place and outputs a copy with a desired datatype.
+ */
+template <typename T, typename GRAD_T, typename PARAM_OUT_T>
+struct DistAdamCapturableFunctor
+{
+  // Vectorized local compute
+  __device__ __forceinline__ static void local_step(
+    T p[ILP],
+    T m[ILP],
+    T v[ILP],
+    const GRAD_T g[ILP],
+    const float grad_scale,
+    const float beta1,
+    const float beta2,
+    const float beta1_correction,
+    const float beta2_correction,
+    const float eps,
+    const float lr,
+    adamMode_t mode,
+    const float weight_decay) {
+    if (mode == ADAM_MODE_0) { // L2
+#pragma unroll
+      for (int ii = 0; ii < ILP; ii++) {
+        float scaled_grad = (g[ii] * grad_scale) + (weight_decay * p[ii]);
+        float next_m = lerp(beta1, scaled_grad, m[ii]);
+        float next_v = lerp(beta2, scaled_grad*scaled_grad, v[ii]);
+        float next_m_unbiased = next_m / beta1_correction;
+        float next_v_unbiased = next_v / beta2_correction;
+        float denom = sqrtf(next_v_unbiased) + eps;
+        float update = next_m_unbiased / denom;
+        m[ii] = next_m;
+        v[ii] = next_v;
+        p[ii] -= lr * update;
+      }
+    } else { // weight decay
+#pragma unroll
+      for (int ii = 0; ii < ILP; ii++) {
+        float scaled_grad = g[ii] * grad_scale;
+        float next_m = lerp(beta1, scaled_grad, m[ii]);
+        float next_v = lerp(beta2, scaled_grad*scaled_grad, v[ii]);
+        float next_m_unbiased = next_m / beta1_correction;
+        float next_v_unbiased = next_v / beta2_correction;
+        float denom = sqrtf(next_v_unbiased) + eps;
+        float update = (next_m_unbiased / denom) + (weight_decay * p[ii]);
+        m[ii] = next_m;
+        v[ii] = next_v;
+        p[ii] -= lr * update;
+      }
+    }
+  }
+
+  __device__ __forceinline__ void operator()(
+    int chunk_size,
+    volatile int* noop_gmem,
+    TensorListMetadata<5>& tl,
+    const float* grad_scale_ptr,
+    const float beta1,
+    const float beta2,
+    const int* step,
+    const int bias_correction,
+    const float eps,
+    const float* lr,
+    adamMode_t mode,
+    const float weight_decay) const
+  {
+    if(*noop_gmem == 1)
+      return;
+
+    float beta1_correction = 1.0f, beta2_correction = 1.0f;
+    if (bias_correction == 1) {
+      beta1_correction = 1 - pow(beta1, *step);
+      beta2_correction = 1 - pow(beta2, *step);
+    }
+
+    int tensor_loc = tl.block_to_tensor[blockIdx.x];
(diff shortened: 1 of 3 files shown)


apex/contrib/csrc/optimizers/multi_tensor_distopt_adam_kernel.cu
```diff
@@ -194,6 +194,170 @@ struct DistAdamFunctor
...
+    if (bias_correction == 1) {
+      beta1_correction = 1 - pow(beta1, *step);
+      beta2_correction = 1 - pow(beta2, *step);
+    }
+
+    int tensor_loc = tl.block_to_tensor[blockIdx.x];
+    int chunk_idx = tl.block_to_chunk[blockIdx.x];
+    int n = tl.sizes[tensor_loc];
+
+    const float grad_scale = *grad_scale_ptr;
```
crcrpar, line 277: nit: might better to check whether or not `grad_scale_ptr` is `nullptr`

apex/contrib/csrc/optimizers/multi_tensor_distopt_adam_kernel.cu
```diff
@@ -194,6 +194,170 @@ struct DistAdamFunctor
...
+      local_step(
+        local_p, local_m, local_v, local_g, grad_scale,
+        beta1, beta2, beta1_correction, beta2_correction,
+        eps, *lr, mode, weight_decay);
+#pragma unroll
+      for (int ii = 0; ii < ILP; ii++) {
+        local_p_out[ii] = static_cast<PARAM_OUT_T>(local_p[ii]);
+      }
+
+      // Store
```
crcrpar, line 340: would there be any appetite to use gradients after `step`? if so, it'd be necessary to store unscaled gradients as well.

apex/contrib/optimizers/distributed_fused_adam.py
```diff
@@ -2329,7 +2392,8 @@ def step(
                     )
 
         # Apply optimizer step
-        self.state["step"] += 1
+        self.state["step"] += 1 if not self.capturable else \
+            (self._dummy_overflow_buf != 1).to(torch.int)
```
crcrpar, line 2396: Q: where would we decrement this value when `self.capturable` is `True` and invalid grads are found?

apex/contrib/csrc/optimizers/multi_tensor_distopt_adam_kernel.cu
alpha0422: We haven't encounter such cases yet, after optimizer stepping a new iteration starts, where the gradients will be zero out first. This is the same behavior as other optimizer kernels in the repo, so I think we can leave it as it is until there're cases we need to store unscaled gradients in future.

apex/contrib/optimizers/distributed_fused_adam.py
alpha0422: As you know `self.state["step"]` is to track how many steps the optimizer has advanced, it is used for bias correction in the CUDA kernel. When invalid grads are found, `self._dummy_overflow_buf` is 1, then it's `self.state["step"] += 0`, otherwise it's `self.state["step"] += 1`.

timmoon10 (changes requested):
As we've discussed, the grad clipping behavior is incorrect because plain PyTorch optimizers don't handle grad scaling gracefully:
```python
# Plain PyTorch
torch.nn.clip_grad_norm_(model.parameters())
scaler.step(optim)  # Clipped grads are scaled

# Distributed optimizer
optim.clip_grad_norm()
scaler.step(optim)  # Clipped grads are scaled
```
I'd prefer if distopt were as close as possible to a drop-in optimizer replacement, so I don't think the current behavior is a bug.

Supporting correct grad clipping is important though. I propose the following API:
```python
# Plain PyTorch
scaler.unscale_(optim)
torch.nn.clip_grad_norm_(model.parameters())
scaler.step(optim)  # Grads are not scaled

# Distributed optimizer
optim.unscale_grads(grad_scaler=scaler)
optim.clip_grad_norm()
scaler.step(optim)  # Grads are not scaled
```

apex/contrib/optimizers/distributed_fused_adam.py
crcrpar: uh, I misread it, thank you for correcting me.

another question: would this be really host-device sync free?

alpha0422: Yes, in this case distributed fused adam is sync-free.

apex/contrib/csrc/optimizers/multi_tensor_distopt_adam_kernel.cu
alpha0422: Added some assertion at the beginning.

timmoon10:
I've implemented [my proposed API](https://github.com/NVIDIA/apex/pull/1794#pullrequestreview-1993212897) at https://github.com/timmoon10/apex/commit/8293ae20636c587b2415db5007e16f3f24551b36, although I haven't tested. 

NeMo GPT avoided these issues because it implemented a custom `GradScaler` that called `DistributedFusedAdam.unscale_grads` within `unscale_`: https://github.com/NVIDIA/NeMo/blob/c5738263d8b4bedb0957374116d3e90746a51c37/nemo/collections/nlp/parts/nlp_overrides.py#L1235. See https://github.com/NVIDIA/apex/pull/1512 and https://github.com/NVIDIA/NeMo/pull/4900.

`_step_supports_amp_scaling` is needed because otherwise `GradScaler` would attempt to access the parameters' `.grad`s, which have probably already been reduce-scattered and set to `None`. The only way I can see to avoid this is to disable overlapping grad reduce-scatters with backward compute.

alpha0422:
But there's also the issue when `_step_supports_amp_scaling` set, `GradScaler.unscale_` will never be called from PyTorch or PyTorch Lightning. I saw you tried to unscale at here: [nlp_overrides.py#L1202](https://github.com/NVIDIA/NeMo/blob/v1.23.0/nemo/collections/nlp/parts/nlp_overrides.py#L1202), but this function was never called, I confirmed with Stable Diffusion and LLM.

Overlapping reduce-scatter with bprop is quite important to the performance, so I think it is necessary.

timmoon10:
I see, we need `_step_support_amp_scaling=False` specifically when using [`nemo.collections.nlp.parts.nlp_overrides.GradScaler`](https://github.com/NVIDIA/NeMo/blob/main/nemo/collections/nlp/parts/nlp_overrides.py). However, `_step_support_amp_scaling=True` is needed for correct behavior with `torch.amp.GradScaler`. I think the cleanest solution is to set `_step_support_amp_scaling=False` in [NeMo's distopt wrapper](https://github.com/NVIDIA/NeMo/blob/main/nemo/core/optim/distributed_adam.py). That helps keep the NeMo-specific logic separate from the general PyTorch logic. Reverting the changes to the grad clipping logic (e.g. with https://github.com/timmoon10/apex/commit/0fa8e3a3a231e655f7e1690dfaf9f2f7c53be4e6) is needed to preserve correct behavior with `torch.amp.GradScaler`.

PR #1801, opened 2024-05-08
Avoid importing apex transformer automatically 

nWEIdia:
and make error message more clear when apex.transformer is explicitly called on unsupported platform

In some configurations, pytorch may not be built with distributed support. 
So we avoid the auto import of transformer, which calls torch.distributed api explicitly.

### DIFF
--- a/apex/__init__.py
+++ b/apex/__init__.py
@@ -24,7 +24,6 @@ from . import fp16_utils
 # load time) the error message is timely and visible.
 from . import optimizers
 from . import normalization
-from . import transformer
 
 
 # Logging utilities for apex.transformer module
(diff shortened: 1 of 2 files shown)


apex/transformer/utils.py
```diff
@@ -8,6 +8,7 @@
 # The following 4 lines are for backward comparability with
 # older PyTorch.
 if "all_gather_into_tensor" not in dir(torch.distributed):
+    assert torch.distributed.is_available(), "PyTorch Distributed is Not available or Disabled."
```
crcrpar, line 11: as we can technically ignore `assert` ([ref](https://docs.python.org/3/using/cmdline.html#cmdoption-O))
```suggestion
    if not torch.distributed.is_available():
        raise RuntimeError("PyTorch Distributed is Not available or Disabled.")
```

apex/__init__.py
crcrpar, line 1: also, remove `"transformer"` from the dunder all

crcrpar (approved):

apex/__init__.py
nWEIdia: Sorry, "the dunder all" means?  :)

PR #1806, opened 2024-06-07
Use torch.testing.all_close instead of get_max_diff in test_lamb.py

Fuzzkatt:
Previously various tests in test_lamb.py were using the custom get_max_diff comparison function, which was causing numerical mismatches to appear in CI. This PR updates them to use the much more standard torch.testing.all_close, after which they are cleanly passing CI.


Fuzzkatt:
cc @eqy @crcrpar for review

crcrpar (approved):
great

PR #1818, opened 2024-07-19
NCCLAllocator: Fix build failure

Aidyn-A:
This PR adds `shareIpcHandle` to `NCCLAllocator` to satisfy its base class definition `CUDAAllocator` which has recently changed (https://github.com/pytorch/pytorch/pull/130888). 

cc @xwang233

### DIFF
--- a/apex/contrib/csrc/nccl_allocator/NCCLAllocator.cpp
+++ b/apex/contrib/csrc/nccl_allocator/NCCLAllocator.cpp
@@ -324,7 +324,7 @@ c10::cuda::CUDACachingAllocator::SnapshotInfo NCCLAllocator::snapshot() {
   return result;
 }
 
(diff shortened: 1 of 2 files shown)


crcrpar:
Q: what would happen if I build this with pytorch prior without the linked pr merged?

Aidyn-A:
> Q: what would happen if I build this with pytorch prior without the linked pr merged?

It will fail because `ShareableHandle` struct gets defined only in that PR :unamused:

xwang233:
> Q: what would happen if I build this with pytorch prior without the linked pr merged?

I think we can guard against TORCH_VERSION (or similar macro) for 2.5. Torch 2.4 will be released soon and won't have this field.

Aidyn-A:
Closing in favor of #1863

PR #1821, opened 2024-07-24
Allow Configurable Cache Directory

leimao:
Apex ASP will generate a cache file called `master_list.pkl` at the root of the working directory and it just looks abrupt.

Change the default behavior to creating a cache directory `.cache` at the root of the working directory and saving/loading the `master_list.pkl` there.

The user can also change the cache directory by setting the environment variable `ASP_CACHE_DIR` in the OS.

### DIFF
--- a/apex/contrib/sparsity/permutation_search_kernels/exhaustive_search.py
+++ b/apex/contrib/sparsity/permutation_search_kernels/exhaustive_search.py
@@ -60,6 +60,7 @@ def generate_unique_combinations(built_permutation, remaining_columns, full_perm
                 remaining_columns.insert(c, built_permutation.pop(-1))
 
 import pickle
+import os
 from os import environ, path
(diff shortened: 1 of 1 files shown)


apex/contrib/sparsity/permutation_search_kernels/exhaustive_search.py
```diff
@@ -1,5 +1,8 @@
 from .permutation_utilities import *
 
+ASP_CACHE_DIR_ENV_VAR = "ASP_CACHE_DIR"
```
crcrpar, line 3: How about `APEX_ASP_CACHE_DIR` to clarify the env var is for apex's asp module?

leimao: Done.

crcrpar (approved):

PR #1825, opened 2024-08-13
Fix illegal memory access with multi_tensor_apply size above INT_MAX

gdb:
Currently, multi_tensor_apply causes an illegal memory access due to an overflow in the `sizes` field of `TensorListMetadata`. This can be reproduced using the following standalone script:

```python
import torch, amp_C
from apex.multi_tensor_apply import multi_tensor_applier
multi_tensor_adam = amp_C.multi_tensor_adam

size = 2**32+1
g_32 = [torch.zeros(size, dtype=torch.float32, device='cuda')]
p_32 = [torch.zeros(size, dtype=torch.float32, device='cuda')]
m_32 = [torch.zeros(size, dtype=torch.float32, device='cuda')]
v_32 = [torch.zeros(size, dtype=torch.float32, device='cuda')]
_dummy_overflow_buf = torch.zeros(1, dtype=torch.int32, device='cuda')

multi_tensor_applier(multi_tensor_adam, _dummy_overflow_buf, [g_32, p_32, m_32, v_32], 0.0, 0.9, 0.95, 1e-08, 1, 1, 1, 0.1)
print(g_32)
```

### DIFF
--- a/csrc/multi_tensor_apply.cuh
+++ b/csrc/multi_tensor_apply.cuh
@@ -19,7 +19,7 @@ constexpr int depth_to_max_blocks[6] = {320, 320, 320, 320, 320, 320};
 template<int n> struct TensorListMetadata
 {
   void* addresses[n][depth_to_max_tensors[n-1]];
-  int sizes[depth_to_max_tensors[n-1]];
+  int64_t sizes[depth_to_max_tensors[n-1]];
   unsigned char block_to_tensor[depth_to_max_blocks[n-1]];
   int block_to_chunk[depth_to_max_blocks[n-1]]; // I fear this needs to be a full int.
   int start_tensor_this_launch;


awgu:
cc: @crcrpar are the following out of date:
https://github.com/NVIDIA/apex/blob/b3bd26a8004007e6d2d098934e063b952cab86f1/csrc/multi_tensor_apply.cuh#L15-L17
I see the same limits [in PyTorch](https://github.com/pytorch/pytorch/blob/c17d26c3c1b86e9ed7827e1a6e1a3dcc810b5ff8/aten/src/ATen/native/cuda/MultiTensorApply.cuh#L17-L20) where you already updated to use `int64_t` in https://github.com/pytorch/pytorch/pull/101760. Otherwise, I would expect that changing to use `int64_t` increases the `TensorListMetadata` struct size and hence the kernel arg size.

(Though, it seems that CUDA 12.1 on Volta+ [increased the kernel arg size limit](https://developer.nvidia.com/blog/cuda-12-1-supports-large-kernel-parameters/) from 4 KB to 32 KB.)

crcrpar:
> I would expect that changing to use `int64_t` increases the `TensorListMetadata` struct size and hence the kernel arg size.

Yes, but apex does not have multi-tensor-apply with a list of scalars so we might be able to dodge a tweak of depth_to_max_tensors and depth_to_max_blocks

crcrpar (approved):
excuse my delay, thank you

firoj0:
[C/C++ CI](https://github.com/firoj0/apex/actions/workflows/c-cpp.yml)

firoj0 (commented):
[C/C++ CI](https://github.com/firoj0/apex/actions/workflows/c-cpp.yml)

PR #1832, opened 2024-08-22
Enhance Distributed Fused Adam

alpha0422:
Continue of https://github.com/NVIDIA/apex/pull/1794.

This PR enhances distributed fused adam by:

* Support NHWC layout (required by some Conv related models, e.g. Diffusion models);
* Fix the gradient clipping bug;
* Support CUDA graph;

@timmoon10 @crcrpar Please help review, thanks.

### DIFF
--- a/apex/contrib/csrc/optimizers/multi_tensor_distopt_adam_kernel.cu
+++ b/apex/contrib/csrc/optimizers/multi_tensor_distopt_adam_kernel.cu
@@ -6,25 +6,26 @@
 // #include <torch/all.h>
 
 #include <assert.h>
+
 #include <cmath>
-#include "type_shim.h"
+
 #include "multi_tensor_apply.cuh"
+#include "type_shim.h"
 
 #define BLOCK_SIZE 512
 #define ILP 4
 
-template<typename T>
-__device__ __forceinline__ bool is_aligned(const T* p){
-  return ((uint64_t)p) % (ILP*sizeof(T)) == 0;
+template <typename T>
+__device__ __forceinline__ bool is_aligned(const T* p) {
+  return ((uint64_t)p) % (ILP * sizeof(T)) == 0;
 }
 
-template<typename T>
-__device__ __forceinline__ void load_store(
-  T* dst,
-  const T* src,
-  int dst_offset = 0,
-  int src_offset = 0){
-  typedef typename std::aligned_storage<ILP*sizeof(T), ILP*alignof(T)>::type LT;
+template <typename T>
+__device__ __forceinline__ void load_store(T* dst, const T* src,
+                                           int dst_offset = 0,
+                                           int src_offset = 0) {
+  typedef
+      typename std::aligned_storage<ILP * sizeof(T), ILP * alignof(T)>::type LT;
   ((LT*)dst)[dst_offset] = ((const LT*)src)[src_offset];
 }
 
@@ -34,9 +35,9 @@ __device__ __forceinline__ float lerp(float t, float x, float y) {
   return fma(t, y, fma(-t, x, x));
 }
 
-typedef enum{
-  ADAM_MODE_0   =0, // L2 regularization mode
-  ADAM_MODE_1   =1  // Decoupled weight decay mode(AdamW)
+typedef enum {
+  ADAM_MODE_0 = 0,  // L2 regularization mode
+  ADAM_MODE_1 = 1   // Decoupled weight decay mode(AdamW)
 } adamMode_t;
 
 /* Multi-tensor Adam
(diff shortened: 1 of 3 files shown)


apex/contrib/csrc/optimizers/multi_tensor_distopt_adam_kernel.cu
```diff
@@ -194,6 +194,175 @@ struct DistAdamFunctor
...
+        load_store(v + i_start, local_v);
+        load_store(p_out + i_start, local_p_out);
+      } else {
+#pragma unroll
+        for (int ii = 0, i = i_start; ii < ILP; ii++, i++) {
+          if (i < n) {
+            p_in[i] = local_p[ii];
+            m[i] = local_m[ii];
+            v[i] = local_v[ii];
+	    p_out[i] = local_p_out[ii];
```
crcrpar, line 358: could you use the same indent as the lines above?

apex/contrib/csrc/optimizers/multi_tensor_distopt_adam_kernel.cu
```diff
@@ -382,6 +551,51 @@ void multi_tensor_fused_adam_cuda(
...
+  // Expect p_in, m, v, g, p_out
+  size_t tl_sz = tensor_lists.size();
+  TORCH_CHECK(tl_sz == 5, "expected tensor lists of size 5");
+  const auto p_in_type = tensor_lists[0][0].scalar_type();
+  const auto g_type = tensor_lists[3][0].scalar_type();
+  const auto p_out_type = tensor_lists[4][0].scalar_type();
+
+  DISPATCH_FLOAT_HALF_AND_BFLOAT(p_in_type, 0, "dist_adam_capturable_cuda_kernel",
+    DISPATCH_FLOAT_HALF_AND_BFLOAT(g_type, 1, "dist_adam_capturable_cuda_kernel",
+      DISPATCH_FLOAT_HALF_AND_BFLOAT(p_out_type, 2, "dist_adam_capturable_cuda_kernel",
```
crcrpar, line 579: I thought g_type would basically be either `p_in_type` or `p_out_type`, and wondering if it's possible to remove one of these three dispatch's

also, which, p_in or p_out is `model.parameters()`? I'm confused

apex/contrib/optimizers/distributed_fused_adam.py
```diff
@@ -565,7 +568,24 @@ def __init__(
...
+            raise Exception(f"Capturable {self.__class__.__name__} relies on "
+                "multi_tensor_copy to set dummy_overflow_buf to indicate "
+                "whether there's gradient Inf/NaN, build APEX with "
+                "`--deprecated_fused_adam` is essential.")
+
+        # If capturable for CUDA graph
+        self.capturable: bool = capturable
+        # If the optimizer is capturable then LR should be a tensor (on GPU)
+        lr: torch.Tensor | float = torch.tensor(lr, dtype=torch.float32) \
+            if capturable else lr
```
crcrpar, line 587: is `lr` really on cuda device with this initialization?

timmoon10: Also, union typing with `|` is only supported in Python 3.10 (https://peps.python.org/pep-0604/):
```suggestion
        if capturable:
            lr = torch.tensor([lr], dtype=torch.float32, device=device)
```

apex/contrib/optimizers/distributed_fused_adam.py
```diff
@@ -709,6 +730,7 @@ def __init__(
         # Determine bucket sizes
         dtype_size = torch.finfo(self.grad_sync_dtype).bits // 8
         self.alignment: int = 128 // dtype_size
+        self.bucket_cap_mb: float = bucket_cap_mb
```
timmoon10, line 733: Nit: I don't think `bucket_cap_mb` is used.
```suggestion
```
If we want to use it externally, we could recompute something more accurate:
```python
bucket_size = self.default_shard_size * self.distributed_size
self.default_bucket_size_mb: float = bucket_size * dtype_size / (1024 * 1024)
```

apex/contrib/csrc/optimizers/multi_tensor_distopt_adam_kernel.cu
timmoon10: `p_in` corresponds to the master weights (FP32) and `p_out` corresponds to the model weights (BF16).

There are some cases where we don't want to output to `p_out` and I hackily implemented by setting `p_out = p_in`, so it's not guaranteed that `g_type` matches `p_in_type` or `p_out_type`. This is mainly for the FP8 case, and I wanted to keep FP8 logic out of Apex.

timmoon10 (approved):
Overall looks good to me aside from some stylistic suggestions.

apex/contrib/csrc/optimizers/multi_tensor_distopt_adam_kernel.cu
alpha0422: Done. Formatted the code with clang-format Google style.

apex/contrib/optimizers/distributed_fused_adam.py
alpha0422: Sorry but I have to revert this. `self.bucket_cap_mb` is indeed not used anymore in the code, but it's helpful for use to ensure all Dist Adam arguments are correct. There's a code snippet to print Dist Adam with all important arguments in `__repr__`, `self.bucket_cap_mb` is used there. So I think we need to keep it.

crcrpar (commented):
could you add a test case for capturable?

apex/contrib/optimizers/distributed_fused_adam.py
alpha0422: Hmm, `torch.tensor([lr])` is 1-dim tensor, ` torch.tensor(lr)` is 0-dim tensor. PyTorch LR scheduler expects the later one.

crcrpar (approved):
agreed on @alpha0422 will add a test in a follow-up.

PR #1845, opened 2024-09-26
fix groupnorm int32 index overflow

tlogn:
This PR fix groupnorm int32 index calculate overflow when h*w*c is large, as hwc is of int data type. The problem could be reproduced by code below.
@crcrpar 

```python
from apex.contrib.group_norm import GroupNorm as ApexGroupNorm
import torch
layer = ApexGroupNorm(32, 128, dtype=torch.bfloat16, device='cuda', act='silu')

x = torch.randn(1, 128, 16128, 1200, dtype=torch.bfloat16, device='cuda').to(memory_format=torch.channels_last)

o = layer(x)
print(o[0][0][0][0])

```

### DIFF
--- a/apex/contrib/csrc/group_norm/group_norm_nhwc.h
+++ b/apex/contrib/csrc/group_norm/group_norm_nhwc.h
@@ -123,7 +123,7 @@ struct Group_norm_nhwc_fwd_params {
   // The number of instances in the batch.
   int n;
   // The height and width of each activation map. The number of channels.
-  int h, w, c, hw, hwc;
+  int64_t h, w, c, hw, hwc;
   // The number of groups.
   int groups;
   // Do we apply the Swish activation function?
@@ -138,7 +138,7 @@ struct Group_norm_nhwc_fwd_params {
   // The number of groups in each block.
   int groups_per_block;
   // The number of channels per group = c / groups.
-  int channels_per_group; 
+  int channels_per_group;
   // The number of channels per block = groups_per_block * channels_per_group.
   int channels_per_block;
   // The inverse of hwc in floats (to compute mean/var).
@@ -149,7 +149,7 @@ struct Group_norm_nhwc_fwd_params {
 
 ////////////////////////////////////////////////////////////////////////////////////////////////////
 
-void group_norm_nhwc_fwd_two_passes_setup(Group_norm_nhwc_fwd_params&, 
+void group_norm_nhwc_fwd_two_passes_setup(Group_norm_nhwc_fwd_params&,
                                           size_t &red_buffer_elts);
 
 ////////////////////////////////////////////////////////////////////////////////////////////////////
@@ -190,7 +190,7 @@ struct Group_norm_nhwc_bwd_params {
   // The number of instances in the batch.
   int n;
   // The height and width of each activation map. The number of channels.
-  int h, w, c, hw, hwc;
+  int64_t h, w, c, hw, hwc;
   // The number of groups.
   int groups;
   // Do we apply the Swish activation function?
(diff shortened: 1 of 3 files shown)


crcrpar (approved):
sorry for having let this one slip off, could you add a test case which this PR allows?

tlogn:
@crcrpar thanks for your advise. I've added a test case and tested with H100.

with the main branch, error occurs:
```
Mismatched elements: 1866930499 / 2477260800 (75.4%)
Greatest absolute difference: nan at index (0, 0, 2146, 1184) (up to 0.04 allowed)
Greatest relative difference: nan at index (0, 0, 2146, 1184) (up to 0 allowed)

----------------------------------------------------------------------
Ran 16 tests in 8.897s

FAILED (failures=1)
```

with the fixed branch, the atol should be adjusted to 7e-2 to account for potential increased reduction accuracy error
```
................
----------------------------------------------------------------------
Ran 16 tests in 8.079s

OK
```

apex/contrib/test/group_norm/test_group_norm.py
```diff
@@ -89,10 +89,10 @@ def verify_group_norm(self,
...
 
         # compare
-        torch.testing.assert_close(y_tst, y_ref, atol=4e-2, rtol=0)
-        torch.testing.assert_close(dx_tst, dx_ref, atol=4e-2, rtol=0)
-        torch.testing.assert_close(dw_tst, dw_ref, atol=4e-2, rtol=0)
-        torch.testing.assert_close(db_tst, db_ref, atol=4e-2, rtol=0)
+        torch.testing.assert_close(y_tst,  y_ref,  atol=7e-2, rtol=0)
+        torch.testing.assert_close(dx_tst, dx_ref, atol=7e-2, rtol=0)
+        torch.testing.assert_close(dw_tst, dw_ref, atol=7e-2, rtol=0)
+        torch.testing.assert_close(db_tst, db_ref, atol=7e-2, rtol=0)
```
crcrpar, line 95: is this for the large tensor?

apex/contrib/test/group_norm/test_group_norm.py
```diff
@@ -177,6 +177,7 @@ def test_16_groups(self):
             [8, 1920, 32, 32],
             [8, 1920, 16, 16],
             [8, 2560, 8, 8],
+            [1, 128, 16128, 1200],
```
crcrpar, line 180: this seems to require about 50 GB, how about checking the memory size like

```
diff --git a/apex/contrib/test/group_norm/test_group_norm.py b/apex/contrib/test/group_norm/test_group_norm.py
index 5675749..687c2e1 100644
--- a/apex/contrib/test/group_norm/test_group_norm.py
+++ b/apex/contrib/test/group_norm/test_group_norm.py
@@ -177,8 +177,9 @@ class GroupNormTest(unittest.TestCase):
             [8, 1920, 32, 32],
             [8, 1920, 16, 16],
             [8, 2560, 8, 8],
-            [1, 128, 16128, 1200],
         ]
+        if torch.cuda.get_device_properties().total_memory > 50_000_000_000:
+            sizes.append([1, 128, 16128, 1200])
         for sz in sizes:
             n, c, h, w = sz
             self.verify_group_norm(GroupNorm,
```

apex/contrib/test/group_norm/test_group_norm.py
tlogn: yes, float reduction accuracy error would grow up with larger tensor

apex/contrib/test/group_norm/test_group_norm.py
tlogn: I've checked seriously, this tensor gots 128*16128*1200=2,477,260,800, about 5GB. This kernel will take about 10GB, so this check seems not so necessary ?

tlogn:
@crcrpar 
Hi there, is there any problem remained ?

PR #1847, opened 2024-09-28
Avoid unnecessary NCCL collective coalescing in distributed optimizer

timmoon10:
I've been experiencing some data corruption in distributed optimizer checkpoints because PyTorch is not properly synchronizing the NCCL stream with the main CUDA stream. All indications point to a bug in PyTorch's infrastructure for coalesced NCCL calls and I've isolated it down to cases where we enter [PyTorch's `_coalescing_manager`](https://github.com/pytorch/pytorch/blob/156ca01e51f766b1b069c5c6f3d57112a5c8f9ff/torch/distributed/distributed_c10d.py#L2244) but do not perform any NCCL collectives. The debugger suggests that `_coalescing_manager` sets [this flag](https://github.com/pytorch/pytorch/blob/156ca01e51f766b1b069c5c6f3d57112a5c8f9ff/torch/csrc/distributed/c10d/ProcessGroupNCCL.hpp#L1103) when it enters the context and fails to unset it, resulting in weird behavior in later NCCL calls. I haven't fully bottomed out this bug, but this PR fixes the issue for me.

### DIFF
--- a/apex/contrib/optimizers/distributed_fused_adam.py
+++ b/apex/contrib/optimizers/distributed_fused_adam.py
@@ -1966,7 +1966,7 @@ class DistributedFusedAdam(torch.optim.Optimizer):
         comm_stream.wait_stream(main_stream)
 
         # Reduce-scatter over distributed process group
-        if self.distributed_size > 1:
+        if buckets and self.distributed_size > 1:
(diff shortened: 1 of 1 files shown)


crcrpar (approved):
Looks reasonable to me, thank you

PR #1854, opened 2024-10-29
[PT2] Normalisation: use manual impl when compiling

alexdremov:
Using kernel triggers recompilations due to the superfluous guards torch generates for custom kernels. Executing manual implementation should be better


crcrpar (approved):

alexdremov:
@crcrpar merging?

PR #1835, opened 2024-08-28
Traceable GroupNorm

alpha0422:
This PR tries to make GroupNorm can be traced by TorchInductor and don't invoke graph break.

@crcrpar, please take a review, thanks!

PR #1861, opened 2024-11-28
Traceable RMSNorm

yanboliang:
Make ```FusedRMSNorm``` can be traced through by TorchDynamo to provide a better composibility with ```torch.compile```. If this being accepted, I'll make the rest modules traceable as follow-up PRs.
Similar to https://github.com/NVIDIA/apex/pull/1835
cc @alpha0422 @crcrpar


crcrpar (approved):
fantastic. Thank you so much

PR #1862, opened 2024-12-01
Traceable RMSNorm part 2: non affine fused rms norm

yanboliang:
Make non affine ```FusedRMSNorm``` can be traced through by TorchDynamo to provide a better composibility with torch.compile.

cc @crcrpar

### DIFF
--- a/apex/normalization/fused_layer_norm.py
+++ b/apex/normalization/fused_layer_norm.py
@@ -13,6 +13,11 @@ global fused_layer_norm_cuda
 fused_layer_norm_cuda = None
 
 
+# PyTorch supports `torch.library.custom_op` since 2.4.0.
+def supports_custom_op() -> bool:
+    return hasattr(torch.library, "custom_op")
+
+
 # Reference implementation from Huggingface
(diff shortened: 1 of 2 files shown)


J-shang:
hi @yanboliang , `torch.library.custom_op` is introduced in torch 2.4, I find it break my env (torch 2.3), it would be kind if you could forward compatibility the torch version or add the torch version requirement

yanboliang:
@J-shang Sounds good! I'll update this PR to include a shim layer.

yanboliang:
Gently ping @crcrpar

crcrpar:
I was on vacation last week :sweat_smile:

crcrpar (approved):

crcrpar:
@yanboliang Thank you for another PR and adding the availability check of `custom_op`
@J-shang The recommended way of using APEX is NGC PyTorch containers and/or source build with the latest/nightly PyTorch as in https://github.com/NVIDIA/apex?tab=readme-ov-file#installation. We try not to break things that frequently but by default we build and test against nightly pytorch usually, for your information

J-shang:
Thanks @crcrpar , that's cool!

PR #1887, opened 2025-03-12
Add group_norm_v2

yuantailing:
GroupNorm v2 outperforms the original GroupNorm extension by utilizing coalesced memory access, tested on B200 with a set of shapes.

### DIFF
--- a/apex/contrib/csrc/group_norm_v2/gn.cpp
+++ b/apex/contrib/csrc/group_norm_v2/gn.cpp
@@ -97,7 +97,7 @@ auto gn_bwd(torch::Tensor grad_output, torch::Tensor x, torch::Tensor w, torch::
     return std::make_tuple(grad_input, grad_weight, grad_bias);
 }
 
-}
+}  // namespace group_norm_v2
 
 PYBIND11_MODULE(TORCH_EXTENSION_NAME, m) {
     m.def("gn", &group_norm_v2::gn, py::arg("x"), py::arg("w"), py::arg("b"), py::arg("eps"), py::arg("silu"), py::arg("num_groups"), py::arg("mean_var_out") = py::none(), py::arg("sm_margin") = 0, "");
--- a/apex/contrib/csrc/group_norm_v2/gn.cpp
+++ b/apex/contrib/csrc/group_norm_v2/gn.cpp
@@ -1,0 +1,105 @@
+#include <torch/extension.h>
+#include <ATen/cuda/CUDAContext.h>
+
+#include "gn.hpp"
+
+
+namespace group_norm_v2 {
+
+torch::Tensor gn(torch::Tensor x, torch::Tensor w, torch::Tensor b, float eps, bool silu, int num_groups, std::optional<torch::Tensor> mean_var_out, int sm_margin) {
+    if (w.dtype() != b.dtype() || (mean_var_out.has_value() && mean_var_out->dtype() != torch::kFloat32)) {
+        throw std::invalid_argument("gn dtype mismatch");
+    }
+    torch::Tensor out = torch::empty_like(x);
+    float *ptr_mean_var_out = mean_var_out.has_value() ? mean_var_out->data_ptr<float>() : nullptr;
+    cudaStream_t stream = at::cuda::getCurrentCUDAStream().stream();
+    int device_id = at::cuda::getCurrentCUDAStream().device().index();
+    group_norm_v2::Meta meta;
+    if (x.dtype() == torch::kHalf && w.dtype() == torch::kHalf) {
+        group_norm_v2::gn_cuda(
+            (half *)out.data_ptr(), (half *)x.data_ptr(), (half *)w.data_ptr(), (half *)b.data_ptr(),
+            eps, silu, x.size(0), x.size(2) * x.size(3), num_groups, x.size(1) / num_groups, ptr_mean_var_out,
+            nullptr, nullptr, sm_margin, stream, device_id, &meta, true);
+    } else if (x.dtype() == torch::kBFloat16 && w.dtype() == torch::kBFloat16) {
+        group_norm_v2::gn_cuda(
+            (__nv_bfloat16 *)out.data_ptr(), (__nv_bfloat16 *)x.data_ptr(), (__nv_bfloat16 *)w.data_ptr(), (__nv_bfloat16 *)b.data_ptr(),
+            eps, silu, x.size(0), x.size(2) * x.size(3), num_groups, x.size(1) / num_groups, ptr_mean_var_out,
+            nullptr, nullptr, sm_margin, stream, device_id, &meta, true);
+    } else {
+        throw std::invalid_argument("gn only supports half or bfloat16 input and weight");
+    }
+    torch::Tensor red_buffer = torch::empty({meta.red_buffer_size}, torch::TensorOptions().dtype(torch::kFloat32).device(torch::kCUDA));
+    thread_local torch::Tensor barrier;
+    if (barrier.size(0) < meta.barrier_size) {
+        barrier = torch::zeros({meta.barrier_size}, torch::TensorOptions().dtype(torch::kUInt32).device(torch::kCUDA));
+    }
+    if (x.dtype() == torch::kHalf && w.dtype() == torch::kHalf) {
+        group_norm_v2::gn_cuda(
+            (half *)out.data_ptr(), (half *)x.data_ptr(), (half *)w.data_ptr(), (half *)b.data_ptr(),
+            eps, silu, x.size(0), x.size(2) * x.size(3), num_groups, x.size(1) / num_groups, ptr_mean_var_out,
+            red_buffer.data_ptr<float>(), barrier.data_ptr<unsigned>(), sm_margin, stream, device_id, nullptr, false);
+    } else if (x.dtype() == torch::kBFloat16 && w.dtype() == torch::kBFloat16) {
+        group_norm_v2::gn_cuda(
+            (__nv_bfloat16 *)out.data_ptr(), (__nv_bfloat16 *)x.data_ptr(), (__nv_bfloat16 *)w.data_ptr(), (__nv_bfloat16 *)b.data_ptr(),
+            eps, silu, x.size(0), x.size(2) * x.size(3), num_groups, x.size(1) / num_groups, ptr_mean_var_out,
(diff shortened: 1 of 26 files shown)


alpha0422:
Hi @crcrpar, could you also help review this PR?

setup.py
```diff
@@ -491,6 +491,33 @@ def check_cudnn_version_and_warn(global_option: str, required_cudnn_version: int
         )
     )
 
+    # CUDA group norm V2 is tested on SM100
+    if bare_metal_version >= Version("12.7"):
+        arch_flags = ["-gencode=arch=compute_100,code=sm_100"]
```
crcrpar, line 496: nit-picking: I'd use `> Version("12.7")` or `>= Version("12.8")` 
Question: no need to add sm120?

another nit-picking: if `TORHC_CUDA_ARCH_LIST` is correctly set, e.g., `TORCH_CUDA_ARCH_LIST="9.0 10.0 10.0a 12.0 12.0a", then I don't think it necessary to modify define this variable, IIUC.

apex/contrib/group_norm/group_norm.py
```diff
@@ -31,6 +33,15 @@ def one_time_warning(msg: str):
         one_time_warning.has_been_called = True
         print(f'\033[93m{msg}\033[0m')  # hightlight with yellow color
 
+
+@functools.cache
+def get_cc_and_sm_count(device_index: int):
+    props = torch.cuda.get_device_properties(device_index)
+    CC = (props.major, props.minor)
```
crcrpar, line 40: QQ: what would CC mean?

apex/contrib/csrc/group_norm_v2/gn_utils.hpp
```diff
@@ -0,0 +1,49 @@
...
+__host__ __device__ inline int print_rank_0(char const *fmt, Ts &&...args) {
+    if (threadIdx.x + threadIdx.y + threadIdx.z == 0 && blockIdx.x + blockIdx.y + blockIdx.z == 0) {
+        return printf(fmt, std::forward<Ts>(args)...);
+    }
+    return 0;
+}
+
+#endif
+
+}
```
crcrpar, line 49: Could you add give a comment of "namespace <namespace name>" to a right-hand side curly bracket if it's namespace, like what follows? Same applies to the other files.
```suggestion
} // namespace group_norm_v2
```

apex/contrib/group_norm/group_norm.py
Aidyn-A: It is compute capability. Basically a sm version.

apex/contrib/csrc/group_norm_v2/gn_utils.hpp
yuantailing: Good idea.

setup.py
```diff
@@ -491,6 +491,33 @@ def check_cudnn_version_and_warn(global_option: str, required_cudnn_version: int
...
+            name="group_norm_v2_cuda",
+            sources=[
+                "apex/contrib/csrc/group_norm_v2/gn.cpp",
+                "apex/contrib/csrc/group_norm_v2/gn_cuda.cu",
+                "apex/contrib/csrc/group_norm_v2/gn_utils.cpp",
+            ] + glob.glob("apex/contrib/csrc/group_norm_v2/gn_cuda_inst_*.cu"),
+            extra_compile_args={
+                "cxx": ["-O2", "-std=c++20"] + version_dependent_macros,
+                "nvcc": [
+                    "-O2", "-std=c++20", "--use_fast_math", "--ftz=false",
```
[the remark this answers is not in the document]
Aidyn-A: You may write the same thing in C++17 like so:
```C++
#include <tuple>

template<bool SWAP>
constexpr std::tuple<int, int> foo() {
    std::tuple<int, int> tup = {1, 2};
    if constexpr (SWAP) {
        std::tuple<int, int> new_tup = {3, 4};
        return new_tup;
    }
    return tup;
}

int main() {
    constexpr auto a = foo<true>();
}
```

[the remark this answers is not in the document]
Aidyn-A: The reason we are being conservative here is that the PyTorch is compiled with C++17. Would it be safe to make C++20 extensions for PyTorch?

setup.py
yuantailing: I think the condition check should be consistent with https://github.com/NVIDIA/apex/pull/1889/files . Maybe both two pull requests should change 12.7 to 12.8 .

Currently, the extension is only intended to be run on 148 SMs. The reason is that template arguments depend on SM_COUNT, and for future GPU cards whose SM_COUNT is unknown, GroupNorm v2 is not guaranteed to be runnable or to be faster than v1.
As for utilizing compile-time unknown SM_COUNT, I would like to develop the feature in a new pull request in the future.

Thus only `sm_100` is used, no matter which `TORCH_CUDA_ARCH_LIST` is.

apex/contrib/csrc/group_norm_v2/gn_cuda_kernel.cuh
```diff
@@ -0,0 +1,1056 @@
...
+                            sum.y += (float)val.data[i * GCD_VEC_CPG + k] * (float)val.data[i * GCD_VEC_CPG + k];
+                        }
+                        frag_sum_per_channel[i] = sum;
+                    }
+                }
+                for (int i = 0; i < VEC_ELEMS / GCD_VEC_CPG; i++) {
+                    if constexpr (SINGLE_GROUP_PER_BLOCK) {
+                        for (int mask = 16; mask > 0; mask >>= 1) {
+                            frag_sum_per_channel[i].x += __shfl_xor_sync(FINAL_MASK, frag_sum_per_channel[i].x, mask, 32);
+                            frag_sum_per_channel[i].y += __shfl_xor_sync(FINAL_MASK, frag_sum_per_channel[i].y, mask, 32);
```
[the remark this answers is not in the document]
yuantailing: Thanks for you information. Currently the file `gn_cuda_kernel.cuh` does not depend on any non-builtin headers, so I don't want to introduce PyTorch headers.

setup.py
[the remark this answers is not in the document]
yuantailing: > You may write the same thing in C++17 like so:

However, in the constexpr function `compute_gn_params`, the tuple `best_candidate` is to be assigned in a loop, so we cannot change it to a directly return statement.

apex/contrib/csrc/group_norm_v2/gn_cuda.cu
```diff
@@ -0,0 +1,54 @@
...
+
+#include "gn_utils.hpp"
+#include "gn_dispatch_hw_c.hpp"
+
+
+#define DISPATCH_G_SILU(g, silu, G, SILU, ...) [&] { \
+    if (g == 16 && silu == (g == 16)) { constexpr int G = 16; constexpr bool SILU = G == 16; return __VA_ARGS__(); } \
+    if (g == 32 && silu == (g == 16)) { constexpr int G = 32; constexpr bool SILU = G == 16; return __VA_ARGS__(); } \
+    throw std::invalid_argument("DISPATCH_G_SILU " + std::to_string(g) + " " + std::to_string(silu)); \
+    }()
```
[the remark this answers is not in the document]
yuantailing: Changed to `#define DISPATCH_NUM_GROUPS_AND_SILU(num_groups, silu, NUM_GROUPS, SILU, ...)`

apex/contrib/csrc/group_norm_v2/gn_cuda_kernel.cuh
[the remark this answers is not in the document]
crcrpar: that is reasonable

apex/contrib/csrc/group_norm_v2/gn.cpp
```diff
@@ -0,0 +1,105 @@
...
+            (__nv_bfloat16 *)grad_output.data_ptr(), (__nv_bfloat16 *)x.data_ptr(), (__nv_bfloat16 *)w.data_ptr(), (__nv_bfloat16 *)b.data_ptr(), mean_var.data_ptr<float>(),
+            eps, silu, x.size(0), x.size(2) * x.size(3), num_groups, x.size(1) / num_groups,
+            red_buffer.data_ptr<float>(), barrier.data_ptr<unsigned>(), sm_margin, stream, device_id, nullptr, false);
+    } else {
+        throw std::invalid_argument("gn only supports half or bfloat16 input and weight");
+    }
+    return std::make_tuple(grad_input, grad_weight, grad_bias);
+}
+
+}
```
crcrpar, line 100:
```suggestion
}  // namespace gorup_norm_v2
```

crcrpar (commented):
Could you add some test cases, e.g.
- check numeric and functionality of v2
- error is expectedly thrown for invalid inputs
?

apex/contrib/csrc/group_norm_v2/gn.cpp
yuantailing: Added test cases. Updated namespace comment.

setup.py
```diff
@@ -491,6 +491,33 @@ def check_cudnn_version_and_warn(global_option: str, required_cudnn_version: int
         )
     )
 
+    # CUDA group norm V2 is tested on SM100
+    if bare_metal_version >= Version("12.8"):
+        arch_flags = ["-gencode=arch=compute_100,code=sm_100"]
+    else:
+        arch_flags = ["-gencode=arch=compute_90,code=compute_90"]
```
Aidyn-A, line 498: QQ: Looking at this conditions, it is a Blackwell exclusive. Is it supposed to work on Hopper?

yuantailing: It works on Hopper if we add it to `arch_flags` and `DISPATCH_CUDA_ARCH_AND_LOWER_BOUND_SM_COUNT`, but we need a smart way to avoid exhausting `SM_COUNT`. This needs much code change and will not appear in this pull request.

yuantailing: It works on Hopper if we add it to `arch_flags` and `DISPATCH_CUDA_ARCH_AND_LOWER_BOUND_SM_COUNT`. However, I think we need a smarter way to determine template args to support more GPUs, instead of exhausting `SM_COUNT`.

setup.py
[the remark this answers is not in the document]
yuantailing: > The reason we are being conservative here is that the PyTorch is compiled with C++17. Would it be safe to make C++20 extensions for PyTorch?

Update: I have removed tuple assignment in constexpr functions, so it's a C++17 extension now.

crcrpar (approved):

crcrpar:
@nWEIdia yes.

PR #1888, opened 2025-03-12
[contrib] Update `apex.contrib.focal_loss` for B200

crcrpar:
- Main change: Use 8 CTAs per SM instead of 2, which improved perf on B200
- Generalize vectorization (still uses 128-bit)


Aidyn-A (approved):
Works fine on Blackwell machines. Thanks!

PR #1893, opened 2025-04-09
Fix DistributedFusedAdam for grad dtype != param dtype

yuantailing:
The dtype of tensor returned by `grad_buffer_view` should be the same as `_grad_buffers`, but it was `param.dtype`. In some cases that grad dtype != param dtype it leads to convergence issue.


crcrpar (approved):
great fix

PR #1894, opened 2025-04-10
Fix GroupNorm test: properly check graph breaks

yuantailing:
We should check `counters["graph_break"].total()` instead of `"graph_break" in counters` to determine whether there is any graph break.

The reason that the test **cannot pass the latest PyTorch** (but can pass old versions of PyTorch) is that after https://github.com/pytorch/pytorch/pull/146537 , `counters["graph_break"]` may be visited internally, so the origin `assertNotIn` may fail even if there is no graph breaks.

### DIFF
--- a/apex/contrib/test/group_norm/test_group_norm.py
+++ b/apex/contrib/test/group_norm/test_group_norm.py
@@ -18,6 +18,7 @@
 import functools
 import importlib
 import pathlib
+import sys
 import torch
 import unittest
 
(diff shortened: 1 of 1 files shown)


apex/contrib/test/group_norm/test_group_norm.py
```diff
@@ -202,7 +203,12 @@ def test_group_norm_inductor(self):
             y.backward(dy)
 
         from torch._dynamo.utils import counters
-        self.assertNotIn('graph_break', counters, "Shouldn't see any graph breaks.")
+        # TODO: Remove this when 3.9 is no longer supported
+        if sys.version_info < (3, 10):
+            num_graph_breaks = sum(counters["graph_break"].values())
+        else:
+            num_graph_breaks = counters["graph_break"].total()
```
Aidyn-A, line 210: Is `counters["graph_break"]` guaranteed to exist? Would it be necessary to check its existence like so:
```Python
...
        num_graph_breaks = 0
        if "graph_break" in counters:
            if sys.version_info < (3, 10):
                num_graph_breaks = sum(counters["graph_break"].values())
            else:
                num_graph_breaks = counters["graph_break"].total()
...
```

Aidyn-A: Okay, sounds good then!

Aidyn-A (approved):

PR #1896, opened 2025-04-17
remove old features: amp, ddp, and rnn

crcrpar:
- amp
- ddp
- RNN


apex/RNN/models.py
```diff
@@ -1,56 +0,0 @@
...
-    """
-
-    deprecated_warning("`apex.RNN` is deprecated and will be removed by the end of February 2023.")
-    if bidirectional:
-        return bidirectionalRNN(inputRNN, num_layers, dropout = dropout)
-    else:
-        return stackedRNN(inputRNN, num_layers, dropout = dropout)
-
-
-def LSTM(input_size, hidden_size, num_layers, bias=True, batch_first=False, dropout=0, bidirectional=False, output_size = None):
```
tfogal, line 21: I love that there's a `deprecated_warning` for all this code! Thanks for doing that a while ago, and double thanks for revisiting this now.

Rather than removing symbols outright,  though, I'd recommend having a period where the implementations `raise NotImplementedError`. This gives time for the ecosystem to both update its code and, more importantly, update imports.

As a quick background: https://github.com/search?q=%22import+apex.amp%22+AND+path%3A%2F.py%24%2F&type=code shows 350 hits of projects `import apex.amp`, and of course people could also be `from apex import amp`ing. Most of that code will fail to even import after this change, even if they have conditionals that cause it to never actually use apex's `amp` at runtime anyway.

(I realize the search is for amp and i'm commenting on an `LSTM` symbol here; my comment is global, not specific to this symbol!)

tfogal: 🤷 I'm not downvoting or upvoting, just recommending. You know the usage / space better than I do---I could be totally off on how quickly people update their code!

PR #1900, opened 2025-05-05
[Contrib][Openfold Triton] Use json instead of pickle

Aidyn-A:
In this PR I propose to use json serialization instead of pickle, as the latter one is known to be a RCE vulnerable.

cc @crcrpar  @arnowaczynski

### DIFF
--- a/apex/contrib/test/openfold_triton/test_sync_triton_auto_tune_cache_across_gpus.py
+++ b/apex/contrib/test/openfold_triton/test_sync_triton_auto_tune_cache_across_gpus.py
@@ -41,8 +41,10 @@ class SyncTritonAutoTuneCacheTest(MultiProcessTestCase):
 
     def _create_process_group_nccl(self):
         def maybe_export(env, val):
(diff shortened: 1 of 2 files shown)


apex/contrib/test/openfold_triton/test_sync_triton_auto_tune_cache_across_gpus.py
```diff
@@ -0,0 +1,95 @@
...
+        return f"{common_utils.FILE_SCHEMA}{self.file_name}"
+
+    @property
+    def destroy_pg_upon_exit(self) -> bool:
+        return True
+
+    def _create_process_group_nccl(self):
+        def maybe_export(env, val):
+            assert(type(env) == str)
+            assert(type(val) == str)
```
crcrpar, line 45: nit-picking
```suggestion
            assert type(env) == str
            assert type(val) == str
```
or even it'd be better to raise an exception

crcrpar (approved):

PR #1910, opened 2025-06-10
Allow env vars to specify custom extensions to build

crcrpar:
This would be the first step to support build backends other than setuptools and make setup.py a little bit independent of the backend, i.e., better compatibility with PEP517 and relevant PEPs.

With this we would be able to support some commands that have not been functioning well e.g.
```
$ APEX_CPP_EXT=1 APEX_CUDA_EXT=1 APEX_ALL_CONTRIB_EXT=1 python -m build --wheel --no-isolation 
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,24 +26,50 @@ The latest stable release obtainable from https://pytorch.org should also work.
 We recommend installing [`Ninja`](https://ninja-build.org/) to make compilation faster.
 
 ### Linux
-For performance and full functionality, we recommend installing Apex with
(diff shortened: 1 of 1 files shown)


nWEIdia:
LGTM! With this PR in, what is the default/recommended installation command?

crcrpar:
Env vars would be more modern than the apex's conventional install command.
also. I made a var of `APEX_ALL_CONTRIB_EXt` to build all the contrib extentions

PR #1911, opened 2025-06-16
Bump tolerances for fast layer norm

eqy:


crcrpar (approved):
we might switch to `torch.testing.assert_close` in a follow-up

PR #1915, opened 2025-07-10
Add an interface for the NCCL symmetric registration

youngeunkwon0405:


apex/contrib/nccl_allocator/nccl_allocator.py
```diff
@@ -8,9 +8,13 @@
...
 
-def create_nccl_mem_pool():
+def create_nccl_mem_pool(symmetric=False):
     _allocator = _apex_nccl_allocator.get_nccl_allocator()
-    _pool = torch.cuda.MemPool(_allocator)
+    _pool = (
+        torch.cuda.MemPool(_allocator, symmetric=True) 
+        if symmetric 
+        else torch.cuda.MemPool(_allocator)
+    )
```
crcrpar: isn't this equivalent and shorter?
```suggestion
    _pool = torch.cuda.MemPool(_allocator, symmetric=symmetric)
```

youngeunkwon0405: This is for backward compatibility. The `symmetric` argument was very recently added. If we do like your suggestion, then it may crash in older pytorch versions.

youngeunkwon0405: This is the related pytorch change. https://github.com/pytorch/pytorch/pull/155134/files#

crcrpar: Okay I see, then I think a meticulous thing we could do is to have a signature check whether or not `symmetric` argument is available and then show a warning if `symmetric` is set after making creat_nccl_mem_pool's symmetric argument's default value to `None`.

youngeunkwon0405: Hi @crcrpar, I have updated the PR with your suggestion. Thanks!

apex/contrib/nccl_allocator/nccl_allocator.py
```diff
@@ -7,14 +7,19 @@
 
 __all__ = ["init", "nccl_mem", "create_nccl_mem_pool"]
 
+def get_func_args(func):
+    import inspect
+    sig = inspect.signature(func)
+    return [arg.name for arg in sig.parameters.values()]
 
-def create_nccl_mem_pool(symmetric=False):
+def create_nccl_mem_pool(symmetric=None):
```
crcrpar: nit-picking
```suggestion
def create_nccl_mem_pool(symmetric: bool | None = None) -> torch.cuda.MemPool:
```

crcrpar (approved):
thank you for the update, looks good to me

youngeunkwon0405:
May I ask what the next step will be to merge this? @crcrpar

crcrpar:
Basically that's about it. I thought there's a test for this but unfortunately I see none

PR #1920, opened 2025-07-24
Handle temporary arg name divergence in the torch.cuda.MemPool

youngeunkwon0405:
There is a temporary divergence in NVIDIA internal pytorch version related to symmetric memory setting.
This PR handles that divergence.


eqy:
CC @syed-ahmed

crcrpar (approved):

PR #1927, opened 2025-08-13
fix compile group norm [v1/v2]

AlongWY:


AlongWY:
Okay, the code here still cannot handle the issue that GroupNorm v2 fails to compile on CUDA versions below 12.4; this PR resolves this problem.
https://github.com/NVIDIA/apex/blob/214c6b33270a6054933749845c32224fb825463d/setup.py#L515

nWEIdia:
cc @crcrpar

nWEIdia:
Seems low risk, merging

PR #1928, opened 2025-08-15
setup.py: gate GroupNorm SM targets by CUDA version; avoid sm_100

### LINKED ISSUE #1919: APEX GroupNorm Cannot Be Built with CUDA 12.6 Due to Presence of SM100/120 in setup.py
Setup.py introduced sm100/sm120 regardless of cuda version being used.
When a user is using cuda 12.6 still, they would encounter failures due to the presence of sm100/sm120. 

"nvcc fatal   : Unsupported gpu architecture 'compute_100'" 

Creating this issue to track the fix. 

@ptrblck pointed out that https://github.com/NVIDIA/apex/blame/master/setup.py#L534 needs to be indeed improved.  cc @crcrpar

juyterman1000:
CUDA < 12.8; remove duplicate flag append (fixes #1919)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -532,11 +532,17 @@ if has_flag("--group_norm", "APEX_GROUP_NORM"):
 
     # CUDA group norm supports from SM70
     arch_flags = []
-    # FIXME: this needs to be done more cleanly
-    for arch in [70, 75, 80, 86, 90, 100, 120]:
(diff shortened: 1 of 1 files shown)


nWEIdia:
Hi, we merged https://github.com/NVIDIA/apex/pull/1918 which should help with the issue.

Is this PR still needed?

PR #1929, opened 2025-08-20
Update AMP and DDP usage for imagenet main_amp.py

tinglvv:
cc @Aidyn-A

### DIFF
--- a/examples/imagenet/main_amp.py
+++ b/examples/imagenet/main_amp.py
@@ -17,13 +17,10 @@ import torchvision.models as models
 
 import numpy as np
 
-try:
-    from apex.parallel import DistributedDataParallel as DDP
-    from apex.fp16_utils import *
-    from apex import amp, optimizers
-    from apex.multi_tensor_apply import multi_tensor_applier
-except ImportError:
-    raise ImportError("Please install apex from https://www.github.com/nvidia/apex to run this example.")
+from torch.nn.parallel import DistributedDataParallel as DDP
+
+def to_python_float(scalar_tensor: torch.Tensor):
+    return scalar_tensor.float().item()
 
 def fast_collate(batch, memory_format):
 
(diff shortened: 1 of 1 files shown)


examples/imagenet/main_amp.py
```diff
@@ -17,13 +17,10 @@
...
-    from apex.parallel import DistributedDataParallel as DDP
-    from apex.fp16_utils import *
-    from apex import amp, optimizers
-    from apex.multi_tensor_apply import multi_tensor_applier
-except ImportError:
-    raise ImportError("Please install apex from https://www.github.com/nvidia/apex to run this example.")
+from torch.nn.parallel import DistributedDataParallel as DDP
+
+def to_python_float(scalar_tensor: torch.Tensor):
+    return scalar_tensor.float().item()
```
ptrblck, line 23: Where would this be used?

tinglvv: seems used here https://github.com/NVIDIA/apex/blob/7ae82aa365e891add53750a5b92990699b7cb311/examples/imagenet/main_amp.py#L363-L366

tinglvv: Weird it was never defined previously

tinglvv: These lines were added 6 years ago, my guess is the function definition was removed in between? 
The test that is related to this main_amp.py may have been failed for more than 3 years, I saw a bug to deprecate the usage of DDP/AMP in apex that was opened on 2022..

Aidyn-A: It was defined in the removed code for AMP: https://github.com/NVIDIA/apex/blob/e13873debc4699d39c6861074b9a3b2a02327f92/apex/fp16_utils/fp16util.py#L178-L182

(another reason why doing something like `from apex.fp16_utils import *` is undesirable)

PR #1989, opened 2026-03-25
Add 64-bit indexing fallback for large multi_tensor_l2norm kernels

SongXiaoXi:
# Summary

This PR adds a 64-bit indexing fallback for the multi_tensor_l2norm kernel family when any input tensor has numel() above INT_MAX.

The existing int32 fast path is preserved for normal tensor sizes, while large tensors are dispatched to an int64-indexed path.

# Problem

Apex's multi-tensor metadata stores tensor sizes in int64, but the l2norm family still narrows sizes and chunk indexing to int32 inside the device functors.

For tensors larger than INT_MAX elements, this can produce incorrect norm results and may also lead to out-of-bounds accesses once chunk offsets overflow 32-bit indexing.

# Fix

- add a shared helper to detect when tensor lists require 64-bit indexing
- template the l2norm family functors on index type
- dispatch to int64 indexing only for large tensors
- preserve the existing int32 fast path for the common case

# Affected ops

- multi_tensor_l2norm
- multi_tensor_l2norm_mp
- multi_tensor_l2norm_scale

# Testing

Added large-tensor regression tests covering:
- multi_tensor_l2norm
- multi_tensor_l2norm_mp
- multi_tensor_l2norm_scale

The new tests verify correctness for tensors larger than INT_MAX elements while keeping the existing small/normal tensor path unchanged.

### DIFF
--- a/tests/L0/run_optimizers/test_large_tensor_l2norm.py
+++ b/tests/L0/run_optimizers/test_large_tensor_l2norm.py
@@ -1,0 +1,80 @@
+import unittest
+from functools import wraps
+
+import torch
+
+try:
+    import amp_C
+    from apex.multi_tensor_apply import multi_tensor_applier
+except ImportError:
+    HAS_APEX = False
+else:
+    HAS_APEX = True
+
+from torch.testing._internal.common_device_type import largeTensorTest
+
+INT32_MAX = 2_147_483_647
+LARGE_NUMEL = INT32_MAX + 1
+
+@unittest.skipIf(not HAS_APEX, "`apex` is not found.")
+class LargeTensorL2NormTest(unittest.TestCase):
+    def setUp(self):
+        super().setUp()
+        self.noop_flag = torch.zeros([1], dtype=torch.int32, device="cuda")
+
+    def _make_large_tensor(self, dtype=torch.float16):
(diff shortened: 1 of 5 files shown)


tests/L0/run_optimizers/test_large_tensor_l2norm.py
```diff
@@ -0,0 +1,81 @@
...
+else:
+    HAS_APEX = True
+
+from torch.testing._internal.common_device_type import largeTensorTest
+
+INT32_MAX = 2_147_483_647
+LARGE_NUMEL = INT32_MAX + 1
+
+
+@unittest.skipIf(not HAS_APEX, "`apex` is not found.")
```
crcrpar, line 19: would this be really necessary or would other L0 tests have this skipif logic?

SongXiaoXi: I just followed the existing pattern from [tests/L0/run_optimizers/test_adam.py:49](https://github.com/NVIDIA/apex/blob/c6374ac7ea98838338a993d4597cfb85a541210d/tests/L0/run_optimizers/test_adam.py#L49) for Apex-dependent tests. I’m fine removing the skipIf if you’d prefer not to keep that redundancy here.

PR #1932, opened 2025-09-19
Deprecate apex.contrib.fmha and apex.contrib.multihead_attn

Aidyn-A:
cc @crcrpar @eqy

PR #1995, opened 2026-04-15
Remove tests and examples for deprecated apex.contrib.fmha and apex.contrib.multihead_attn

Aidyn-A:
Per title. This PR removes tests and examples as a follow up to #1932.


crcrpar:
ideally https://github.com/Aidyn-A/apex/tree/remove_controb_mha_tests/apex/contrib/multihead_attn#try-performance-tests-yourself should also get updated as 

> apex/contrib/examples/multihead_attn/perf_test_multihead_attn.py:1
>
> Removing this example script leaves stale references in apex/contrib/multihead_attn/README.md (it instructs users to run contrib/examples/multihead_attn/perf_test_multihead_attn.py). Please update or remove that documentation (or relocate the script) so the README doesn’t point to a non-existent path.

crcrpar (approved):

PR #1996, opened 2026-04-20
Add USE_ROCM to op_builder/builder.py

jithunnair-amd:
Changes brought over from https://github.com/ROCm/apex/pull/324/, as these were the only non-whitespace changes in that PR.


crcrpar:
it feels like the target branch should be ROCm/master instead of here?

PR #2000, opened 2026-05-10
Remove NCCL>=2.10.3 check

crcrpar:
as 2.10.3 was released in 2021 April.

ref: https://docs.nvidia.com/deeplearning/nccl/archives/nccl_2303/release-notes/rel_2-10-3.html


Aidyn-A (approved):
Agree, we are now on NCCL v2.30.x.

Issue #1964, opened 2025-11-29
Reduce the number of ignored rules


crcrpar:
Create a pull request per ignored rule listed in https://github.com/NVIDIA/apex/blob/87868b0e3a39c5f4364ac2d8131a7e63fce0f1a4/pyproject.toml#L8-L25

PR #2003, opened 2026-05-15
Fix E721 type comparisons

LeSingh1:
This fixes one ignored Ruff rule from #1964.

**Changes:**
- Replaces direct type comparisons with `isinstance` or `not isinstance`.
- Removes `E721` from the Ruff ignore list.
- Leaves all other ignored Ruff rules untouched.

**Validation:**
```
python -m ruff check . --select E721
python -m ruff check .
```

Both pass.


crcrpar (approved):
thank you

PR #2021, opened 2026-07-26
fix(contrib): call torch.cuda.is_available() in FP16_Optimizer guard

andrewwhitecdw:
## Summary

`apex.contrib.optimizers.FP16_Optimizer.__init__` is supposed to fail fast with a clear `SystemError("Cannot use fp16 without CUDA.")` on machines without CUDA. The guard never fires, so on a CPU-only machine (or a build without the CUDA extensions) construction proceeds and blows up later with a confusing `RuntimeError: FP16_Optimizer requires cuda extensions` from deep in the initializer.

## Root cause

```python
if not torch.cuda.is_available:
    raise SystemError("Cannot use fp16 without CUDA.")
```

`torch.cuda.is_available` (without parentheses) is the bound *method object*, which is always truthy, so `not torch.cuda.is_available` is always `False` and the `raise` is dead code.

## Fix

Call the function:

```python
if not torch.cuda.is_available():
    raise SystemError("Cannot use fp16 without CUDA.")
```

One-line change; behavior on CUDA machines is unchanged (the guard still does not fire there).

## Testing

On a CPU-only machine with an ephemeral env (`uv run --with torch`), using `torch.optim.SGD` as the wrapped optimizer:

- With the fix: `FP16_Optimizer(opt, static_loss_scale=128.0)` raises `SystemError: Cannot use fp16 without CUDA.` immediately — the intended behavior.
- With the fix stashed (unmodified master, 6424da3): the same call skips the guard and instead raises `RuntimeError: FP16_Optimizer requires cuda extensions` from line 64, confirming the guard was dead code.

⚠️ The CUDA-dependent test suites under `tests/` (L0/L1) require a GPU and the compiled `amp_C` extensions, so they could not be run locally; this change does not touch any CUDA code path, but please rely on CI for the full suite.

## Why existing tests missed it

The optimizer tests run exclusively on CUDA machines, where the guard is a no-op either way; nothing exercises the CPU-only error path.


crcrpar (commented):
looks good but I'm afraid I don't think the test is necessary

andrewwhitecdw:
Closing per adversarial audit: branch contains off-scope or not-a-bug changes.

andrewwhitecdw:
script closed on accident

crcrpar (changes requested):
Apparently too many files are changed. Could you minimize the diff?

PR #2029, opened 2026-09-10
Add support for global NVCC append flags

nWEIdia:
Container builds can set NVCC_APPEND_FLAGS to pass extra NVCC flags (e.g. --compress-mode=size) to every CUDA extension build without duplicating the option in each extension.

Commits created with Claude assistance.


Aidyn-A:
This is gonna be redundant as `NVCC_APPEND_FLAGS` sets flags on NVCC level. 
To clarify, doing:
```bash
export NVCC_APPEND_FLAGS="--compress-mode=size"
nvcc main.cu
```
Is the equivalent to:
```bash
nvcc main.cu --compress-mode=size
```
While this PR proposes doing:
```bash
export NVCC_APPEND_FLAGS="--compress-mode=size"
nvcc main.cu --compress-mode=size
```
Which in the end will result in something like this:
```
nvcc main.cu --compress-mode=size --compress-mode=size
```

nWEIdia:
Good point,  NVCC_APPEND_FLAGS is recognized by nvcc, no need for apex setup.py change. Closing.

